PR #22, opened 2017-11-24
Disallows `while true`

hugoabonizio:
In order to standardize the question and avoid [bikeshedding](https://github.com/crystal-lang/crystal/pull/5146#discussion_r146273623), we should enforce the usage of `loop` statement instead of `while true`.

### DIFF
--- a/src/ameba/rule/while_true.cr
+++ b/src/ameba/rule/while_true.cr
@@ -1,0 +1,39 @@
+module Ameba::Rule
+  # A rule that disallows the use of `while true` instead of using the idiomatic `loop`
+  #
+  # For example, this is considered invalid:
+  #
+  # while true
+  #   do_something
+  #   break if some_condition
+  # end
+  #
+  # And should be replaced by the following:
+  #
+  # loop do
+  #   do_something
+  #   break if some_condition
+  # end
+  #
+  # YAML configuration example:
+  #
+  # ```
+  # WhileTrue:
+  #   Enabled: true
+  # ```
+  #
+  struct WhileTrue < Base
+    properties do
+      description = "Disallows while statements with a true literal as condition"
+    end
+
+    def test(source)
+      AST::Visitor.new self, source
+    end
+
+    def test(source, node : Crystal::While)
+      return unless node.cond.true_literal?
+      source.error self, node.location, "While statement using true literal as condition"
+    end
+  end
+end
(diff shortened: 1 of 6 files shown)


straight-shoota:
`while true` and `loop do` are not exactly the same: `loop` uses a block, `while` doesn't. This can make a difference, for example when you have stacked block contexts.

hugoabonizio:
@straight-shoota can you give an example of this difference?

As I see, they are interchangeable 99% of times, including the ocurrencies in this repository.

src/ameba/rule/while_true.cr
```diff
@@ -0,0 +1,39 @@
...
+  #   do_something
+  #   break if some_condition
+  # end
+  #
+  # And should be replaced by the following:
+  #
+  # loop do
+  #   do_something
+  #   break if some_condition
+  # end
```
veelenga: please use syntax in documentation:

veelenga:
The nice thing about `loop do .. end` is that it is also [will be possible](https://github.com/crystal-lang/crystal/pull/5114) to use inline rescue/ensure/else:

```crystal
loop do
  #
rescue
  #
end
```

instead of:

```crystal
begin
  while true
    #
  end
rescue
  #
end
```

hugoabonizio:
Done :smile:

veelenga:
@hugoabonizio also please add it to [config sample](https://github.com/veelenga/ameba/blob/master/config/ameba.yml). Soon it will be possible to generate it automatically, but for now has to be updated manually (alphabetical order).

@straight-shoota are you able to share an example when `loop do ... end` worse than using `while true` ?

straight-shoota:
I'm sorry, but I don't have a working example. I was just stating that it is not the same and that it *can* make a difference.

Btw. your comparison regarding `rescue` is incorrect, the equivalent with `while` would be:
```crystal
while true
  begin
    #
  rescue
    #
  end
end
```

hugoabonizio:
@straight-shoota I agree that they can be different on specific scenarios, but almost always they are semantically the same (and maybe generate the same code in release mode). Surely this rule can be disabled if that's the case on a project.

@veelenga Added the rule to `ameba.yml`.

veelenga:
> Btw. your comparison regarding rescue is incorrect, the equivalent with while would be:

```crystal
while true
  begin
    #
  rescue
    #
  end
end
```

Didn't know that. Will give it a try 👍

PR #41, opened 2018-03-17
Variable scope & useless assignments

veelenga:
# New entities:

### Scope

Represents a context of the local variable visibility. This is where the local variables belong to.

### Variable

Represents a local variable. Has a link to the scope it belongs to, list of assignments of this variable, list of references of this variable.

### Assignment

Represents the assignment. Assignment may be referenced below, may be not. Has a link to the branch it belongs to.

### Branch

Represents a branch is the branchable node.

### Branchable

Represents a branchable node (i.e. `if`, `unless`, `while`, `case`, `rescue` ect.)

## Examples:

```crystal
class A
  def method
     a = 1
    
     3.times { |i| a = a + i }

     if a > 3
       a = 2
     else 
       a = 3
     end
  
     a
  end
end
```

1.  It has 4 scopes:

- top level scope
- class scope
- def scope
- block scope

2. It has 2 variables:

- `a` belongs to def scope
- `i` belongs to block scope

3. It has 4 assignments. Two last assignments belong to different branches:

```crystal
     if a > 3
       a = 2 # branch A
     else 
       a = 3 # branch B
     end
```

4. It has one `Branchable` (if statement) and two branches that belong to it (A and B)

# Useless Assign

This logic above opens new possibilities to Ameba. For example, it is now easy to add the rules that checks for **useless assignments** and **unused variables**.
  
  For example, this is now considered invalid:
  
  ```crystal
  def method
    var = 1       # <-- useless assignment to variable `var`
    do_something
  end
  ```
 
  
  YAML configuration example:
  
  ```yml
  UselessAssign:
    Enabled: true
  ```


bararchy:
@veelenga looks like a great addition

PR #43, opened 2018-04-11
Use shards build command

Sija:
As per https://github.com/veelenga/ameba/pull/42#pullrequestreview-111160575.


veelenga:
Tested it locally and it works. Thank you

PR #46, opened 2018-04-12
PredicateName: Ignores alternative if it's invalid

### LINKED ISSUE #44: alternative naming is syntax error
This of curse is illegal for Crystal.

```
PredicateName: Favour method name '404?' over 'is_404?'
```

hugoabonizio:
Fixes #44.

### DIFF
--- a/src/ameba/rule/predicate_name.cr
+++ b/src/ameba/rule/predicate_name.cr
@@ -1,6 +1,6 @@
 module Ameba::Rule
   # A rule that disallows tautological predicate names, meaning those that
-  # start with the prefix `has_` or the prefix `is_`.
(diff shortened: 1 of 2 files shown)


src/ameba/rule/predicate_name.cr
```diff
@@ -40,8 +40,11 @@ module Ameba::Rule
 
     def test(source, node : Crystal::Def)
       if node.name =~ /^(is|has)_(\w+)\?/
+        alternative = $2
+        return unless alternative =~ /^[a-z_][a-zA-Z_0-9]*$/
```
veelenga: I think we need to ignore the "alternative" if it starts with `_`. That would not report the following predicate names:

```crystal
def is__404?
end
```

WDYT?

hugoabonizio: Agreed, suggesting `_404?` over `is__404?` gives no gain. :+1:

PR #49, opened 2018-04-16
Crystal source formatting rule

### LINKED ISSUE #40: Crystal formatting checker/rule

veelenga:
closes #40 

Reports if `crystal tool format --check` fails.

https://github.com/crystal-lang/crystal/blob/master/src/compiler/crystal/command/format.cr#L83-L84

### DIFF
--- a/spec/ameba/rule/formatting_spec.cr
+++ b/spec/ameba/rule/formatting_spec.cr
@@ -1,0 +1,58 @@
+require "../../spec_helper"
+
+module Ameba
+  describe Rule::Formatting do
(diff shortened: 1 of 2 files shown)


straight-shoota:
You should consider if it is really useful to embed the compiler into ameba binary. It could just execute the compiler/formatter as an external process. The compiler binary should be available anyway where ameba is used.

veelenga:
Ameba already embeds the compiler, so I don't think it is an issue. The problem is both approaches are slow. I need to find a better way do this.

PR #50, opened 2018-04-16
Constantize messages

veelenga:
This should reduce the mem footprint, as a message string is created only once (in a constant) instead of a string per source.

### DIFF
--- a/src/ameba/rule/comparison_to_boolean.cr
+++ b/src/ameba/rule/comparison_to_boolean.cr
@@ -26,6 +26,8 @@ module Ameba::Rule
       description = "Disallows comparison to booleans"
     end
 
+    MSG = "Comparison to a boolean is pointless"
+
     def test(source)
       AST::Visitor.new self, source
     end
@@ -37,7 +39,7 @@ module Ameba::Rule
 
       return unless comparison? && to_boolean?
 
-      source.error self, node.location, "Comparison to a boolean is pointless"
+      source.error self, node.location, MSG
     end
   end
 end
--- a/src/ameba/rule/constant_names.cr
+++ b/src/ameba/rule/constant_names.cr
@@ -38,7 +38,7 @@ module Ameba::Rule
         name = target.names.first
         return if (expected = name.upcase) == name
 
-        source.error self, node.location, MSG % [expected, name]
+        source.error self, node.location, sprintf(MSG, expected, name)
       end
     end
   end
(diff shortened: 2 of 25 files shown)


straight-shoota:
I don't think this should change anything in the compiled binary. Have you actually compared the memory footprint?

veelenga:
@straight-shoota thanks for looking into it.

Honestly, I didn't do the comparison because it was obvious for me. But you force me to do it, so...

### The main difference

```crystal
require "../src/ameba"
require "benchmark"

source = Ameba::Source.new %()

struct ConstMSGRule < Ameba::Rule::Base
  MSG = "This is the error message"

  def test(source)
    source.error self, 1, 1, MSG
  end
end

struct LiteralMSGRule < Ameba::Rule::Base
  def test(source)
    source.error self, 1, 1, "This is the error message"
  end
end

const_msg = ConstMSGRule.new
liternal_msg = LiteralMSGRule.new

Benchmark.ips do |x|
  x.report("literal") do
    1_000_000.times { liternal_msg.catch(source) }
  end

  x.report("const") do
    1_000_000.times { const_msg.catch(source) }
  end
end

```

produces the following:

```sh
$ crystal bench/constantize_messages.cr --release                                                                                                                         
literal    0.1  (  9.88s ) (± 0.00%)  3.03× slower
  const   0.31  (  3.26s ) (±10.27%)       fastest
```

### Memory footprint

```crystal
require "../src/ameba"
require "benchmark"

source = Ameba::Source.new %()

struct ConstMSGRule < Ameba::Rule::Base
  MSG = "This is the error message"

  def test(source)
    source.error self, 1, 1, MSG
  end
end

struct LiteralMSGRule < Ameba::Rule::Base
  def test(source)
    source.error self, 1, 1, "This is the error message"
  end
end

const_msg = ConstMSGRule.new
liternal_msg = LiteralMSGRule.new

# 1_000_000.times { liternal_msg.catch(source) }
1_000_000.times { const_msg.catch(source) }

```

Measuring via [vmmap](https://developer.apple.com/legacy/library/documentation/Darwin/Reference/ManPages/man1/vmmap.1.html)

**String liternal**

```
Physical footprint:         163.6M
Physical footprint (peak):  172.5M
```

**String constant**
```
Physical footprint:         132.9M
Physical footprint (peak):  132.9M
```

### The binary

Well, in benchmarks you can find the worst case, when the rule always adds the error message. So in reality the difference is not that big.

Running ameba on Crystal repo:

|               | Master branch           | This branch  |
| ------------- |:-------------:| -----:|
| execution time (average) | 1.30 sec | 1.28 sec |
| memory(pick)        | 138.1M      |   134.5M |

PR #52, opened 2018-05-04
Unused argument rule

veelenga:
A rule that reports unused arguments. For example, this is considered invalid:

 ```crystal
 def method(a, b, c) # <-- unused argument `c`
   a + b
 end
 ```
 and should be written as:

 ```crystal
 def method(a, b)
   a + b
 end
 ```

or 

```crystal
def method(a, b, _c)
  a + b
end
```

Also it detects unused arguments in blocks and procs:

```crystal
3.times do |i|   # <-- unused argument `i`
  a + b
end

->(s : Int32) { puts "hi" } # <-- unused argument `s`
```

```crystal
# this is valid
def method(&block)
  block.call
end

def method(&block)  # <-- unused argument `block`
  yield 20
end

```

but in this case `&block` arg is still just a dummy argument and can be omitted.

### DIFF
--- a/spec/ameba/ast/variabling/argument_spec.cr
+++ b/spec/ameba/ast/variabling/argument_spec.cr
@@ -1,0 +1,40 @@
+require "../../../spec_helper"
+
+module Ameba::AST
+  describe Argument do
+    arg = Crystal::Arg.new "a"
+    scope = Scope.new as_node "foo = 1"
+    variable = Variable.new(Crystal::Var.new("foo"), scope)
+
+    describe "#initialize" do
+      it "creates a new argument" do
+        argument = Argument.new(arg, variable)
+        argument.node.should_not be_nil
+      end
+    end
+
+    describe "delegation" do
+      it "delegates location to node" do
+        argument = Argument.new(arg, variable)
+        argument.location.should eq arg.location
+      end
+
+      it "delegates to_s to node" do
+        argument = Argument.new(arg, variable)
+        argument.to_s.should eq arg.to_s
+      end
+    end
+
+    describe "#ignored?" do
+      it "is true if arg starts with _" do
+        argument = Argument.new(Crystal::Arg.new("_a"), variable)
+        argument.ignored?.should be_true
+      end
+
+      it "is false otherwise" do
+        argument = Argument.new(arg, variable)
+        argument.ignored?.should be_false
+      end
+    end
+  end
+end
--- a/spec/ameba/base_spec.cr
+++ b/spec/ameba/base_spec.cr
@@ -2,7 +2,7 @@ require "../spec_helper"
 
 module Ameba::Rule
   struct NoProperties < Rule::Base
-    def test(source)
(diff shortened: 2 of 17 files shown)


straight-shoota:
I don't think it makes sense to enforce removing unused arguments. They might be part of a public API and sometimes need to be there even if they're unused.

`method(a, b, _c)` is a different signature than `method(a, b, c)` and would not match if `c` was used as a named argument at a callsite.

With inlined blocks, it is fine to remove unused arguments. But blocks that are cought in a `Proc` likely need the arguments to match the proc type.

It is certainly useful to identify unused arguments, especially in internal methods. But in public API's they're probably more often explicitly wanted than not. They obviously can't be removed but also shouldn't trigger an error (or require manual disabling) every time.

I don't know what would be a good way to do this. Maybe limit to private methods, but private methods can also be part of a defined API (for example the `unbufferd_*` methods in subclasses of `IO::Buffered`). And you couldn't catch unused arguments in public methods that are only internal implementation.

veelenga:
@straight-shoota 

> But blocks that are cought in a Proc likely need the arguments to match the proc type.

And in this case renaming the argument (prepending the `_`) would work (since it doesn't change the signature), woundn't it?

> method(a, b, _c) is a different signature than method(a, b, c) and would not match if c was used as a named argument at callsite. This could be prevented using c as external name, but that would add way to much unnecessary noise.

This is a very good argument. What I think now about is making each of these types (def, block and proc) be disablable. This is very easy to do within the [configuration file](https://github.com/veelenga/ameba#configuration). For example, if the user would want to disable the verification of the unused arguments in defs, he will need to do:

```yml
# .ameba.yml
UnusedArgument:
  def: false
  # block: true
  # proc: true
```

do you think we need to make this default?

veelenga:
@asterite need your help :)

Could you please explain why there is `&block` syntax. Why Crystal needs such an arguments? Since it is not required and can't be used as variable it looks completely useless:

```crystal
def method(a)
  yield a # => no &block needed
end

def method(a, &block)
  p block        # => doesn't compile
  block.call(a)  # => doesn't compile
  &block.call(a) # => doesn't compile
end
```

is there any usecase?

straight-shoota:
> And in this case renaming the argument (prepending the _) would work (since it doesn't change the signature), wouldn't it?

Yes, renaming should be fine there.

> do you think we need to make this default?

Yes, definitely. False positives would produce lots of annoying noise otherwise.

`&block` is for capturing a block as a `Proc` and passing it around, basically as an anonymous function (see [docs](https://crystal-lang.org/docs/syntax_and_semantics/blocks_and_procs.html)). You can certainly call it. However, in your example, the block argument needs an explicit type declaration in order to work:

```cr
def method(a, &block : String ->)
  block.call(a)
end

method("foo") { |s| puts s }
```

veelenga:
TODO:

- unused method args & super:

```crystal
def foo(a) # <-- a should not be reported
  super
end
```

PR #55, opened 2018-05-12
Add sublime linter plugin

epergo:
Hi! I have created a Sublime Text plugin to run ameba and though it could be added to Readme.


veelenga:
@epergo very cool!

image

veelenga:
Would you like to add some screenshot to the Readme?

PR #57, opened 2018-05-23
Support upcoming Crystal 0.25.0

bcardiff:
The changes are compatible with crystal 0.24.2

There are changes in the JSON::Any API and due to some internal changes the Rule::Base module either need to be transformed to a class, or the cast introduced in the PR is needed.

### DIFF
--- a/spec/ameba/formatter/json_formatter_spec.cr
+++ b/spec/ameba/formatter/json_formatter_spec.cr
@@ -26,7 +26,7 @@ module Ameba
     context "sources" do
       it "shows path to the source" do
         result = get_result [Source.new "", "source.cr"]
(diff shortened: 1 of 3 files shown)


veelenga:
@bcardiff oh, thanks a lot. I'm ok with a cast here and don't want to change all structs to classes. Honestly, I expected more incompatibilities with `0.25.0`. Will do a quick test and merge it.

felixbuenemann:
@veelenga Since crystal 0.25.0 is about to be released it would be great to have a new release tagged which includes these fixes.

veelenga:
@felixbuenemann https://github.com/veelenga/ameba/releases/tag/v0.7.0
Thanks for pointing out

PR #73, opened 2018-08-20
Feature/dockerfile

aisrael:
This adds a `Dockerfile` (and a `.dockerignore`) that can be used to build a ready-to-use ameba Docker image, suitable for Gitlab CI, for example.

To build manually, one can do:

```
$ docker build -t ameba/ameba .
```

To use the resulting image on a local source folder, one has to "volume mount" the current (or target) directory into `/src`:

```
$ docker run -v $(pwd):/src ameba/ameba
```

To use it as part of a Gitlab CI build, for example, it's enough to go:

```
ameba:
  stage: test
  image: ameba/ameba
  script:
    - /usr/bin/ameba
```

The Docker image is based off `alpine:3.8` for the smallest footprint, and weighs in at only around 15.4 MB.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,8 @@
+.*
+Dockerfile
+Makefile
+bench/
+bin/
+spec/
+
+*.html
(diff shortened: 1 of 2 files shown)


veelenga:
Please add info above to the Reame.

PR #74, opened 2018-09-02
Performance/Count: identify usage of size calls that follow to object caller names select and reject

veelenga:
This rule is used to identify usage of `size` calls that follow to object caller names `select` and `reject`.
  
  ```crystal
  [1, 2, 3].select { |e| e > 2 }.size
  [1, 2, 3].reject { |e| e < 2 }.size
  [1, 2, 3].select(&.< 2).size
  [0, 1, 2].reject(&.zero?).size
  ```
  
  And it should be written as this:
  
  ```crystal
  [1, 2, 3].count { |e| e > 2 }
  [1, 2, 3].count { |e| e < 2 }
  [1, 2, 3].count(&.< 2)
  [0, 1, 2].count(&.zero?)
  ```

### DIFF
--- a/src/ameba/rule/performance/count.cr
+++ b/src/ameba/rule/performance/count.cr
@@ -1,0 +1,48 @@
+module Ameba::Rule::Performance
+  # This rule is used to identify usage of `size` calls that follow to object
+  # caller names `select` and `reject`.
+  #
+  # For example, this is considered invalid:
+  #
+  # ```
+  # [1, 2, 3].select { |e| e > 2 }.size
+  # [1, 2, 3].reject { |e| e < 2 }.size
+  # [1, 2, 3].select(&.< 2).size
+  # [0, 1, 2].reject(&.zero?).size
+  # ```
+  #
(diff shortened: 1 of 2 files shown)


src/ameba/rule/performance/count.cr
```diff
@@ -0,0 +1,48 @@
...
+  # [1, 2, 3].reject { |e| e < 2 }.size
+  # [1, 2, 3].select(&.< 2).size
+  # [0, 1, 2].reject(&.zero?).size
+  # ```
+  #
+  # And it should be written as this:
+  #
+  # ```
+  # [1, 2, 3].count { |e| e > 2 }
+  # [1, 2, 3].count { |e| e < 2 }
```
kostya: this is not equal with reject, it should invert condition

veelenga: Corrected in 9fa13848bf65de8b2fae212774ebf85d7b69f7b9. Thanks @kostya

PR #76, opened 2018-09-06
New rule: Performance/FirstLastAfterFilter

veelenga:
This rule is used to identify usage of `first/last/first?/last?` calls that follow filters.

 For example, this is considered inefficient:

 ```crystal
 [-1, 0, 1, 2].select { |e| e > 0 }.first
 [-1, 0, 1, 2].select { |e| e > 0 }.last
 ```

 And can be written as this:

 ```crystal
 [-1, 0, 1, 2].find { |e| e > 0 }
 [-1, 0, 1, 2].reverse_each.find { |e| e > 0 }
 ```

 YAML configuration example:

 ```yml
 Performance/FirstLastAfterFilter
   Enabled: true
   FilterNames:
     - select
 ```

### DIFF
--- a/src/ameba/rule/performance/first_last_after_filter.cr
+++ b/src/ameba/rule/performance/first_last_after_filter.cr
@@ -1,0 +1,51 @@
+module Ameba::Rule::Performance
+  # This rule is used to identify usage of `first/last` calls that follow filters.
+  #
+  # For example, this is considered inefficient:
+  #
+  # ```
+  # [-1, 0, 1, 2].select { |e| e > 0 }.first
+  # [-1, 0, 1, 2].select { |e| e > 0 }.last
+  # ```
+  #
+  # And can be written as this:
+  #
+  # ```
+  # [-1, 0, 1, 2].find { |e| e > 0 }
+  # [-1, 0, 1, 2].reverse.find { |e| e > 0 }
+  # ```
+  #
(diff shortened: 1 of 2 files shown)


src/ameba/rule/performance/first_last_after_filter.cr
```diff
@@ -0,0 +1,51 @@
...
+  # ```
+  # [-1, 0, 1, 2].select { |e| e > 0 }.first
+  # [-1, 0, 1, 2].select { |e| e > 0 }.last
+  # ```
+  #
+  # And can be written as this:
+  #
+  # ```
+  # [-1, 0, 1, 2].find { |e| e > 0 }
+  # [-1, 0, 1, 2].reverse.find { |e| e > 0 }
```
kostya: this reverse create new array, this should be much faster `[-1, 0, 1, 2].reverse_each.find { |e| e > 0 }`

veelenga:
@kostya would you like to review PRs on a regular basis? Like a collaborator?

straight-shoota:
I'd be careful with such performance rules because the results will only be equivalent if the `select` block has no side effects, for example.

veelenga:
@straight-shoota could you please give an example?

kostya:
ameba not auto update code, just recomend it, i think such cases can be handled by hands.

PR #78, opened 2018-09-06
Correctly process record declaration at a top level

### LINKED ISSUE #77: Declaration in `record` macro is incorrectly interpreted as useless assignment
```cr
record Foo, foo = "foo" # Useless assignment to variable `foo`
```

I guess all assignments used as macro arguments shouldn't be treated as a useless assign because you can't know how it will be used in the macro.

veelenga:
fixes #77

/cc @straight-shoota

### DIFF
--- a/spec/ameba/rule/lint/useless_assign_spec.cr
+++ b/spec/ameba/rule/lint/useless_assign_spec.cr
@@ -419,6 +419,35 @@ module Ameba::Rule::Lint
         )
         subject.catch(s).should be_valid
       end
+
+      it "doesn't report if this is a record declaration" do
+        s = Source.new %(
+          record Foo, foo = "foo"
+        )
+        subject.catch(s).should be_valid
+      end
+
+      it "does not report if assignment is referenced after the record declaration" do
+        s = Source.new %(
+          foo = 2
+          record Bar, foo = 3 # foo = 3 is not parsed as assignment
+          puts foo
+        )
+        subject.catch(s).should be_valid
+      end
+
+      it "reports if assignment is not referenced after the record declaration" do
+        s = Source.new %(
+          foo = 2
+          record Bar, foo = 3
+        ), "source.cr"
+        subject.catch(s).should_not be_valid
+        s.issues.size.should eq 1
+
+        issue = s.issues.first
+        issue.location.to_s.should eq "source.cr:2:11"
+        issue.message.should eq "Useless assignment to variable `foo`"
+      end
     end
 
     context "branching" do
--- a/src/ameba/ast/visitors/scope_visitor.cr
+++ b/src/ameba/ast/visitors/scope_visitor.cr
@@ -3,20 +3,8 @@ require "./base_visitor"
 module Ameba::AST
   # AST Visitor that traverses the source and constructs scopes.
   class ScopeVisitor < BaseVisitor
-    SUPER = "super"
-    # List of special top level macros that use assignments
-    # as the macro arguments thus such arguments are not treated as assignments.
-    #
-    # For example:
-    #
-    # ```
-    # bar = 1
-    # record Foo, bar = 2 # this is not an assignment
-    # bar                 # => 1
-    # ```
-    SPECIAL_TOP_LEVEL_MACROS = %w(
-      record
-    )
+    SUPER_NODE_NAME  = "super"
+    RECORD_NODE_NAME = "record"
 
     @current_scope : Scope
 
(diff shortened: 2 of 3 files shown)


spec/ameba/rule/lint/useless_assign_spec.cr
```diff
@@ -419,6 +419,35 @@ module Ameba::Rule::Lint
...
+      it "reports if assignment is not referenced after the record declaration" do
+        s = Source.new %(
+          foo = 2
+          record Bar, foo = 3
+        ), "source.cr"
+        subject.catch(s).should_not be_valid
+        s.issues.size.should eq 1
+
+        issue = s.issues.first
+        issue.location.to_s.should eq "source.cr:2:11"
```
straight-shoota: The line number is actually correct, but it is hard to follow why this is in 2:11. Why don't you use heredoc instead of a regular string literal? This makes it much easier.

```cr
source = Source.new <<-'SOURCE',
  foo = 2
  record Bar, foo = 3
  SOURCE
  "source.cr"
# ...
issue.location.to_s.should eq "source.cr:1:1"
```

src/ameba/ast/visitors/scope_visitor.cr
```diff
@@ -3,6 +3,21 @@ require "./base_visitor"
 module Ameba::AST
   # AST Visitor that traverses the source and constructs scopes.
   class ScopeVisitor < BaseVisitor
+    SUPER = "super"
+    # List of special top level macros that use assignments
+    # as the macro arguments thus such arguments are not treated as assignments.
```
straight-shoota: I'd add a note that this is essentially just guessing that such a call (like "record") probably calls a macro.

src/ameba/ast/visitors/scope_visitor.cr
```diff
@@ -3,6 +3,21 @@ require "./base_visitor"
...
+    #
+    # For example:
+    #
+    # ```
+    # bar = 1
```
straight-shoota: This line is uneccessary for a demonstration. Ditto two below.

src/ameba/ast/visitors/scope_visitor.cr
```diff
@@ -3,6 +3,21 @@ require "./base_visitor"
 module Ameba::AST
   # AST Visitor that traverses the source and constructs scopes.
   class ScopeVisitor < BaseVisitor
+    SUPER = "super"
```
straight-shoota: What is this constant for? Unlike Ruby (where strings are mutable), Crystal doesnt gain any benefit from extracting string literals to constants.

src/ameba/ast/visitors/scope_visitor.cr
```diff
@@ -140,12 +155,18 @@ module Ameba::AST
...
+      case @current_scope
+      when .def?
+        if node.name == SUPER && node.args.empty?
+          @current_scope.arguments.each do |arg|
+            variable = arg.variable
+            variable.reference(variable.node, @current_scope).explicit = false
+          end
+        end
+      when .top_level?
+        return false if SPECIAL_TOP_LEVEL_MACROS.includes?(node.name)
```
straight-shoota: For higher precision matching the signature of the `record` macro, you could check if the first argument is a `Crystal::Path`. This obviously requires more customization than just a list of special macro names.

spec/ameba/rule/lint/useless_assign_spec.cr
veelenga: This is a convention I follow across the project. Heredoc syntax actually looks weird itself and I don't want to use it here.

src/ameba/ast/visitors/scope_visitor.cr
veelenga: Constant is created once across the application, literals (in a method) are created each time you call a method.

veelenga: We already discussed that https://github.com/veelenga/ameba/pull/50#issuecomment-381723855

spec/ameba/rule/lint/useless_assign_spec.cr
straight-shoota: Well, I think `source.cr:2:11` looks super weird. If you don't want to use heredoc, fine. But maybe you can make this less weird in a different way...

src/ameba/ast/visitors/scope_visitor.cr
straight-shoota: `record Foo, bar = 2 # this is not an assignment` says it all.

spec/ameba/rule/lint/useless_assign_spec.cr
veelenga: Okay. This is a topic of a different PR.

src/ameba/ast/visitors/scope_visitor.cr
veelenga: Thanks. I would prefer to leave it as is.

PR #82, opened 2018-11-05
Basic implementation of UnreachableCode rule

veelenga:
This is a very basic implementation of unreachable code rule.

**Work in progress**

### Overview

  A rule that reports unreachable code.
  
   ```crystal
   def method(a)
     return 42
     a + 1    # <= unreachable code
   end
   ```
  
   ```crystal
   a = 1
   loop do
     break
     a += 1    # <= unreachable code
   end
   ```
  
   And has to be written as the following:
  
   ```crystal
   def method(a)
     return 42 if a == 0
     a + 1
   end
   ```
  
   ```crystal
   a = 1
   loop do
     break a > 3
     a += 1
   end
   ```
  
   YAML configuration example:
  
   ```yml
   Lint/UnreachableCode:
     Enabled: true
   ```

### DIFF
--- a/spec/ameba/ast/flow_expression_spec.cr
+++ b/spec/ameba/ast/flow_expression_spec.cr
@@ -1,0 +1,61 @@
+require "../../spec_helper"
+
+module Ameba::AST
+  describe FlowExpression do
+    describe "#initialize" do
+      it "creates a new flow expression" do
+        node = as_node("return 22")
+        parent_node = as_node("def foo; return 22; end")
+        flow_expression = FlowExpression.new node, parent_node
+        flow_expression.node.should_not be_nil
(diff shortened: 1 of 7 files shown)


veelenga:
Thanks. This is still planned to be added to this PR. There are even other words: exit, abort etc

veelenga:
Ameba is now able to detect unreachable code after `return/break/next/raise/exit/abort`. However, this is still a very basic implementation which doesn't work in more difficult situations (in branches). For example:

````crystal
if a > 0
  return :positive
else
  return :negative 
end

a + 1 # => unreachable code, which is not detected by Ameba for a moment. 
````

The extension which will handle such cases will be added later.

PR #89, opened 2019-02-04
Use ~> operator in README installation section

Sija:
That will pickup _PATCH_ version updates from default.


veelenga:
Makes sense. Thank you.

PR #99, opened 2019-03-31
Stop calculating cyclomatic complexity for methods which have Macro c…

### LINKED ISSUE #98: Avoid CyclomaticComplexity for macros
Example: https://github.com/Blacksmoke16/athena/blob/cors/src/routing/handlers/route_handler.cr#L9

veelenga:
…onditions

Fixes #98

We skip cyclomatic complexity for methods which contain `Macro::If/Macro::For` nodes. Because macros usually can't be splitted out into multiple methods.  

/cc @Blacksmoke16

### DIFF
--- a/spec/ameba/ast/visitors/counting_visitor_spec.cr
+++ b/spec/ameba/ast/visitors/counting_visitor_spec.cr
@@ -18,6 +18,32 @@ module Ameba::AST
         visitor.count.should eq 1
       end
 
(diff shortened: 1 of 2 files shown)


Blacksmoke16:
:100: Will this also cover like `{% unless condition %}`?

veelenga:
Yes, unless is transformed to if node

PR #123, opened 2019-12-15
Fix link to crystal-lang.org

hanneskaeufler:
Without the scheme prefix it is being treated as a relative link inside this repo.


veelenga:
Thank you

PR #135, opened 2020-03-24
New rule: `Lint/SharedVarInFiber`

### LINKED ISSUE #132: detect misused variables for spawn
Might be nice to detect this type of problem:

https://forum.crystal-lang.org/t/concurrency-channels/1765

The one at top :)

veelenga:
A rule that disallows using shared variables in fibers.

Using a shared variable in the `spawn` block in most cases leads to unexpected behaviour and is undesired.

For example, having this example:

```crystal
n = 0
channel = Channel(Int32).new

while n < 3
  n = n + 1
  spawn { channel.send n }
end

3.times { puts channel.receive } # => # 3, 3, 3
```

The problem is there is only one shared between fibers variable `i` and when `channel.receive` is executed its value is `3`.

To solve this, the code above needs to be rewritten to the following:

```crystal
n = 0
channel = Channel(Int32).new

while n < 3
  n = n + 1
  m = n
  spawn do { channel.send m }
end

3.times { puts channel.receive } # => # 1, 2, 3
```

This rule is able to find the shared variables between fibers, which are mutated during iterations. So it reports the issue on the first sample and passes on the second one.

There are also other technics to solve the problem above which are [officially documented](https://crystal-lang.org/reference/guides/concurrency.html)

YAML configuration example:

```yml
Lint/SharedVarInFiber:
  Enabled: true
```

### DIFF
--- a/spec/ameba/ast/scope_spec.cr
+++ b/spec/ameba/ast/scope_spec.cr
@@ -33,6 +33,20 @@ module Ameba::AST
     end
   end
 
+  describe "#references" do
+    it "can return an empty list of references" do
+      scope = Scope.new as_node("")
+      scope.references.should be_empty
+    end
+
+    it "allows to add variable references" do
+      scope = Scope.new as_node("")
+      nodes = as_nodes "a = 2"
+      scope.references << Reference.new(nodes.var_nodes.first, scope)
+      scope.references.size.should eq 1
+    end
+  end
+
   describe "#add_variable" do
     it "adds a new variable to the scope" do
       scope = Scope.new as_node("")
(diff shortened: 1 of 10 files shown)


jkthorne (approved):
☕️

straight-shoota:
This is really very tricky behaviour. Sharing a closured variable over multiple fibers is only an issue if the value is mutated. Otherwise this is completely fine. So I'm somewhat skeptical of this change because it doesn't necessarily incentivice to write better code. It might even suggest the opposite.

veelenga:
The rule reports only if the variable is being mutated. Do you see anything wrong?

straight-shoota:
Sorry, I missed the part about mutation. Seems fine like that.

PR #137, opened 2020-03-26
New rule: Lint/RedundantStringCoercion

veelenga:
A rule that disallows string conversion in string interpolation,
which is redundant.

```crystal
"Hello, #{name.to_s}"
```

And this is valid:

```crystal
"Hello, #{name}"
```

YAML configuration example:

```yml
Lint/RedundantStringCoersion
  Enabled: true
```


wontruefree:
you have a check for `to_s` args.  I am assuming this is so the linter does not fail on formatting or other options pass to `to_s`.  But there are no specs for this.

veelenga:
@wontruefree good catch. Thanks

PR #141, opened 2020-03-29
New rule: Style/IsANil

veelenga:
A rule that disallows calls to `is_a?(Nil)` in favor of `nil?`.

This is considered bad:

```crystal
var.is_a? Nil
```

And needs to be written as:

```crystal
var.nil?
```

YAML configuration example:

```yml
Style/IsANil:
  Enabled: true
```

### DIFF
--- a/src/ameba/rule/style/is_a_nil.cr
+++ b/src/ameba/rule/style/is_a_nil.cr
@@ -1,0 +1,38 @@
+module Ameba::Rule::Style
+  # A rule that disallows calls to `is_a?(Nil)` in favor of `nil?`.
+  #
+  # This is considered bad:
+  #
+  # ```
+  # var.is_a? Nil
+  # ```
+  #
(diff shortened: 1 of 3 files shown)


src/ameba/rule/style/is_a_nil.cr
```diff
@@ -0,0 +1,39 @@
...
+  # ```
+  #
+  struct IsANil < Base
+    properties do
+      description "Disallows calls to `is_a?(Nil)` in favor of `nil?`"
+    end
+
+    MSG            = "Use `nil?` instead of `is_a?(Nil)`"
+    PATH_NIL_NAMES = %w(Nil)
+    IS_A_CALL_SIZE = "is_a?".size
```
Sija: That const seems unused.

veelenga: Removed, thanks.

PR #143, opened 2020-04-05
New rule: Lint/BadDirective

veelenga:
A rule that reports incorrect comment directives for Ameba.

For example, the user can mistakenly add a directive to disable a rule that even doesn't exist:

```crystal
# ameba:disable BadRuleName
def foo
  :bar
end
```

YAML configuration example:

```yml
Lint/BadDirective:
  Enabled: true
```

### DIFF
--- a/spec/ameba/rule/lint/bad_directive_spec.cr
+++ b/spec/ameba/rule/lint/bad_directive_spec.cr
@@ -1,0 +1,66 @@
+require "../../../spec_helper"
+
+module Ameba::Rule::Lint
+  describe BadDirective do
+    subject = BadDirective.new
+
+    it "does not report if rule is correct" do
+      s = Source.new %(
+        # ameba:disable Lint/BadDirective
+      )
+      subject.catch(s).should be_valid
+    end
+
+    it "reports if there is incorrect action" do
+      s = Source.new %(
+        # ameba:foo Lint/BadDirective
+      ), "source.cr"
+      subject.catch(s).should_not be_valid
+      s.issues.size.should eq 1
+
+      issue = s.issues.first
+      issue.message.should eq(
+        "Bad action in comment directive: 'foo'. Possible values: disable,enable"
+      )
+      issue.location.to_s.should eq "source.cr:1:1"
+      issue.end_location.to_s.should eq ""
+    end
+
+    it "reports if there are incorrect rule names" do
+      s = Source.new %(
+        # ameba:enable BadRule1,BadRule2
+      ), "source.cr"
+      subject.catch(s).should_not be_valid
+      s.issues.size.should eq 1
+
+      issue = s.issues.first
+      issue.message.should eq(
+        "Such rules do not exist: 'BadRule1,BadRule2'"
+      )
+      issue.location.to_s.should eq "source.cr:1:1"
+      issue.end_location.to_s.should eq ""
+    end
+
+    it "does not report if there no action and rules at all" do
+      s = Source.new %(
+        # ameba:
+      )
+      subject.catch(s).should be_valid
+    end
+
+    it "does not report if there are no rules" do
+      s = Source.new %(
+        # ameba:enable
+        # ameba:disable
+      )
+      subject.catch(s).should be_valid
+    end
+
+    it "does not report if there are group names in the directive" do
+      s = Source.new %(
+        # ameba:disable Style Performance
+      )
+      subject.catch(s).should be_valid
+    end
+  end
+end
--- a/spec/ameba/rule/lint/bad_directive_spec.cr
+++ b/spec/ameba/rule/lint/bad_directive_spec.cr
@@ -20,7 +20,7 @@ module Ameba::Rule::Lint
 
       issue = s.issues.first
       issue.message.should eq(
-        "Bad action in comment directive: 'foo'. Possible values: disable,enable"
+        "Bad action in comment directive: 'foo'. Possible values: disable, enable"
       )
       issue.location.to_s.should eq "source.cr:1:1"
       issue.end_location.to_s.should eq ""
@@ -35,7 +35,7 @@ module Ameba::Rule::Lint
 
       issue = s.issues.first
       issue.message.should eq(
-        "Such rules do not exist: 'BadRule1,BadRule2'"
+        "Such rules do not exist: BadRule1, BadRule2"
       )
       issue.location.to_s.should eq "source.cr:1:1"
       issue.end_location.to_s.should eq ""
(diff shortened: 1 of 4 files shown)


src/ameba/rule/lint/bad_directive.cr
```diff
@@ -0,0 +1,54 @@
...
+  #   Enabled: true
+  # ```
+  #
+  struct BadDirective < Base
+    properties do
+      description "Reports bad comment directives"
+    end
+
+    AVAILABLE_ACTIONS = InlineComments::Action.names.join(',', &.downcase)
+    ALL_RULE_NAMES    = Rule.rules.map(&.rule_name)
```
wontruefree, line 27: If someone extends or adds custom rules will this catch those?

veelenga: Good question. Yes, this method returns all available rules at compile time.

src/ameba/rule/lint/bad_directive.cr
```diff
@@ -0,0 +1,54 @@
...
+  # Lint/BadDirective:
+  #   Enabled: true
+  # ```
+  #
+  struct BadDirective < Base
+    properties do
+      description "Reports bad comment directives"
+    end
+
+    AVAILABLE_ACTIONS = InlineComments::Action.names.join(',', &.downcase)
```
Sija, line 26:
```suggestion
    AVAILABLE_ACTIONS = InlineComments::Action.names.map(&.downcase)
```

src/ameba/rule/lint/bad_directive.cr
```diff
@@ -0,0 +1,54 @@
...
+        check_rules source, token, directive[:rules]
+      end
+    end
+
+    private def check_action(source, token, action)
+      return if InlineComments::Action.parse?(action)
+
+      issue_for token,
+        "Bad action in comment directive: '%s'. Possible values: %s" % {
+          action, AVAILABLE_ACTIONS,
```
Sija, line 45:
```suggestion
          action, AVAILABLE_ACTIONS.join(", "),
```

src/ameba/rule/lint/bad_directive.cr
```diff
@@ -0,0 +1,54 @@
...
+        }
+    end
+
+    private def check_rules(source, token, rules)
+      return if (bad_names = rules - ALL_RULE_NAMES - ALL_GROUP_NAMES).empty?
+      issue_for token, "Such rules do not exist: '%s'" % bad_names.join(',')
```
Sija, line 51:
```suggestion
      issue_for token, "Such rules do not exist: %s' % bad_names.join(", ")
```

spec/ameba/rule/lint/bad_directive_spec.cr
```diff
@@ -0,0 +1,66 @@
...
+      )
+      subject.catch(s).should be_valid
+    end
+
+    it "does not report if there are no rules" do
+      s = Source.new %(
+        # ameba:enable
+        # ameba:disable
+      )
+      subject.catch(s).should be_valid
```
Sija, line 56: I'd suggest disallowing empty directives - they serve no purpose and they might be misleading to their function (or lack thereof).

```suggestion
      subject.catch(s).should_not be_valid
```

src/ameba/inline_comments.cr
```diff
@@ -1,7 +1,7 @@
 module Ameba
   # A module that utilizes inline comments parsing and processing logic.
   module InlineComments
-    COMMENT_DIRECTIVE_REGEX = Regex.new "# ameba : (\\w+) ([\\w/, ]+)".gsub(" ", "\\s*")
+    COMMENT_DIRECTIVE_REGEX = Regex.new "# ameba : (\\w+) ([\\w/, ]*)".gsub(" ", "\\s*")
```
Sija, line 4: I'd suggest a pattern that's a lil' bit more strict:

```suggestion
    COMMENT_DIRECTIVE_REGEX = /# ameba:(?<action>\w+) (?<rules>\w+(?:, \w+)*)/
```

src/ameba/rule/lint/bad_directive.cr
```diff
@@ -0,0 +1,54 @@
...
+      return if InlineComments::Action.parse?(action)
+
+      issue_for token,
+        "Bad action in comment directive: '%s'. Possible values: %s" % {
+          action, AVAILABLE_ACTIONS,
+        }
+    end
+
+    private def check_rules(source, token, rules)
+      return if (bad_names = rules - ALL_RULE_NAMES - ALL_GROUP_NAMES).empty?
```
Sija, line 50:
```suggestion
      bad_names = rules - ALL_RULE_NAMES - ALL_GROUP_NAMES
      return if bad_names.empty?

```

src/ameba/inline_comments.cr
veelenga: This will break the backward compatibility which I wouldn't want to do

```crystal
# ameba : disable AnyRule
```

will not disable, neither reported.

spec/ameba/rule/lint/bad_directive_spec.cr
veelenga: Need to think about it. It was planned to disable all rules with the empty directive.

Sija: Sounds good but this change doesn't do that anyway (and it's out of scope of this PR).

src/ameba/inline_comments.cr
Sija: That's true, but honestly I haven't seen this notation in a wild - and it's weird anyway, why would you write it like that? I don't think any other linters I know supports that either.

veelenga: Why does having a space in the notation look weird to you?

Sija: It seems unnecessary, less concise and breaks up visual flow, which otherwise is, at least for me more consistent - `ameba:disable` - pragma and then whatever comes next are the rules/groups vs `ameba` -> `disable` and then rules/groups. Also, as I said before, all other linters use such notation so there's one thing less to "learn".

spec/ameba/rule/lint/bad_directive_spec.cr
veelenga: Exactly, so I don't want to add restrictions that are not formalized yet.

src/ameba/inline_comments.cr
veelenga: Thanks. Okay, i will change it in the next PR.

PR #146, opened 2020-04-10
Make InlineComments::COMMENT_DIRECTIVE_REGEX more strict

### LINKED ISSUE #145: More strict comment directive
Follow up: https://github.com/crystal-ameba/ameba/pull/143#discussion_r403760570

We would like to make the comment directive more strict and avoid the side effects with spaces. I.e.

Before it was possible:

```crystal
# ameba : disable Rule1, Rule2,   Rule3
```

Now, the only way should be:

```crystal
# ameba:disable Rule1, Rule2, Rule3
```

This is a little breaking change, but since it is not widely used we will just add a note to release notes.

Sija:
Closes #145

### DIFF
--- a/src/ameba/rule/lint/bad_directive.cr
+++ b/src/ameba/rule/lint/bad_directive.cr
@@ -48,7 +48,7 @@ module Ameba::Rule::Lint
 
     private def check_rules(source, token, rules)
       bad_names = rules - ALL_RULE_NAMES - ALL_GROUP_NAMES
-      issue_for token, "Such rules do not exist: %s" % bad_names.join(", ") if bad_names.any?
+      issue_for token, "Such rules do not exist: %s" % bad_names.join(", ") unless bad_names.empty?
     end
   end
 end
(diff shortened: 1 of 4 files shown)


src/ameba/rule/lint/bad_directive.cr
```diff
@@ -48,7 +48,9 @@ module Ameba::Rule::Lint
 
     private def check_rules(source, token, rules)
       bad_names = rules - ALL_RULE_NAMES - ALL_GROUP_NAMES
-      issue_for token, "Such rules do not exist: %s" % bad_names.join(", ") if bad_names.any?
+      unless bad_names.empty?
+        issue_for token, "Such rules do not exist: %s" % bad_names.join(", ")
+      end
```
veelenga, line 53: What the purpose of this change? I would stick to the previous approach if you don't mind :)

If this is about `.any?` vs `.emtpy?` performance, please change it to one-liner `unless bad_names.empty?`

Sija: I've changed it to `unless bad_names.empty?` as requested.

veelenga:
Thank you!

PR #153, opened 2020-05-05
Travis CI: test against Crystal nightlies

Sija:
SSIA

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,9 +1,22 @@
 language: crystal
+
+crystal:
+  - latest
+  - nightly
+
+jobs:
+  allow_failures:
+    - crystal: nightly
+
+install:
+  - shards install
+
 script:
   - make test
   - crystal tool format --check
   - sed -i -e 's:<.*>::g' README.md
   - crystal docs
+
 deploy:
   provider: pages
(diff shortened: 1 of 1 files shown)


Sija:
@veelenga See https://travis-ci.org/github/crystal-ameba/ameba/jobs/683338138 🔴

veelenga:
Yeah, i know. We have [crystal-head](https://github.com/crystal-ameba/ameba/tree/crystal-head) branch with periodic build setup which notifies. I just need to reserve time and look in.

.travis.yml
```diff
@@ -1,15 +1,28 @@
...
+
 script:
   - make test
   - crystal tool format --check
   - sed -i -e 's:<.*>::g' README.md
   - crystal docs
+
 deploy:
   provider: pages
-  github_token: $GITHUB_TOKEN
```
veelenga, line 9: please rever the unrelated changes. Moreover, `github_token` is an alias and imo just works here.

Sija: Will do. Being an alias, adds to the Travis CI validation notes, so I reckoned it's better to use the actual key name instead.

veelenga:
Thank you.

PR #156, opened 2020-06-09
Use crystal-lang/yaml_mapping.cr

bcardiff:
These are the changes I noticed to upgrade to Crystal 0.35 (without warnings). They are backward compatible.

### DIFF
--- a/shard.yml
+++ b/shard.yml
@@ -16,3 +16,7 @@ scripts:
 
 executables:
   - ameba
+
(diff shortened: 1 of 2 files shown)


Sija:
Why not move to `Serializable` instead?

veelenga:
Thanks. Closing in favor of https://github.com/crystal-ameba/ameba/pull/157

PR #173, opened 2020-11-30
drhuffman12 bump crystal-lang version to 1.0

drhuffman12:
Bump version of Crystal up to 1.0 (using the `crystallang/crystal:nightly-alpine-build` docker image).

Make runnable with Crystal 1.0.
(Some logic adjustments might be needed for getting 'correctly format 1.0 code', but you should at least be able to run Ameba with a Crystal-1.0-based app.)

PR #174, opened 2020-12-01
Change supported crystal version to >= 0.35.0

Sija:
Supersedes #173

### DIFF
--- a/shard.yml
+++ b/shard.yml
@@ -1,8 +1,6 @@
 name: ameba
 version: 0.13.2
 
-crystal: 0.29.0
-license: MIT
 authors:
   - Vitalii Elenhaupt <velenhaupt@gmail.com>
 
@@ -16,3 +14,7 @@ scripts:
 
 executables:
   - ameba
+
+crystal: ">= 0.35.0"
+
+license: MIT


veelenga (approved):
I went through the release notes and I think it is still compatible with 0.29. But since this compatibility is not that important and there is no much value in supporting it, i'm ok with this change.

shard.yml
```diff
@@ -1,8 +1,6 @@
 name: ameba
 version: 0.13.2
```
drhuffman12, line 2: Should the version be bumped up?

Sija: Not in this PR, no.

veelenga: @drhuffman12 do you need a new tagged release? I can issue one.

Blacksmoke16: @veelenga Yes please.  https://github.com/athena-framework/validator/pull/9/checks?check_run_id=1496168419

veelenga: @Blacksmoke16 @drhuffman12 published https://github.com/crystal-ameba/ameba/releases/tag/v0.13.3

PR #179, opened 2021-01-09
Add --rules switch to the CLI

Sija:
Adds `--rules` switch allowing to list all available rules.

image


veelenga (approved):
Looks nice. Thanks

PR #180, opened 2021-01-10
Misc refactors

Sija:


src/ameba/config.cr
```diff
@@ -138,7 +139,7 @@ class Ameba::Config
   # ```
   #
   def update_rule(name, enabled = true, excluded = nil)
-    index = @rules.index { |r| r.name == name }
+    index = @rules.index(&.name.== name)
```
veelenga, line 142: TBH i find the previous version mode readable

src/ameba/severity.cr
```diff
@@ -9,8 +9,12 @@ module Ameba
...
     # Severity::Warning.symbol # => 'W'
     # ```
-    def symbol
-      to_s[0]
+    def symbol : Char
+      case self
+      in Error      then 'E'
+      in Warning    then 'W'
+      in Convention then 'C'
+      end
```
veelenga, line 17: this would have to raise in else branch.

veelenga (commented):
Thanks for refactoring. Please take a look at the 2 suggestions below.

PS: Generally speaking, I would like to reject refactoring PRs in the future since they produce a small value but have a risk of introducing a regression.

src/ameba/severity.cr
Sija: That's an exhaustive case (notice `in` instead of `when`) so there's no option for an `else` branch.

src/ameba/config.cr
Sija: Ok, I've reverted this change.

veelenga:
Thank you!

PR #181, opened 2021-01-10
Add support for showing code context lines

### LINKED ISSUE #93: Affected code context
Just the affected line alone doesn't always provide enough information, thus oftentimes it's helpful to see several (ought to be configurable, usually 3/5 is a good default) lines of context before and after it.

Sija:
PoC of affected code context lines feature. Resolves #93

I've pushed it for others to judge, but I believe that showing context lines adds no value practical value and amounts to visual noise instead.

At the moment looks like this:

image

### DIFF
--- a/spec/ameba/formatter/util_spec.cr
+++ b/spec/ameba/formatter/util_spec.cr
@@ -22,7 +22,38 @@ module Ameba::Formatter
           a = 1
         )
         location = Crystal::Location.new("filename", 1, 1)
-        subject.affected_code(source, location).should eq "> a = 1\n  \e[33m^\e[0m"
+        subject.deansify(subject.affected_code(source, location))
+          .should eq "> a = 1\n  ^"
+      end
+
+      it "returns correct line if it is found" do
+        source = Source.new <<-EOF
+          # pre:1
+            # pre:2
+              # pre:3
+                # pre:4
+                  # pre:5
+          a = 1
+                  # post:1
+                # post:2
+              # post:3
+            # post:4
+          # post:5
+          EOF
+
+        location = Crystal::Location.new("filename", 6, 1)
+        subject.deansify(subject.affected_code(source, location, context_lines: 3))
+          .should eq <<-STR
+            >     # pre:3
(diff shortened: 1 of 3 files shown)


veelenga:
I agree about the noise. However, i still think it has a value while inspecting the results on per issue basis. Would be nice have this when [explaining and issue](https://github.com/crystal-ameba/ameba/blob/bedf3a68d7b05769bc1ec68a183f3122bbe15abf/src/ameba/formatter/explain_formatter.cr#L4-L82):

````
$ ameba src/sidekiq/api.cr:108:22
                                                                                                                                                            
## ISSUE INFO

 Shadowing outer local variable `stat`
 src/sidekiq/api.cr:108:22

## AFFECTED CODE

 >       stats = stat.nil? ? all : all & [stat]
 >
 >       mset_args = Hash(String, Int32).new
 >       stats.each do |stat|
                        ^
 >         mset_args["stat:#{stat}"] = 0
 >       end
 >       Sidekiq.redis do |conn|


## RULE INFO

 Warning
 Lint/ShadowingOuterLocalVar
 Disallows the usage of the same name as outer local variables for block or proc arguments.

## DETAILED DESCRIPTION

 A rule that disallows the usage of the same name as outer local variables
 for block or proc arguments.

 For example, this is considered incorrect:

 ```
 def some_method
   foo = 1

   3.times do |foo| # shadowing outer `foo`
   end
 end
 ```

 and should be written as:

 ```
 def some_method
   foo = 1

   3.times do |bar|
   end
 end
 ```

 YAML configuration example:

 ```
 Lint/ShadowingOuterLocalVar:
   Enabled: true
 ```
````

Sija:
I've forgot about explain, good point. I'll change it to be used only there.

Sija:
@veelenga I've made the suggested changes and fixed specs, ready for review.

veelenga (commented):
That looks super nice. Could you please add some specs for `context_lines`?

Sija:
@veelenga Done.

veelenga:
Thanks!

PR #186, opened 2021-01-17
Add Performance/JoinAfterMap rule

### LINKED ISSUE #183: [Feature Request] Performance/JoinAfterMap
I see an opportunity to squash pretty common idiom of using `Enumerable#map(& : -> String)` + `#join(String)` together, amounting to double iteration, while one could use single transformation via `#join(String, & : -> String)` instead.

Sija:
Resolves #183


veelenga (approved):
Cool. Be free to merge if this is ready.

PR #191, opened 2021-01-18
Change Rule to a class

Sija:
Due to `Rule`s having at least one `Array` ivar and being manipulated pretty often, turning it into a class instead should yield performance improvement. It also allows to simplify the code around rule handling.

### DIFF
--- a/spec/spec_helper.cr
+++ b/spec/spec_helper.cr
@@ -3,7 +3,7 @@ require "../src/ameba"
 
 module Ameba
   # Dummy Rule which does nothing.
-  struct DummyRule < Rule::Base
+  class DummyRule < Rule::Base
     properties do
       description : String = "Dummy rule that does nothing."
     end
@@ -35,7 +35,7 @@ module Ameba
     end
   end
 
-  struct NamedRule < Rule::Base
+  class NamedRule < Rule::Base
     properties do
       description "A rule with a custom name."
     end
@@ -45,7 +45,7 @@ module Ameba
     end
   end
 
-  struct ErrorRule < Rule::Base
+  class ErrorRule < Rule::Base
     properties do
       description "Always adds an error at 1:1"
     end
@@ -55,7 +55,7 @@ module Ameba
     end
   end
 
-  struct ScopeRule < Rule::Base
+  class ScopeRule < Rule::Base
     @[YAML::Field(ignore: true)]
     getter scopes = [] of AST::Scope
 
@@ -68,7 +68,7 @@ module Ameba
     end
   end
 
-  struct FlowExpressionRule < Rule::Base
+  class FlowExpressionRule < Rule::Base
     @[YAML::Field(ignore: true)]
     getter expressions = [] of AST::FlowExpression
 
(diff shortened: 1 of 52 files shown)


straight-shoota:
"should yield performance improvement" should be backed by a benchmark ;)

veelenga:
I would like to see some performance proof before doing this migration. 

Rule is a very simple stucture. It has 1-3 properties and methods to change mutate sources (i.e. add issues). In my understanding `struct` is a perfect match for the rule.

Sija:
> Rule is a very simple stucture. It has 1-3 properties and methods to mutate sources (i.e. add issues). In my understanding `struct` is a perfect match for the rule.

Not exactly - it hold several properties, with one of them is an `Array`.
Then, there are Issue which has a `Rule` as a property.
Also, `Config` holds groups with `Rule` values.
If I'm not mistaken all of these places will benefit from having `Rule` as a class.

Not mentioning cutting the cruft that accounts for the fact that you have to update the struct within the parent container every time you update it.

veelenga:
Could you explain how those places would benefit from having `Rule` as a class?

In my understanding, the struct is not re-allocated to the memory until it is mutated (even if they are passed to the methods). Rules do not mutate (at least they should not) after being configured in the config.

```crystal

struct Rule
  property name : String

  def initialize(@name : String = a)
  end
end

class Issue
  getter rule : Rule

  def initialize(@rule : Rule)
  end
end

rule = Rule.new(name: "Rule1")
puts rule.hash            #=> 9015142100968234226

issue = Issue.new(rule: rule)
puts issue.rule.hash      #=> 9015142100968234226

# this part should not be the case in Ameba
# it mutates the struct, thus it is beining re-allocated. 
issue.rule.name = "Rule2"
puts issue.rule.hash     # => 13206662786131213590


```

Sija:
@veelenga `Rule#hash` delegates to `Rule#name.hash` so that won't yield any sensible results for you.

Sija:
> Could you explain how those places would benefit from having `Rule` as a class?

@veelenga In every places mentioned the `Rule` struct is being copied instead of referenced (as it would be if `Rule` was a _class_).

veelenga:
That makes sense. 

There are few other places to update, mostly in docs: [here](https://github.com/crystal-ameba/ameba/blob/3ca811fd608a1508255e218dc89fe565c2b13d8f/src/ameba/rule/base.cr#L59) and below in this file.

Sija:
@veelenga Fixed.

Sija:
> "should yield performance improvement" should be backed by a benchmark ;)

@straight-shoota Putting pressure off the GC is enough to justify that claim but sure, I agree :D

veelenga:
Should be the last piece:

https://github.com/crystal-ameba/ameba/blob/cb12551867da65b0c76a4d6d3b970a12800550fd/src/ameba/rule/base.cr#L73-L75

Sija:
@veelenga Changed.

veelenga (approved):

straight-shoota:
So nobody did actually verify that this improves something? 🤞

Sija:
@straight-shoota My measurements show that runtime got ~5% improvement (statistical error), while memory consumption stayed the same. There's lower number of memory allocations overall and cleaner code - that's IMO a noticeable gain.

PR #195, opened 2021-01-20
Add Performance/ChainedCallsWithNoBang rule

Sija:


src/ameba/rule/performance/chained_calls_with_no_bang.cr
```diff
@@ -0,0 +1,69 @@
...
+  # ```
+  class ChainedCallsWithNoBang < Base
+    properties do
+      description "Identifies usage of chained calls not utilizing the bang method variants."
+    end
+
+    # All of those have bang method variants returning `self`
+    # and are not modifying the reveiver type (like `compact` does),
+    # thus are safe to switch to the bang variant.
+    CALL_NAMES = %w(uniq sort sort_by shuffle reverse rotate)
```
veelenga, line 39: i would make a configurable property to control this list. Example:

https://github.com/crystal-ameba/ameba/blob/7aa7efd4bdf08bfefb1901c97fe6db13ea1aafdf/src/ameba/rule/performance/size_after_filter.cr#L29-L31

src/ameba/rule/performance/chained_calls_with_no_bang.cr
```diff
@@ -0,0 +1,69 @@
...
+  # Performance/ChainedCallsWithNoBang
+  #   Enabled: true
+  # ```
+  class ChainedCallsWithNoBang < Base
+    properties do
+      description "Identifies usage of chained calls not utilizing the bang method variants."
+    end
+
+    # All of those have bang method variants returning `self`
+    # and are not modifying the reveiver type (like `compact` does),
```
veelenga, line 37:
```suggestion
    # and are not modifying the receiver type (like `compact` does),
```

veelenga (approved):
:+1:

Issue #177, opened 2020-12-28
Analysis of invidious repo takes a very long time


CaptainTyphoo:
I am trying to analyze the [invidious repo](https://github.com/iv-org/invidious), but the analysis has been running for about 7 days now and the ameba process has accumulated about 17500 minutes of CPU time according to top. So far the output of ameba is just
```
$ ./ameba
Inspecting 36 files.
```

Are there settings to speed up the analysis? I already set `CRYSTAL_WORKERS=4` but ameba is not even using two full cores.

Is it supposed to take that long?

veelenga:
@CaptainTyphoo hey, thanks for reporting.

I made a quick test and noticed it works well without a single rule `Lint/UselessAssign`. Try disabling it:

```
$ ameba --except Lint/UselessAssign
```

Seems like this rule is kinda inefficient and should be optimized.

CaptainTyphoo:
Thank you! 
The analysis takes less than a second if I disable `Lint/UselessAssign`.

veelenga:
Reduced:

```crystal
response = JSON.build do |json|
   json.object do
     json.object do
       json.object do
         json.object do
           json.object do
             json.object do
               json.object do
                 json.object do
                   json.object do
                     json.object do
                       json.object do
                         anything
                       end
                     end
                   end
                 end
               end
             end
           end
         end
       end
     end
   end
 end
 response = JSON.parse(response)
 response = prepare(response)
 response       

```

PR #196, opened 2021-01-20
Do not report if variable is assigned and referenced in MacroFor/Macr…

### LINKED ISSUE #194: Possible Issue with Lint/UselessAssign
Hey there! Thank you to the contributors for this library, which has become a staple of my Crystal development.

To that end, I'm currently trying to help out the Stripe shard at https://github.com/confact/stripe.cr by getting them onto GitHub actions with Ameba checks. However, when I started cleaning up the repo, I noticed this kind of warning popping up all over the place:

```
src/stripe/methods/core/billing_portal/sessions/create_session.cr:8:5
[W] Lint/UselessAssign: Useless assignment to variable `customer`
> customer = customer.id if customer.is_a?(Customer)
  ^
```

That's coming from this line:
https://github.com/confact/stripe.cr/blob/master/src/stripe/methods/core/billing_portal/sessions/create_session.cr#L8

To the best of my knowledge, this variable _is_ being used down in the macro on this line:
https://github.com/confact/stripe.cr/blob/master/src/stripe/methods/core/billing_portal/sessions/create_session.cr#L13

I'll throw the full class in-line to make it easier to parse quickly:

```crystal
class Stripe::BillingPortal::Session
  def self.create(
    customer : String | Stripe::Customer,
    return_url : String? = nil,
    expand : Array(String)? = nil
  ) : Stripe::BillingPortal::Session
    customer = customer.id if customer.is_a?(Customer) # <---- Warning is thrown here

    io = IO::Memory.new
    builder = ParamsBuilder.new(io)

    {% for x in %w(customer return_url expand) %} # <---- But variable is used here
      builder.add({{x}}, {{x.id}}) unless {{x.id}}.nil?
    {% end %}

    response = Stripe.client.post("/v1/billing_portal/sessions", form: io.to_s)

    if response.status_code == 200
      Stripe::BillingPortal::Session.from_json(response.body)
    else
      raise Error.from_json(response.body, "error")
    end
  end
end
```

I can't speak to whether or not this is the "right" way to write this, but the codebase's tests pass and I'm using the class above successfully in production without issues. It's a pattern that's quite prevalent in the codebase (43 instances at the moment), so I'd rather not perform a bulk refactor if possible.

Is there perhaps another implementation of the line of code throwing the warning that I should be using? Happy to provide more details if necessary!

veelenga:
…oIf/MacroExpression

closes #194

### DIFF
--- a/spec/ameba/rule/lint/useless_assign_spec.cr
+++ b/spec/ameba/rule/lint/useless_assign_spec.cr
@@ -944,6 +944,66 @@ module Ameba::Rule::Lint
         )
         subject.catch(s).should be_valid
       end
+
+      it "doesn't report if assignment is referenced in a macro expression as string" do
+        s = Source.new %(
+          foo = 1
+          puts {{ "foo".id }}
+        )
+        subject.catch(s).should be_valid
+      end
+
+      it "doesn't report if assignement is referenced in for macro in exp" do
+        s = Source.new %(
+          foo = 22
+
+          {% for x in %w(foo) %}
+            add({{x.id}})
+          {% end %}
+        )
+        subject.catch(s).should be_valid
+      end
+
+      it "doesn't report if assignement is referenced in for macro in body" do
+        s = Source.new %(
+          foo = 22
+
+          {% for x in %w(bar) %}
+            puts {{ "foo".id }}
+          {% end %}
+        )
+        subject.catch(s).should be_valid
+      end
+
+      it "doesn't report if assignement is referenced in if macro in cond" do
+        s = Source.new %(
+          foo = 22
+          {% if "foo".id %}
+          {% end %}
+        )
+        subject.catch(s).should be_valid
+      end
+
+      it "doesn't report if assignement is referenced in if macro in then" do
+        s = Source.new %(
+          foo = 22
+          {% if true %}
+             puts {{ "foo".id }}
+          {% end %}
+        )
+        subject.catch(s).should be_valid
+      end
+
+      it "doesn't report if assignement is referenced in if macro in else" do
+        s = Source.new %(
+          foo = 22
+          {% if true %}
+          {% else %}
+             puts {{ "foo".id }}
+          {% end %}
+        )
+        subject.catch(s).should be_valid
+      end
     end
 
     context "uninitialized" do
--- a/src/ameba/ast/variabling/variable.cr
+++ b/src/ameba/ast/variabling/variable.cr
@@ -120,12 +120,13 @@ module Ameba::AST
       false
     end
 
-    # Returns true if current variable potentially referenced in a macro literal,
+    # Returns true if current variable potentially referenced in a macro,
     # false if not.
     def used_in_macro?(scope = @scope)
       scope.inner_scopes.each do |inner_scope|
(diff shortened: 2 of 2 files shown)


straight-shoota:
Using `to_s.includes` is very imprecise. It leads to a lot of failures and more confusing behaviour.

This snippet would pass, for example:
```cr
foo = 1
puts {{ "foo" }}
```

#194 can only be solved by actually expanding macro expressions. Then you can analyze the expanded source code.

spec/ameba/rule/lint/useless_assign_spec.cr
```diff
@@ -944,6 +944,66 @@ module Ameba::Rule::Lint
...
+      end
+
+      it "doesn't report if assignement is referenced in if macro in cond" do
+        s = Source.new %(
+          foo = 22
+          {% if "foo".id %}
+          {% end %}
+        )
+        subject.catch(s).should be_valid
+      end
```
Sija, line 985: That should report.

veelenga:
Yes, it would pass but prevent the false positiveness. The main intention is to exclude the check if the macro could have a reference to the variable in the same (or inner) scope. 

This rule is not designed to work with macro at all.

spec/ameba/rule/lint/useless_assign_spec.cr
veelenga: If you have ideas on how this can be solved, i would be happy to review the suggestion. For now, i leave it as is

straight-shoota:
This just trades false positives for true negatives. I'd rather try to avoid the latter, even though it's less annoying that way.

Sija:
I agree with @straight-shoota here, false positives for this rule are quite rare, so putting `# ameba:disable` won't be much of a burden, OTOH having brittle code leading to false negatives is quite a smell.

veelenga:
> I'd rather try to avoid the latter, even though it's less annoying that way.

I disagree. True negatives are not a big problem here. Having an unused variable doesn't necessarily mean there is a problem in code that leads to incorrect behavior of the app. 

> false positives for this rule are quite rare, so putting # ameba:disable won't be much of a burden

It would be a pain on a big codebase because you can't easily distinguish the correct and incorrect reported issues. For instance, for crystal repo:

```
> ameba --only Lint/UselessAssign
1620 inspected, 225 failures.
```

For each of the cases, the user doesn't know if he needs to disable a rule or remove the unused variable thus will have to do an investigation before making an action.

veelenga:
False positives should have a very well defined reason. For example: "This is potentially a security risk, be sure to verify this and this before disabling the check" or "This is a potential deadlock, try testing your app in different environments...". The incorrectly detected unused variable is just not what we would like to provide.

spec/ameba/rule/lint/useless_assign_spec.cr
Sija: @veelenga It shouldn't be used with `MacroIf`, I think.

Sija:
@veelenga btw, this won't help with issues like #177 :(

veelenga:
@Sija for sure, this is a different problem and will be solved differently. I will take a look.

PR #197, opened 2021-01-22
Add Performance/AnyInsteadOfEmpty rule

Sija:
Using `Enumerable#any?` instead of `Enumerable#empty?` might lead to subtle bugs like `[nil, false].any? # => false`. In some cases it also might be less efficient, since it iterates until the block will return a _truthy_ value, instead of just checking if there's at least one value (or returning an internal `@size`).

### DIFF
--- a/src/ameba/rule/performance/any_instead_of_empty.cr
+++ b/src/ameba/rule/performance/any_instead_of_empty.cr
@@ -1,0 +1,44 @@
+module Ameba::Rule::Performance
+  # This rule is used to identify usage of arg-less `Enumerable#any?` calls.
+  #
+  # Using `Enumerable#any?` instead of `Enumerable#empty?` might lead to an
+  # unexpected results (like `[nil, false].any? # => false`). In some cases
+  # it also might be less efficient, since it iterates until the block will
+  # return a _truthy_ value, instead of just checking if there's at least
+  # one value present.
+  #
+  # For example, this is considered invalid:
+  #
+  # ```
+  # [1, 2, 3].any?
+  # ```
+  #
+  # And it should be written as this:
+  #
+  # ```
+  # ![1, 2, 3].empty?
+  # ```
+  #
+  # YAML configuration example:
+  #
+  # ```
+  # Performance/AnyInsteadOfEmpty:
+  #   Enabled: true
+  # ```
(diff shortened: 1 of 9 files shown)


src/ameba/rule/performance/any_instead_of_empty.cr
```diff
@@ -0,0 +1,38 @@
+module Ameba::Rule::Performance
+  # This rule is used to identify usage of arg-less `any?` calls.
```
veelenga, line 2: This is not obvious why. Could you please extend the rule doc and add more explanation, so the end-user will be able to spot it while explaining the reported issue.

src/ameba/rule/performance/first_last_after_filter.cr
```diff
@@ -45,7 +45,7 @@ module Ameba::Rule::Performance
     def test(source, node : Crystal::Call)
       return unless node.name.in?(CALL_NAMES) && (obj = node.obj)
       return unless obj.is_a?(Crystal::Call) && obj.block
-      return if !node.block.nil? || node.args.any?
+      return unless node.block.nil? && node.args.empty?
```
veelenga, line 48: using unless with multiple conditions is hard to read. This is not enforced but would be nice to avoid this imho. 

```suggestion
      return if if !node.block.nil? || !node.args.empty?
```

Sija: I agree except for `&&` chaining which reads IMO pretty natural - `unless /all of the following/`

src/ameba/rule/performance/any_instead_of_empty.cr
Sija: Good point, I think many more rules could have their details expanded btw.

veelenga: Agree, but I think at least this one is not straightforward and may require the details to avoid questions.

Sija: Fair enough, I'll expand the rule description.

Sija: I've added to the documentation a section explaining the reasons behind this rule.

veelenga (approved):

PR #201, opened 2021-01-24
Add Style/VerboseBlock rule

Sija:
Preview:

image

### DIFF
--- a/src/ameba/rule/style/verbose_block.cr
+++ b/src/ameba/rule/style/verbose_block.cr
@@ -210,7 +210,6 @@ module Ameba::Rule::Style
       while obj.is_a?(Crystal::Call)
         obj = obj.obj
       end
-      return unless obj.is_a?(Crystal::Var)
 
       # only calls with a first argument used as a receiver are the valid game
       return unless obj == arg
--- a/src/ameba/rule/style/verbose_block.cr
+++ b/src/ameba/rule/style/verbose_block.cr
@@ -21,6 +21,7 @@ module Ameba::Rule::Style
   #   Enabled: true
   #   ExcludeMultipleLineBlocks: true
   #   ExcludeCallsWithBlocks: false
+  #   ExcludePrefixOperators: true
   #   ExcludeOperators: false
   #   ExcludeSetters: false
   #   MaxLineLength: ~
@@ -32,6 +33,7 @@ module Ameba::Rule::Style
 
       exclude_calls_with_block true
       exclude_multiple_line_blocks false
+      exclude_prefix_operators true
       exclude_operators false
       exclude_setters false
 
@@ -49,8 +51,13 @@ module Ameba::Rule::Style
       a_location.line_number == b_location.line_number
     end
 
-    private OPERATOR_CHARS =
+    private PREFIX_OPERATORS = {"+", "-"}
+    private OPERATOR_CHARS   =
       {'[', ']', '!', '=', '>', '<', '~', '+', '-', '*', '/', '%', '^', '|', '&'}
+
+    protected def prefix_operator?(node)
+      node.name.in?(PREFIX_OPERATORS) && node.args.empty?
+    end
 
     protected def operator?(name)
       name.each_char do |char|
@@ -138,9 +145,11 @@ module Ameba::Rule::Style
       CALL_PATTERN % {call.name, args, name}
     end
 
+    # ameba:disable Metrics/CyclomaticComplexity
     protected def issue_for_valid(source, call : Crystal::Call, body : Crystal::Call)
       return if exclude_calls_with_block && body.block
       return if exclude_multiple_line_blocks && !same_location_lines?(call, body)
+      return if exclude_prefix_operators && prefix_operator?(body)
       return if exclude_operators && operator?(body.name)
       return if exclude_setters && setter?(body.name)
 
--- a/src/ameba/rule/style/verbose_block.cr
+++ b/src/ameba/rule/style/verbose_block.cr
@@ -23,6 +23,7 @@ module Ameba::Rule::Style
   #   ExcludeCallsWithBlocks: false
   #   ExcludeOperators: false
   #   ExcludeSetters: false
+  #   MaxLineLength: ~
   #   MaxLength: 50 # use ~ to disable
   # ```
   class VerboseBlock < Base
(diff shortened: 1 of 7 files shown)


veelenga (commented):
I tested this on Crystal repo and results look great 👍 

However, i find the suggestion with operator for less readeable:

```crystal
flat_map { |x| x + [100])
flat_map(&.+([100]))
```

that might be subjective.

Sija:
@veelenga re: the operators, I think it's a subjective matter (personally I don't mind using it), but there's an option to exclude them if needed.

src/ameba/rule/style/verbose_block.cr
```diff
@@ -0,0 +1,162 @@
...
+        arr << node_to_s(body)
+      end
+
+      name =
+        call_chain.join('.')
+
+      CALL_PATTERN % {call.name, args, name}
+    end
+
+    # ameba:disable Metrics/CyclomaticComplexity
```
veelenga, line 120: I suggest splitting this into few self-explained smaller methods. Even without `Metrics/CyclomaticComplexity` check this method looks a bit complicated and messy.

src/ameba/rule/style/verbose_block.cr
```diff
@@ -0,0 +1,162 @@
...
+  #
+  # ```
+  # Style/VerboseBlock:
+  #   Enabled: true
+  #   ExcludeMultipleLineBlocks: true
+  #   ExcludeCallsWithBlocks: false
+  #   ExcludeOperators: false
+  #   ExcludeSetters: false
+  #   MaxLineLength: ~
+  #   MaxLength: 50 # use ~ to disable
```
veelenga, line 27: Would be great to document the params. Example:
https://github.com/crystal-ameba/ameba/blob/51b0a07e810501895a71a8f8183ccf752d10d083/src/ameba/rule/style/redundant_next.cr#L49-L89

src/ameba/rule/style/verbose_block.cr
Sija: I'm open to a suggestions since I couldn't come up with anything less complicated for one method and easily readable/comprehensible at the same time.

Sija: With the addition of `exclude_prefix_operators` option, issue is back again, and this time I really have no other idea how to avoid it.

src/ameba/rule/style/verbose_block.cr
```diff
@@ -0,0 +1,229 @@
...
+          named_args.join io, ", " do |arg, inner_io|
+            inner_io << arg.name << ": " << arg.value
+          end
+        end
+      end
+    end
+
+    protected def args_to_s(node : Crystal::Call, skip_last_arg = false)
+      String.build { |io| args_to_s(io, node, skip_last_arg) }
+    end
```
veelenga, line 126: this method seems to be unused.

src/ameba/rule/style/verbose_block.cr
```diff
@@ -0,0 +1,229 @@
...
+            str << '('
+            args_to_s(str, node)
+            str << ')'
+          end
+          str << " {...}" if node.block
+        end
+      end
+    end
+
+    protected def call_code(call, body)
```
veelenga, line 155: I think it would be great to give it just a bit more meaningful name and extract this to the utility class (or mixin) to make it reusable in the future.

src/ameba/rule/style/verbose_block.cr
```diff
@@ -0,0 +1,229 @@
...
+      # we filter out the blocks that are of call type - `i.to_i64.odd?`
+      return unless (body = block.body).is_a?(Crystal::Call)
+
+      # we need to "unwind" the chain challs, so the final receiver object
+      # ends up being a variable - `i`
+      obj = body.obj
+      while obj.is_a?(Crystal::Call)
+        obj = obj.obj
+      end
+      return unless obj.is_a?(Crystal::Var)
```
veelenga, line 217: Could you please elaborate on what does this handle? Also, it is not covered by the test.

src/ameba/rule/style/verbose_block.cr
```diff
@@ -0,0 +1,229 @@
...
+
+      # only calls with a first argument used as a receiver are the valid game
+      return unless obj == arg
+
+      # we bail out if the block node include the block argument
+      return if reference_count(body, arg) > 1
```
veelenga, line 223: I think this can be a bit optimized. Since we don't need to count all the references, we just need to return if at least two are found.

veelenga (commented):
This looks awesome. Please take a look at few suggestions

src/ameba/rule/style/verbose_block.cr
Sija: That's correct, I've added it for the "symmetry", but in the end I was also thinking of getting rid of it ;)

src/ameba/rule/style/verbose_block.cr
Sija: https://github.com/crystal-ameba/ameba/blob/0c5ae963d3e8bf36e39413565ed3b8fd000aa715/src/ameba/rule/style/verbose_block.cr#L211-L217

veelenga: If i comment the last line, tests still pass. I wonder why that is needed

Sija: Good catch! I've removed this check since it's already covered by `return unless obj == arg` guard below.

src/ameba/rule/style/verbose_block.cr
Sija: Fair point, any suggestions onto where should it go?

veelenga: I would put it to [ast/util](https://github.com/crystal-ameba/ameba/blob/feature/verbose-block-rule/src/ameba/ast/util.cr) class. However, if you feel uncomfortable with it, it can be done later on demand.

veelenga (approved):

src/ameba/rule/style/verbose_block.cr
Sija: Cool, I'd prefer to do it in the followup PR though.

src/ameba/rule/style/verbose_block.cr
Sija: That's correct. Would you like it to be done in this PR or could it wait for the followup?

veelenga: This is not very important. I’m totally fine with a followup. Thanks, @Sija

veelenga:
@Sija please be free to merge if this is ready

PR #202, opened 2021-01-24
Add Style/IsAFilter rule

### LINKED ISSUE #175: [Feature Request] .reject(Nil)
I recently used the code
```cr
...
  .reject(&.nil?)
  .map(&.not_nil!)
```
but I learned (ty ary) there is a much better way.
```cr
...
  .reject(Nil)
```

Sija:
Resolves #175

### DIFF
--- a/src/ameba/rule/style/is_a_filter.cr
+++ b/src/ameba/rule/style/is_a_filter.cr
@@ -1,0 +1,74 @@
+module Ameba::Rule::Style
+  # This rule is used to identify usage of `is_a?/nil?` calls within filters.
+  #
+  # For example, this is considered invalid:
+  #
+  # ```
+  # matches = %w[Alice Bob].map(&.match(/^A./))
+  #
+  # matches.any?(&.is_a?(Regex::MatchData)) # => true
+  # matches.one?(&.nil?)                    # => true
+  #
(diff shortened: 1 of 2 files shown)


src/ameba/rule/style/is_a_filter.cr
```diff
@@ -0,0 +1,74 @@
...
+      return unless (block = node.block)
+      return unless (body = block.body).is_a?(Crystal::IsA)
+      return unless (path = body.const).is_a?(Crystal::Path)
+      return unless body.obj.is_a?(Crystal::Var)
+
+      name = path.names.join("::")
+      name = "::#{name}" if path.global? && !body.nil_check?
+
+      end_location = node.end_location
+      if !end_location || end_location.try(&.column_number.zero?)
```
veelenga, line 60: haven't seen the column number to be zero. Can it be so?

Sija: I'm not sure as to how but I've seen it happen.

Sija: IIRC it occurred when expression had no parenthesis.

veelenga: might be a bug in the compiler. But anyway, better to have a condition here for safety.

veelenga (approved):

PR #203, opened 2021-01-30
Avoid exponential recursion while finding variable references in scopes

### LINKED ISSUE #177: Analysis of invidious repo takes a very long time

veelenga:
closes #177 

### Before:

```sh
$ invidious git:master ❯ ameba --only Lint/UselessAssign src/invidious.cr 
Inspecting 1 file.

... hangs out
```

### After:

```sh
$ invidious git:master ❯ ameba --only Lint/UselessAssign src/invidious.cr                               ✭
Inspecting 1 file.

F

src/invidious.cr:407:3
[W] Lint/UselessAssign: Useless assignment to variable `locale`
> locale = LOCALES[env.get("preferences").as(Preferences).locale]?
  ^
src/invidious.cr:410:3
[W] Lint/UselessAssign: Useless assignment to variable `sid`
> sid = env.get? "sid"

...

Finished in 5.89 seconds

1 inspected, 77 failures.
```

---

### DIFF
--- a/spec/ameba/rule/lint/useless_assign_spec.cr
+++ b/spec/ameba/rule/lint/useless_assign_spec.cr
@@ -1006,6 +1006,47 @@ module Ameba::Rule::Lint
       end
     end
 
+    it "does not report if variable is referenced and there is a deep level scope" do
+      s = Source.new %(
+        response = JSON.build do |json|
+          json.object do
+            json.object do
+              json.object do
+                json.object do
+                  json.object do
+                    json.object do
+                      json.object do
+                        json.object do
+                          json.object do
+                            json.object do
+                              json.object do
+                                json.object do
+                                  json.object do
+                                    json.object do
+                                      anything
+                                    end
+                                  end
+                                end
+                              end
+                            end
+                          end
+                        end
+                      end
+                    end
+                  end
+                end
+              end
+            end
+          end
+        end
+
+        response = JSON.parse(response)
+        response = prepare(response)
+        response
+       )
+      subject.catch(s).should be_valid
+    end
+
     context "uninitialized" do
       it "reports if uninitialized assignment is not referenced at a top level" do
         s = Source.new %(
--- a/spec/ameba/rule/lint/useless_assign_spec.cr
+++ b/spec/ameba/rule/lint/useless_assign_spec.cr
@@ -1023,7 +1023,9 @@ module Ameba::Rule::Lint
                                 json.object do
                                   json.object do
                                     json.object do
-                                      anything
+                                      json.object do
+                                        anything
+                                      end
                                     end
                                   end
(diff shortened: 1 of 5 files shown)


src/ameba/ast/variabling/variable.cr
```diff
@@ -113,7 +113,7 @@ module Ameba::AST
     # ```
     def captured_by_block?(scope = @scope)
       scope.inner_scopes.each do |inner_scope|
-        return true if inner_scope.block? && inner_scope.references?(self)
+        return true if inner_scope.block? && inner_scope.references?(self, check_inner_scopes: false)
```
veelenga, line 116: for every **inner scope**:

- we took a variable reference
- and checked if the scope **or any inner scope** had that reference

meaning the number of iterations grows exponentially based on the number of variable references and the number of inner scopes.

spec/ameba/rule/lint/useless_assign_spec.cr
```diff
@@ -1006,6 +1006,47 @@ module Ameba::Rule::Lint
...
+                    end
+                  end
+                end
+              end
+            end
+          end
+        end
+
+        response = JSON.parse(response)
+        response = prepare(response)
```
Sija, line 1044: Is this line needed?

```suggestion
```

src/ameba/rule/lint/useless_assign.cr
```diff
@@ -39,7 +39,7 @@ module Ameba::Rule::Lint
 
     def test(source, node, scope : AST::Scope)
       scope.variables.each do |var|
-        next if var.captured_by_block? || var.used_in_macro? || var.ignored?
+        next if var.ignored? || var.used_in_macro? || var.captured_by_block?
```
Sija, line 42: 👍

spec/ameba/rule/lint/useless_assign_spec.cr
veelenga: Just to increase a number of variable references and make the recursion cycles to grow.

Sija: I see, and could it be replaced with some builtin method, like `pointerof` maybe?

veelenga: Ok, i removed it and added one more level of nesting

PR #218, opened 2021-03-23
Change Style/VerboseBlock - ExcludeOperators - setting default

### LINKED ISSUE #214: Input on Style/VerboseBlock
Hey, there! Forgive me if the Issues tab isn't the place for this, and I'll happily recreate this elsewhere if that's more helpful.

I was recently adding Ameba to a project, and noticed the following suggestion:

```
[C] Style/VerboseBlock: Use short block notation instead: `find(&.creator.==(user))`
> elsif user && (revision = snippet.revisions.find { |rev| rev.creator == user })
                                              ^--------------------------------^
```

This could totally be a personal thing where the answer is "If you don't like it, configure it off in your `.ameba.yml` file", but I thought I'd at least raise the question since I'm trying **very** hard to stick to a zero-customization version of Ameba. I went down the config-file-per-project route with RuboCop in my Ruby projects, and as a result I now pretty much exclusively use StandardRB to avoid maintaining those config files.

I *love* this correction for the simpler substitutions. If I were writing `snippet.revisions.map { |rev| rev.to_s }`, I'd certainly want Ameba to tell me that I should use `snippet.revisions.map(&.to_s)`.

However, for a `find` call like this, `find(&.creator.==(user))` seems to be more difficult to visually parse in the same way as `find { |rev| rev.creator == user }`, and a little bit less "pretty".

Again, I know this is somewhat personal and subjective, ~~and I'm not sure how to really "fix" this other than to not recommend `VerboxBlock` fixes for operations that require a comparator like `find`,~~ but thought I'd at least put it out into the universe in an attempt to continue not needing to manage a `.ameba.yml` config file.

stephendolan:
Closes #214, which also provides some examples and input from the core team.

### DIFF
--- a/spec/ameba/rule/style/verbose_block_spec.cr
+++ b/spec/ameba/rule/style/verbose_block_spec.cr
@@ -89,9 +89,11 @@ module Ameba::Rule::Style
         rule = VerboseBlock.new
         rule
           .tap(&.exclude_prefix_operators = true)
+          .tap(&.exclude_operators = true)
           .catch(source).should be_valid
         rule
           .tap(&.exclude_prefix_operators = false)
+          .tap(&.exclude_operators = false)
           .catch(source).should_not be_valid
       end
 
@@ -165,9 +167,6 @@ module Ameba::Rule::Style
         %|map(&.to_s.[start: 0.to_i64, count: 3]?)|,
         %|map(&.to_s.[0.to_i64, count: 3]?)|,
         %|map(&.to_s.[0.to_i64, 3]?)|,
-        %|map(&.to_s.[start: 0.to_i64, count: 3]=("foo"))|,
-        %|map(&.to_s.[0.to_i64, count: 3]=("foo"))|,
-        %|map(&.to_s.[0.to_i64, 3]=("foo"))|,
         %|map(&.to_s.camelcase(lower: true))|,
         %|map(&.to_s.camelcase)|,
         %|map(&.to_s.gsub('_', '-'))|,
@@ -180,9 +179,6 @@ module Ameba::Rule::Style
         (1..3).map { |i| i.to_s[start: 0.to_i64, count: 3]? }
         (1..3).map { |i| i.to_s[0.to_i64, count: 3]? }
         (1..3).map { |i| i.to_s[0.to_i64, 3]? }
-        (1..3).map { |i| i.to_s[start: 0.to_i64, count: 3] = "foo" }
-        (1..3).map { |i| i.to_s[0.to_i64, count: 3] = "foo" }
-        (1..3).map { |i| i.to_s[0.to_i64, 3] = "foo" }
         (1..3).map { |i| i.to_s.camelcase(lower: true) }
         (1..3).map { |i| i.to_s.camelcase }
         (1..3).map { |i| i.to_s.gsub('_', '-') }
--- a/spec/ameba/rule/style/verbose_block_spec.cr
+++ b/spec/ameba/rule/style/verbose_block_spec.cr
@@ -89,7 +89,6 @@ module Ameba::Rule::Style
         rule = VerboseBlock.new
         rule
           .tap(&.exclude_prefix_operators = true)
-          .tap(&.exclude_operators = true)
           .catch(source).should be_valid
         rule
           .tap(&.exclude_prefix_operators = false)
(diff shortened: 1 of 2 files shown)


spec/ameba/rule/style/verbose_block_spec.cr
```diff
@@ -89,9 +89,11 @@ module Ameba::Rule::Style
         rule = VerboseBlock.new
         rule
           .tap(&.exclude_prefix_operators = true)
+          .tap(&.exclude_operators = true)
```
stephendolan, line 92: Only the test on line 94 actually needed an accompanying `exclude_operators` setting to pass, but I've added it to both tests for consistency until we figure out if this is an acceptable solution.

spec/ameba/rule/style/verbose_block_spec.cr
```diff
@@ -89,9 +89,11 @@ module Ameba::Rule::Style
...
           .catch(source).should be_valid
         rule
           .tap(&.exclude_prefix_operators = false)
+          .tap(&.exclude_operators = false)
```
stephendolan, line 96: This is where I was a bit confused. **Without** the additional setting of `exclude_operators` to `false` (since the default is now `true`), this was returning all of the following as issues:

```crystal
source = Source.new %(
          (1..3).sum { |i| +i }
          (1..3).sum { |i| -i }
          (1..3).sum { |i| ~i }
        )
```

I tried to do some basic `puts` debugging in the `prefix_operator?` and `operator?` methods to determine whether this was intentional or not, but I'm not 100% sure.

spec/ameba/rule/style/verbose_block_spec.cr
```diff
@@ -165,9 +167,6 @@ module Ameba::Rule::Style
         %|map(&.to_s.[start: 0.to_i64, count: 3]?)|,
         %|map(&.to_s.[0.to_i64, count: 3]?)|,
         %|map(&.to_s.[0.to_i64, 3]?)|,
-        %|map(&.to_s.[start: 0.to_i64, count: 3]=("foo"))|,
```
stephendolan, line 168: It feels weird to just rip out test cases, but if we introduce a new default that would make certain scenarios non-issues, is this the right approach?

Sija: Nope, all of these should stay as they're testing the formatting, regardless of the default setting.

spec/ameba/rule/style/verbose_block_spec.cr
Sija: Setting it here is not needed.

spec/ameba/rule/style/verbose_block_spec.cr
Sija: That's correct. `ExcludePrefixOperators` setting is additionally narrowing `ExcludeOperators`.

Sija (changes requested):

spec/ameba/rule/style/verbose_block_spec.cr
stephendolan: Right, I was mainly setting it in case we change the defaults again, to make it more apparent that the prefix setting is closely tied to the more general operator setting.

I can remove it!

spec/ameba/rule/style/verbose_block_spec.cr
stephendolan: Got it! I wonder if I should maybe add a comment after the docblock reference to the setting to add some verbiage saying as much more explicitly?

spec/ameba/rule/style/verbose_block_spec.cr
Sija: Please do as it's not needed :)

spec/ameba/rule/style/verbose_block_spec.cr
stephendolan: The trouble is that I think this uses the defaults for running the checks, so these three lines weren't showing as issues when I changed the default value.

Should I move these test cases up to `#exclude_operators`? It was my assumption that _that_ was where we'd put formatting tests that might change based on the settings.

Sija: No, these belong here as they test the formatting, not the option behavior. Since the default changed, just set it back to needed value at the beginning of the test.

stephendolan:
@Sija Thank you for taking the time to review this. I believe I've incorporated all of your feedback, and this is ready for a final glance.

veelenga (approved):

Sija (approved):
LGTM 🚀

Issue #167, opened 2020-10-16
Ameba does not install on FreeBSD


yujiri8:
Ameba's postinstall script seems to not work on FreeBSD. I initialized an empty project with `crystal init app .`, added the following to `shard.yml`:
```yaml
dependencies:
  ameba:
    github: crystal-ameba/ameba
```
And ran `shards`. Error message:
```
Resolving dependencies
Fetching https://github.com/crystal-ameba/ameba.git
Installing ameba (0.13.2)
Postinstall of ameba: make bin && make run_file
Failed postinstall of ameba on make bin && make run_file:
build
make: exec(build) failed (No such file or directory)
*** Error code 1

Stop.
make: stopped in <project dir>/lib/ameba
```
This is a FreeBSD machine that has GNU make installed from packages (though I suspect the problem is that the FreeBSD package installs GNU make as `gmake` since `make` is BSD make).

yujiri8:
The error is also produced using `version: ~> 0.12.0` as mentioned in the README.

veelenga:
@yujiri8 `0.12.0` in readme is a mistake, I've just corrected it. Will try to reproduce the issue with make and fix. Thanks for the heads-up

veelenga:
Hm, I thought I would be able to run FreeBSD docker container on Mac, but seems like FreeBSD containers must be run from FreeBSD hosts. 

@yujiri8 I would like to ask you for helping me with this and providing a fix before I ended up running FreeBSD, on top of Vagrant, on top of macOS. Will that be possible for you?

yujiri8:
I was right that the problem is GNU make vs BSD make. BSD make doesn't support the `$(shell )` syntax for running shell commands. As far as I can tell, it only supports `var != command`, but that isn't the same as GNU's `?=`, it's like `:=`. I'm not sure about a compatible alternative to `?=`.

Here are some links I read:
https://www.gnu.org/software/make/manual/html_node/Setting.html
https://stackoverflow.com/questions/52912099/how-to-conditionally-assign-value-to-a-variable-in-a-makefile-bsd-gnu/52912640
https://stackoverflow.com/questions/21874468/run-a-command-in-gnu-and-bsd-makefile-and-store-the-result-into-a-variable
https://stackoverflow.com/questions/2019989/how-to-assign-the-output-of-a-command-to-a-makefile-variable

Ugghhhh this is why I hate `make`...

straight-shoota:
[...]

PR #221, opened 2021-04-02
Make Makefile freebsd compatible

carlhoerberg:
Maybe there's a good reason to have `shell` there but it isn't freebsd compatible.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,5 +1,5 @@
-CRYSTAL_BIN ?= $(which crystal)
-SHARDS_BIN ?= $(which shards)
+CRYSTAL_BIN ?= $(shell which crystal)
+SHARDS_BIN ?= $(shell which shards)
 PREFIX ?= /usr/local
 SHARD_BIN ?= ../../bin
 
--- a/Makefile
+++ b/Makefile
@@ -1,5 +1,5 @@
-CRYSTAL_BIN ?= $(shell which crystal)
-SHARDS_BIN ?= $(shell which shards)
+CRYSTAL_BIN ?= $(which crystal)
+SHARDS_BIN ?= $(which shards)
 PREFIX ?= /usr/local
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -1,5 +1,5 @@
-CRYSTAL_BIN ?= $(shell which crystal)
-SHARDS_BIN ?= $(shell which shards)
+CRYSTAL_BIN ?= $(which crystal)
+SHARDS_BIN ?= $(which shards)
```
veelenga, line 2: Unfortunately, that broke the CI builds on non freebsd env. I think it can be a bit simplified. 

```suggestion
CRYSTAL_BIN ?= crystal
SHARDS_BIN ?= shards
```

veelenga:
Relates to https://github.com/crystal-ameba/ameba/issues/167

carlhoerberg:
Ok, what out not relying on make at all in the post_install script? See latest changes in this PR

veelenga:
Is there any problem with the suggestion above?

carlhoerberg:
oh i see, no sure, taht works, just figured it was even easier to not rely on `make` at all, but can update the PR with your changes.

veelenga:
Sorry closed in favor of https://github.com/crystal-ameba/ameba/pull/221

straight-shoota:
You mean #222 probably?

straight-shoota:
Makefiles typically use GNU makefile format, both in general and for Crystal projects in particular, because that's the default on Linux and macOS. So you should use GNU make to run it. FreeBSD's default `make` is the BSD implementation which has a few incompatibilities. It's usually best to just run `gmake` on FreeBSD instead of trying to maintain a portable Makefile.

straight-shoota:
That being said, the change here is probably good anyway. There's no reason to resolve the executables.

PR #222, opened 2021-04-03
Support freebsd

### LINKED ISSUE #167: Ameba does not install on FreeBSD

veelenga:
- [x] remove incompatible makefile instructions
- [x] added CI step (refs https://github.com/marketplace/actions/freebsd-vm)

duplicates #221
closes #167

image

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -42,11 +42,10 @@ jobs:
     - uses: actions/checkout@v2
     - name: Install Crystal and test
       id: test
-      uses: vmactions/freebsd-vm@v0.1.2
+      uses: vmactions/freebsd-vm@v0.1.3
       with:
         usesh: true
-        prepare: pkg install -y crystal shards
+        prepare: pkg install -y libyaml libevent crystal shards
         run: |
           shards install
           make test
-          crystal tool format --check
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -45,7 +45,7 @@ jobs:
       uses: vmactions/freebsd-vm@v0.1.3
       with:
         usesh: true
-        prepare: pkg install -y libyaml libevent crystal shards
+        prepare: pkg install -y libyaml libevent crystal shards gmake
         run: |
           shards install
-          make test
+          gmake test
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -34,3 +34,19 @@ jobs:
 
       - name: Check formatting
         run: crystal tool format --check
+
+  testfreebsd:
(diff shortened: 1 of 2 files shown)


.github/workflows/ci.yml
```diff
@@ -34,3 +34,18 @@ jobs:
...
+    - uses: actions/checkout@v2
+    - name: Install Crystal and test
+      id: test
+      uses: vmactions/freebsd-vm@v0.1.3
+      with:
+        usesh: true
+        prepare: pkg install -y libyaml libevent crystal shards
+        run: |
+          shards install
+          make test
```
straight-shoota, line 51:
```suggestion
        prepare: pkg install -y libyaml libevent crystal shards gmake
        run: |
          shards install
          gmake test
```

Even if the current makefile works for both, GNU and BSD dialects, it's better to just use GNU make everywhere.
See https://github.com/crystal-ameba/ameba/pull/221#issuecomment-812908622

.github/workflows/ci.yml
```diff
@@ -34,3 +34,18 @@ jobs:
...
+    - uses: actions/checkout@v2
+    - name: Install Crystal and test
+      id: test
+      uses: vmactions/freebsd-vm@v0.1.3
+      with:
+        usesh: true
+        prepare: pkg install -y libyaml libevent crystal shards gmake
+        run: |
+          shards install
+          gmake test
```
Sija, line 51: That would be better fitting under the `matrix.include` key.

veelenga: Not sure why that is needed, since the matrix allows to spawn several jobs which is not planned here.

veelenga:
Can't make the freebsd job build stable. Might be some issue with [freebsd-vm](https://github.com/marketplace/actions/freebsd-vm) since it just fails randomly. Closing this in favor of https://github.com/crystal-ameba/ameba/commit/ecafa271ee616b641bc9956b4178ade1b145c573

PR #226, opened 2021-04-12
Don't warn on "map + join" in MapInsteadOfBlock rule

### LINKED ISSUE #208: Question on Performance/MapInsteadOfBlock
I just ran ameba 0.14 against Avram, and got a lot of the `Performance/MapInsteadOfBlock` suggestions telling me "Use `join {...}` instead of `map {...}.join`". I wanted to see how much of a performance boost we were going to get from this, so I made a quick little sample:

```crystal
require "benchmark"

DATA = [1, 2, 3, 4, 5, 6, 1, 2, 3, 4, 5, 6]

def use_map_join
  DATA.map(&.to_s).join(", ")
end

def use_join
  DATA.join(", ") { |d| d.to_s }
end

Benchmark.ips do |x|
  x.report("map join") {
    use_map_join
  }
  x.report("join") {
    use_join
  }
end
```

But when I ran this, it tells me that the map/join variant is faster 🤔 

```
❯ ./speed
map join   2.81M (355.56ns) (± 4.78%)  353B/op        fastest
    join   1.83M (546.32ns) (± 3.48%)  337B/op   1.54× slower
```

I'm running this on
```
❯ crystal -v
Crystal 0.36.1 (2021-02-02)

LLVM: 11.0.1
Default target: x86_64-apple-macosx
```

Is there a specific case where the join with block is faster? (i.e. arrays over X amount of elements)

Sija:
Resolves #208 without resorting to disabling the entire rule.
Refs #183

### DIFF
--- a/spec/ameba/rule/performance/map_instead_of_block_spec.cr
+++ b/spec/ameba/rule/performance/map_instead_of_block_spec.cr
@@ -6,30 +6,29 @@ module Ameba::Rule::Performance
   describe MapInsteadOfBlock do
     it "passes if there is no potential performance improvements" do
       source = Source.new %(
-        (1..3).join(&.to_s)
         (1..3).sum(&.*(2))
         (1..3).product(&.*(2))
       )
       subject.catch(source).should be_valid
     end
 
-    it "reports if there is map followed by join without a block" do
+    it "reports if there is map followed by sum without a block" do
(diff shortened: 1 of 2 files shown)


veelenga (approved):
Makes sense. Thank you.

PR #229, opened 2021-04-15
Disable performance rules for spec files

### LINKED ISSUE #220: Performance/AnyInsteadOfEmpty in spec context
```
flash_store.any?.should be_true
```

is suggested to be:

```
!(flash_store.empty?).should
```

which is incorrect. 

refs https://github.com/luckyframework/lucky/pull/1447

veelenga:
closes #220

- [x] added `Performance::Base` class which disables performance rules for spec files 
- [x] added logic to properly list all available rules having more than 1 level of inheritance
- [x] added tests

### DIFF
--- a/spec/ameba/base_spec.cr
+++ b/spec/ameba/base_spec.cr
@@ -8,6 +8,17 @@ module Ameba::Rule
         rules.should_not be_nil
         rules.should contain DummyRule
       end
+
+      it "contains rules across all the available groups" do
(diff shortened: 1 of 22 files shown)


Sija (commented):
Instead of adding one test for each of the performance spec I'd add a `spec/ameba/rule/performance/base_spec.cr` with those/this.

veelenga:
I think that adding a spec in each file still makes sense because we change each rule. However, I also added a spec for `Performance::Base`

PR #238, opened 2021-06-05
Count an exhaustive `case` complexity as 1

caspiano:
Methods with exhaustive cases over enums triggering false positives for cyclomatic complexity.
For example, one may have a state machine where an exhaustive case over enum members is used to dispatch to methods.

This PR counts an exhaustive case's complexity as 1, otherwise treats an inexhaustive case's `when`s as it did previously.

### DIFF
--- a/spec/ameba/ast/visitors/counting_visitor_spec.cr
+++ b/spec/ameba/ast/visitors/counting_visitor_spec.cr
@@ -48,8 +48,8 @@ module Ameba::AST
         code = %(
           def hello(a : Int32 | Int64 | Float32 | Float64)
             case a
-            in Int32 then "int32"
-            in Int64 then "int64"
+            in Int32   then "int32"
+            in Int64   then "int64"
             in Float32 then "float32"
             in Float64 then "float64"
             end
--- a/spec/ameba/ast/visitors/counting_visitor_spec.cr
+++ b/spec/ameba/ast/visitors/counting_visitor_spec.cr
@@ -44,14 +44,33 @@ module Ameba::AST
         visitor.count.should eq 1
       end
 
+      it "increases count for every exhaustive case" do
+        code = %(
+          def hello(a : Int32 | Int64 | Float32 | Float64)
+            case a
+            in Int32 then "int32"
+            in Int64 then "int64"
+            in Float32 then "float32"
+            in Float64 then "float64"
+            end
+          end
+        )
+        node = Crystal::Parser.new(code).parse
+        visitor = CountingVisitor.new node
+        visitor.count.should eq 2
+      end
+
       {% for pair in [
                        {code: "if true; end", description: "conditional"},
                        {code: "while true; end", description: "while loop"},
                        {code: "until 1 < 2; end", description: "until loop"},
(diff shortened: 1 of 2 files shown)


src/ameba/ast/visitors/counting_visitor.cr
```diff
@@ -23,13 +23,26 @@ module Ameba::AST
...
     {% end %}
 
+    # :nodoc:
+    def visit(node : Crystal::Case)
+      # Count the complexity of an exhaustive `Case` as 1
+      # Otherwise count the number of `Whens`
+      if node.exhaustive?
+        @complexity += 1 unless macro_condition
+      else
+        node.whens.each { @complexity += 1 } unless macro_condition
```
Sija, line 40:
```suggestion
        @complexity += node.whens.size unless macro_condition
```

spec/ameba/ast/visitors/counting_visitor_spec.cr
```diff
@@ -44,14 +44,31 @@ module Ameba::AST
         visitor.count.should eq 1
       end
 
+      it "increases count for every exhaustive case" do
+        code = %(
+          def hello(a : Int32 | Int64)
```
Sija, line 49: I'd add more branches (types) to make the difference more visible.

src/ameba/ast/visitors/counting_visitor.cr
caspiano: Woops! thanks for that

veelenga (approved):

src/ameba/ast/visitors/counting_visitor.cr
```diff
@@ -23,13 +23,22 @@ module Ameba::AST
...
       end
     {% end %}
 
+    # :nodoc:
+    def visit(node : Crystal::Case)
+      # Count the complexity of an exhaustive `Case` as 1
+      # Otherwise count the number of `When`s
+      case_complexity = node.exhaustive? ? 1 : node.whens.size
+      @complexity += case_complexity unless macro_condition
+      true
```
Sija, line 39:
```suggestion
      return true if macro_condition

      # Count the complexity of an exhaustive `Case` as 1
      # Otherwise count the number of `When`s
      @complexity += node.exhaustive? ? 1 : node.whens.size

      true
```

spec/ameba/ast/visitors/counting_visitor_spec.cr
```diff
@@ -44,14 +44,33 @@ module Ameba::AST
...
 
+      it "increases count for every exhaustive case" do
+        code = %(
+          def hello(a : Int32 | Int64 | Float32 | Float64)
+            case a
+            in Int32 then "int32"
+            in Int64 then "int64"
+            in Float32 then "float32"
+            in Float64 then "float64"
+            end
```
Sija, line 55: Tiny indent tweak:

```suggestion
            case a
            in Int32   then "int32"
            in Int64   then "int64"
            in Float32 then "float32"
            in Float64 then "float64"
            end
```

Sija (approved):
LGTM ✨

Sija:
Thanks, @caspiano! 🙏

PR #248, opened 2021-10-25
Add autocorrect

FnControlOption:
User-facing changes:

- Added CLI flag `--fix`.
- Added autocorrect for `Style/LargeNumbers` and partially for `Layout/TrailingBlankLines`.
- Added optional `block` parameter to `Reportable#add_issue`. The `block` receives a `Source::Corrector` (described below) to perform autocorrection.
- Added `expect_correction` and `expect_no_corrections` spec helpers.

Implementation details:

- Added `Source::Rewriter` and `Source::Rewriter::Action`, which are ports of [`Parser::Source::TreeRewriter`](https://github.com/whitequark/parser/blob/master/lib/parser/source/tree_rewriter.rb) and [`Parser::Source::TreeRewriter::Action`](https://github.com/whitequark/parser/blob/master/lib/parser/source/tree_rewriter/action.rb) from the Ruby gem [`parser`](https://rubygems.org/gems/parser). These classes accept source positions as integer indexes (instead of line and column).
- Added `Source::Corrector`, which wraps around a `Source::Rewriter`. This class provides helper methods that accept source locations as `Crystal::Location` or `{Int32, Int32}`. (The length of each source line is stored in an array, which is used to convert AST locations to `Rewriter` positions.)

Changes required for recursive autocorrect (f39a7a4):

- Added `Source#correct`, which runs autocorrection and updates the underlying `@code` property.
- `Issue.new` now requires an additional `code` parameter (because `Source#code` is no longer immutable).
- `Formatter::Util#affected_code` now has an additional method overload.
  - The original method now receives a `String` as the first parameter instead of `Source`.
  - The new overload accepts an `Issue` so that we can now do `affected_code(issue)` instead of `affected_code(issue.code, issue.location, issue.end_location)`

Todo:

- [x] Update other formatters in `src/ameba/formatter` to indicate whether any issues have been autocorrected.
- [x] Add missing methods from [`RuboCop::Cop::Corrector`](https://github.com/rubocop/rubocop/blob/master/lib/rubocop/cop/corrector.rb): `remove_preceding`, `remove_leading`, `remove_trailing`
- [x] Add documentation for:
  - [x] `Config#autocorrect`
  - [x] `Runner#autocorrect`
  - [x] `Source#correct`
  - [x] `Source::Corrector` and its methods
- [x] Add `:nodoc:` directive to `Source::Rewriter::Action` since it is only used internally.

Future PR:

Implement policy configuration for `different_replacements` and `swallowed_insertions` (and maybe also `crossing_deletions`). I didn't realize that `RuboCop::Cop::Corrector` [configures `different_replacements` and `swallowed_insertions` to `:raise`](https://github.com/rubocop/rubocop/blob/master/lib/rubocop/cop/corrector.rb#L20-L25).

### DIFF
--- a/spec/ameba/rule/layout/trailing_blank_lines_spec.cr
+++ b/spec/ameba/rule/layout/trailing_blank_lines_spec.cr
@@ -13,18 +13,18 @@ module Ameba::Rule::Layout
     end
 
     it "fails if there is no blank lines at the end" do
-      s = expect_issue subject, "no-blankline # error: Trailing newline missing"
-      expect_correction s, "no-blankline\n"
+      source = expect_issue subject, "no-blankline # error: Trailing newline missing"
+      expect_correction source, "no-blankline\n"
     end
 
     it "fails if there more then one blank line at the end of a source" do
-      s = expect_issue subject, "a = 1\n \n # error: Excessive trailing newline detected", normalize: false
-      expect_no_corrections s
+      source = expect_issue subject, "a = 1\n \n # error: Excessive trailing newline detected", normalize: false
+      expect_no_corrections source
     end
 
     it "fails if last line is not blank" do
-      s = expect_issue subject, "\n\n\n puts 22 # error: Trailing newline missing", normalize: false
-      expect_correction s, "\n\n\n puts 22\n"
+      source = expect_issue subject, "\n\n\n puts 22 # error: Trailing newline missing", normalize: false
+      expect_correction source, "\n\n\n puts 22\n"
     end
 
     context "when unnecessary blank line has been detected" do
--- a/spec/ameba/rule/layout/trailing_blank_lines_spec.cr
+++ b/spec/ameba/rule/layout/trailing_blank_lines_spec.cr
@@ -13,18 +13,18 @@ module Ameba::Rule::Layout
     end
 
     it "fails if there is no blank lines at the end" do
-      expect_issue subject, "no-blankline # error: Trailing newline missing"
-      expect_correction "no-blankline\n"
+      s = expect_issue subject, "no-blankline # error: Trailing newline missing"
+      expect_correction s, "no-blankline\n"
     end
 
     it "fails if there more then one blank line at the end of a source" do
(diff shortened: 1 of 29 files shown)


src/ameba/spec/expect_issue.cr
```diff
@@ -47,6 +47,8 @@ require "./util"
 module Ameba::Spec::ExpectIssue
   include Spec::Util
 
+  class_property source : Source?
+
```
veelenga, line 51: this may not work well if someday we parallelize the tests

veelenga (commented):
I tested this locally and it is looking great. Thanks for you work @FnControlOption 

I left two small suggestions/comments

src/ameba/spec/expect_issue.cr
FnControlOption: What about returning `source` from `expect_issue`?

veelenga: yeah, this sounds good to me. Also, that will allow using `expect_correction` without `expect_issue` at all, right?

FnControlOption: That is correct (pun intended)

spec/ameba/rule/layout/trailing_blank_lines_spec.cr
```diff
@@ -5,28 +5,26 @@ module Ameba::Rule::Layout
...
     it "passes if source is empty" do
-      source = Source.new ""
-      subject.catch(source).should be_valid
+      expect_no_issues subject, ""
     end
 
     it "fails if there is no blank lines at the end" do
-      source = Source.new "no-blankline"
-      subject.catch(source).should_not be_valid
+      s = expect_issue subject, "no-blankline # error: Trailing newline missing"
```
Sija, line 16:
```suggestion
      source = expect_issue subject, "no-blankline # error: Trailing newline missing"
```

ditto elsewhere

Sija (approved):
🚀

PR #249, opened 2021-11-02
Remove `normalize` parameter from `expect_issue`

FnControlOption:
Also added documentation for #246 and #248

### DIFF
--- a/spec/ameba/rule/lint/debugger_statement_spec.cr
+++ b/spec/ameba/rule/lint/debugger_statement_spec.cr
@@ -22,12 +22,14 @@ module Ameba::Rule::Lint
     end
 
     it "fails if there is a debugger statement" do
-      expect_issue subject, <<-CRYSTAL
+      source = expect_issue subject, <<-CRYSTAL
           a = 2
           debugger
         # ^^^^^^^^ error: Possible forgotten debugger statement detected
           a = a + 1
         CRYSTAL
+
+      expect_no_corrections source
     end
 
     it "reports rule, pos and message" do
--- a/spec/ameba/rule/lint/debugger_statement_spec.cr
+++ b/spec/ameba/rule/lint/debugger_statement_spec.cr
@@ -5,7 +5,7 @@ module Ameba::Rule::Lint
 
   describe DebuggerStatement do
     it "passes if there is no debugger statement" do
-      expect_no_issues subject, %(
+      expect_no_issues subject, <<-CRYSTAL
         "this is not a debugger statement"
         s = "debugger"
 
@@ -18,16 +18,16 @@ module Ameba::Rule::Lint
           end
         end
         A.new.debugger
-      )
+        CRYSTAL
     end
 
     it "fails if there is a debugger statement" do
-      expect_issue subject, %(
-        a = 2
-        debugger
-        # ^{} error: Possible forgotten debugger statement detected
-        a = a + 1
-      )
+      expect_issue subject, <<-CRYSTAL
+          a = 2
+          debugger
+        # ^^^^^^^^ error: Possible forgotten debugger statement detected
+          a = a + 1
+        CRYSTAL
     end
 
     it "reports rule, pos and message" do
--- a/spec/ameba/rule/lint/debugger_statement_spec.cr
+++ b/spec/ameba/rule/lint/debugger_statement_spec.cr
@@ -23,10 +23,10 @@ module Ameba::Rule::Lint
 
     it "fails if there is a debugger statement" do
       source = expect_issue subject, <<-CRYSTAL
-          a = 2
-          debugger
-        # ^^^^^^^^ error: Possible forgotten debugger statement detected
-          a = a + 1
+        a = 2
+        debugger
+        # ^^^^^^ error: Possible forgotten debugger statement detected
+        a = a + 1
         CRYSTAL
 
       expect_no_corrections source
--- a/src/ameba/rule/style/is_a_filter.cr
+++ b/src/ameba/rule/style/is_a_filter.cr
@@ -44,7 +44,9 @@ module Ameba::Rule::Style
       filter_names : Array(String) = %w(select reject any? all? none? one?)
     end
 
-    MSG = "Use `%s(%s)` instead of `%s {...}`"
+    MSG = "Use `%s` instead of `%s`"
+    NEW = "%s(%s)"
+    OLD = "%s {...}"
 
     def test(source)
       AST::NodeVisitor.new self, source, skip: [
@@ -57,6 +59,7 @@ module Ameba::Rule::Style
 
     def test(source, node : Crystal::Call)
       return unless node.name.in?(filter_names)
+      return unless (filter_location = node.name_location)
       return unless (block = node.block)
       return unless (body = block.body).is_a?(Crystal::IsA)
       return unless (path = body.const).is_a?(Crystal::Path)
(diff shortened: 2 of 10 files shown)


straight-shoota:
I'd recommend to keep it simple. Using heredocs is a better solution.

src/ameba/rule/style/is_a_filter.cr
```diff
@@ -77,8 +80,16 @@ module Ameba::Rule::Style
...
 
-      issue_for node.name_location, end_location,
-        MSG % {node.name, name, node.name}
+      old = OLD % node.name
+      new = NEW % {node.name, name}
+      msg = MSG % {new, old}
+      if end_location
+        issue_for(filter_location, end_location, msg) do |corrector|
+          corrector.replace(filter_location, end_location, new)
+        end
```
veelenga, line 89: can this two conditions be merged into a single one? 

```crystal
issue_for(filter_location, end_location, msg) do |corrector|
  corrector.replace(filter_location, end_location, new) if end_location
end
````

Comment (FnControlOption) [reply src/ameba/rule/style/is_a_filter.cr:89]: No, not currently because the presence of a block is used to infer whether an issue can be autocorrected. RuboCop works around this by creating a new `Corrector` instance for every block and then checking if the corrector is empty.

**Review by veelenga — APPROVED** [review]

File: spec/ameba/rule/lint/debugger_statement_spec.cr [hunk 2]
```diff
@@ -18,16 +18,18 @@ module Ameba::Rule::Lint
...
-        a = 2
-        debugger
-        # ^{} error: Possible forgotten debugger statement detected
-        a = a + 1
-      )
+      source = expect_issue subject, <<-CRYSTAL
+          a = 2
+          debugger
+        # ^^^^^^^^ error: Possible forgotten debugger statement detected
+          a = a + 1
```
Sija, line 29: Indenting all of the code to accommodate the difference in comment seems pretty off.
I like the old way with `# ^{}` construct more, to be honest.

Sija (commented):
Other than one comment LGTM

spec/ameba/rule/lint/debugger_statement_spec.cr
FnControlOption: Thanks, I wasn't sure about making this change. RuboCop reserves `^{}` for zero-length cop offenses, but in Ameba it's actually impossible for issues to have zero length since we always add `+ 1` to the end position. So I think it is acceptable to deviate from RuboCop and have `# ^{}` mean "this issue starts at column 1 or 2". However, I'll throw out some ideas because why not?

Option 1:

```
debugger
# ^{} error: Possible forgotten debugger statement detected
```

Option 2: More carets

```
debugger
# ^{}^^^ error: Possible forgotten debugger statement detected
```

Option 3: Replace space after `#` with caret

```
debugger
#^{}^^^^ error: Possible forgotten debugger statement detected
```

Option 4: All carets

```
debugger
#^^^^^^^ error: Possible forgotten debugger statement detected
```

Sija: Thanks for providing a rationale for this change and the comparison with RuboCop! ❤️ 
I like the alternatives, yet I'm missing my favorite, so I'll throw one more option up for discussion :)

Option 5: All carets with space

```cr
debugger
# ^^^^^^ error: Possible forgotten debugger statement detected
```

FnControlOption: There’s also another option that I’ve been reluctant to consider:

Option 6: Last Resort

```cr
debugger # error: Possible forgotten debugger statement detected
```

This is already possible in `master`, we just haven’t used it yet. I’ve been avoiding using it because it has no precision other than the line number that it’s on. However, given all the other options we’ve considered, this is probably the simplest and/or cleanest one. I’m not sure though, so to avoid bikeshedding, I’ll defer to you or @veelenga.

Sija: Yeah, Option 6 seems to be the least accurate, that's no good.

My picks are (ordered by the preference factor):
- Option 5
- Option 2
- Option 1

veelenga: I agree with @Sija. Except "Option 2" because it looks unnecessary to add more carets if there is already that indicator of full-line issue `^{}`

Sija: Shall we go forward with no. 5 then?

veelenga: I'm good for 5 or 1. If you prefer 5, let's do it. Thanks guys 👏

FnControlOption: I've pushed a commit that does both: carets are added if the issue is long enough or `^{}` if not.

Old:

```cr
  debugger
# ^^^^^^^^ error: Possible forgotten debugger statement detected
  require "big"
# ^ error: Duplicated require of `big`
```

New:

```cr
debugger
# ^^^^^^ error: Possible forgotten debugger statement detected
require "big"
# ^{} error: Duplicated require of `big`
```

src/ameba/spec/expect_issue.cr
```diff
@@ -11,51 +13,96 @@ require "./util"
...
 #     issue.message.should eq(
 #       "Avoid chaining a method call on a do...end block."
 #     )
 #
+# Autocorrection can be tested using `expect_correction` after
+# `expect_issue`.
+#
+#     source = expect_issue subject, <<-CRYSTAL
+#         x % 2 == 0
+#       # ^^^^^^^^^^ error: Replace with `Int#even?`.
```
Sija, line 45: Still in the previous format, ditto all below.

FnControlOption:
```cr
expect_issue subject, <<-CRYSTAL, keyword: keyword
    %{keyword} Exception.new(msg)
  # ^{keyword}^^^^^^^^^^^^^^^^^^^ error: Redundant `Exception.new` [...]
  CRYSTAL
```

What about this one?

FnControlOption: Actually... I can just delete two carets

src/ameba/spec/expect_issue.cr
```diff
@@ -11,51 +13,96 @@ require "./util"
...
+#       expect_issue subject, <<-CRYSTAL, keyword: keyword
+#         %{keyword} Exception.new(msg)
+#         # ^{keyword}^^^^^^^^^^^^^^^^^ error: Redundant `Exception.new` [...]
+#         CRYSTAL
+#
+#     %w[has_one has_many].each do |type|
+#       expect_issue subject, <<-CRYSTAL, type: type
+#         class Book
+#           %{type} :chapter, foreign_key: "book_id"
+#           _{type}         # ^^^^^^^^^^^^^^^^^^^^^^ error: Specifying the default [...]
```
Sija, line 84: The last one standing.

```suggestion
#           _{type}           # ^^^^^^^^^^^^^^^^^^^^ error: Specifying the default [...]
```

FnControlOption: No, this one's good already

Sija (approved):

veelenga (approved):

veelenga:
Thank you!

PR #253, opened 2021-11-13
Autocorrect various rules

FnControlOption:
Note: autocorrecting `Style/RedundantBegin` and `Layout/TrailingWhitespace` should be followed by `crystal tool format` to remove extra empty lines around method bodies.

### DIFF
--- a/src/ameba/ext/location.cr
+++ b/src/ameba/ext/location.cr
@@ -14,7 +14,7 @@ module Ameba::Ext::Location
 
   # Seeks to a given *offset* relative to `self`.
   def seek(offset : self) : self
-    if offset.filename.as?(String).try(&.presence) && @filename != offset.filename
+    if offset.filename.as?(String).presence && @filename != offset.filename
       raise ArgumentError.new <<-MSG
         Mismatching filenames:
           #{@filename}
--- a/src/ameba/ext/location.cr
+++ b/src/ameba/ext/location.cr
@@ -1,0 +1,35 @@
+# Extensions to Crystal::Location
+module Ameba::Ext::Location
+  # Returns the same location as this location but with the line and/or column number(s) changed
+  # to the given value(s).
+  def with(line_number = @line_number, column_number = @column_number) : self
+    self.class.new(@filename, line_number, column_number)
+  end
+
+  # Returns the same location as this location but with the line and/or column number(s) adjusted
+  # by the given amount(s).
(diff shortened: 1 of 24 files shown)


src/ameba/ext/location.cr
```diff
@@ -0,0 +1,35 @@
...
+
+  # Returns the same location as this location but with the line and/or column number(s) adjusted
+  # by the given amount(s).
+  def adjust(line_number = 0, column_number = 0) : self
+    self.class.new(@filename, @line_number + line_number, @column_number + column_number)
+  end
+
+  # Seeks to a given *offset* relative to `self`.
+  def seek(offset : self) : self
+    if offset.filename.as?(String).try(&.presence) && @filename != offset.filename
```
Sija, line 17:
```suggestion
    if offset.filename.as?(String).presence && @filename != offset.filename
```

src/ameba/rule/style/method_names.cr
```diff
@@ -46,17 +48,10 @@ module Ameba::Rule::Style
 
     def test(source, node : Crystal::Def)
       return if (expected = node.name.underscore) == node.name
+      return unless (location = name_location(node))
+      return unless (end_location = name_end_location(node))
```
Sija, line 52:
```suggestion
      return unless location = name_location(node)
      return unless end_location = name_end_location(node)
```

ditto elsewhere

FnControlOption: Are there any plans for an Ameba rule for parenthesized assignments similar to RuboCop’s [`Lint/AssignmentInCondition`](https://docs.rubocop.org/rubocop/cops_lint.html#lintassignmentincondition)? The setting `AllowSafeAssignment` defaults to `true`, but would the equivalent Ameba setting default to `false` instead?

veelenga: I don’t think it is planned yet. The assignment in condition is much often that in ruby. We can discuss this in a separate issue

src/ameba/ext/location.cr
FnControlOption: This could be a new Ameba rule 🤔

veelenga (commented):
Looks nice. Need some time to test this locally.

veelenga (approved):
👏

Sija (approved):
LGTM 🚀

PR #254, opened 2021-11-17
Add `Style/GuardClause` rule

FnControlOption:
Credit to RuboCop for original cop [implementation](https://github.com/rubocop/rubocop/blob/master/lib/rubocop/cop/style/guard_clause.rb) and [spec](https://github.com/rubocop/rubocop/blob/master/spec/rubocop/cop/style/guard_clause_spec.rb) (Autocorrect to be added with https://github.com/rubocop/rubocop/pull/10255)

Use a guard clause instead of wrapping the code inside a conditional expression

```cr
# bad
def test
  if something
    work
  end
end

# good
def test
  return unless something

  work
end

# also good
def test
  work if something
end

# bad
if something
  raise "exception"
else
  ok
end

# good
raise "exception" if something
ok

# bad
if something
  foo || raise("exception")
else
  ok
end

# good
foo || raise("exception") if something
ok
```

YAML configuration example:

```yml
Style/GuardClause:
  Enabled: true
```

### DIFF
--- a/spec/ameba/rule/style/guard_clause_spec.cr
+++ b/spec/ameba/rule/style/guard_clause_spec.cr
@@ -342,58 +342,58 @@ module Ameba
         end
         CRYSTAL
     end
-  end
 
-  it_reports_control_expression "return"
-  it_reports_control_expression "next"
-  it_reports_control_expression "break"
-  it_reports_control_expression %(raise "error")
+    it_reports_control_expression "return"
+    it_reports_control_expression "next"
+    it_reports_control_expression "break"
+    it_reports_control_expression %(raise "error")
 
-  context "method in module" do
-    it "reports an issue for instance method" do
-      source = expect_issue subject, <<-CRYSTAL
-        module CopTest
-          def test
-            if something
-          # ^^ error: Use a guard clause (`return unless something`) instead of wrapping the code inside a conditional expression.
-              work
+    context "method in module" do
+      it "reports an issue for instance method" do
+        source = expect_issue subject, <<-CRYSTAL
+          module CopTest
+            def test
+              if something
+            # ^^ error: Use a guard clause (`return unless something`) instead of wrapping the code inside a conditional expression.
+                work
+              end
             end
           end
-        end
-        CRYSTAL
+          CRYSTAL
 
-      expect_correction source, <<-CRYSTAL
-        module CopTest
-          def test
-            return unless something
-              work
-           #{trailing_whitespace}
-          end
-        end
-        CRYSTAL
-    end
-
-    it "reports an issue for singleton methods" do
-      source = expect_issue subject, <<-CRYSTAL
-        module CopTest
-          def self.test
-            if something && something_else
-          # ^^ error: Use a guard clause (`return unless something && something_else`) instead of [...]
-              work
+        expect_correction source, <<-CRYSTAL
+          module CopTest
+            def test
+              return unless something
+                work
+             #{trailing_whitespace}
             end
           end
-        end
-        CRYSTAL
+          CRYSTAL
+      end
 
-      expect_correction source, <<-CRYSTAL
-        module CopTest
-          def self.test
-            return unless something && something_else
-              work
-           #{trailing_whitespace}
+      it "reports an issue for singleton methods" do
+        source = expect_issue subject, <<-CRYSTAL
+          module CopTest
+            def self.test
+              if something && something_else
+            # ^^ error: Use a guard clause (`return unless something && something_else`) instead of [...]
+                work
+              end
+            end
           end
-        end
-        CRYSTAL
+          CRYSTAL
+
+        expect_correction source, <<-CRYSTAL
+          module CopTest
+            def self.test
+              return unless something && something_else
+                work
+             #{trailing_whitespace}
+            end
+          end
+          CRYSTAL
+      end
     end
   end
 end
--- a/spec/ameba/rule/style/guard_clause_spec.cr
+++ b/spec/ameba/rule/style/guard_clause_spec.cr
@@ -1,0 +1,399 @@
+require "../../../spec_helper"
+
+module Ameba
+  subject = Rule::Style::GuardClause.new
+
+  def it_reports_body(body, *, line = __LINE__)
+    rule = Rule::Style::GuardClause.new
+
+    it "reports an issue if method body is if / unless without else" do
+      source = expect_issue rule, <<-CRYSTAL, line: line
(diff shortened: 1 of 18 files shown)


src/ameba/ast/visitors/node_visitor.cr
```diff
@@ -48,14 +48,19 @@ module Ameba::AST
...
     {% end %}
 
     def visit(node)
-      return true unless skip = @skip
-      !skip.includes?(node.class)
+      !skip?(node)
+    end
+
+    private def skip?(node)
+      @skip.try &.includes?(node.class)
```
Sija, line 63:
```suggestion
     !!@skip.try(&.includes?(node.class))
```

src/ameba/rule/style/guard_clause.cr
```diff
@@ -0,0 +1,182 @@
+module Ameba::Rule::Style
+  # Use a guard clause instead of wrapping the code inside a conditional
+  # expression
+  #
+  #     # bad
```
Sija, line 5: Please, wrap the code in triple backticks instead.

src/ameba/rule/style/guard_clause.cr
```diff
@@ -0,0 +1,184 @@
...
+  #
+  # ```
+  # Style/GuardClause:
+  #   Enabled: true
+  # ```
+  class GuardClause < Base
+    include AST::Util
+
+    properties do
+      enabled false
```
veelenga, line 58: why disabled?

veelenga (commented):
Noticed an issue with one case:

```crystal
def method
  if bar = foo
    puts bar
  else
    # nothing
  end
end
```

$ ameba --only Style/GuardClause --fix

```crystal
def method
  return unless bar = foo
    puts bar
  else
    # nothing
  
end
```

src/ameba/rule/style/guard_clause.cr
FnControlOption: I still need to complete the todo below: "check if example is too long for single line"

FnControlOption:
> Noticed an issue with one case

Nice catch. I wonder if the Crystal language designers will accept a PR that adds `Location` fields for `if`, `else`, `begin`, `end` keywords.

veelenga:
We can give it a try. The core team is usually very pleasant and open to reasonable changes.

veelenga:
@FnControlOption is there a way to temporarily workaround the issue without changes in the compiler? Please let me know which stage you would like to merge this at (some WIP and disabled rule is also ok).

FnControlOption:
Is it ok to keep the autocorrection as long as the rule is disabled by default? Or what about adding something like RuboCop’s [`SafeAutoCorrect`](https://docs.rubocop.org/rubocop/usage/auto_correct.html#safe-auto-correct)?

veelenga:
I think it is ok to have autocorrect as long the rule is disabled. Could you please add a pending test for the broken case?

> Or what about adding something like RuboCop’s SafeAutoCorrect

Personally, I would like to avoid this feature, at least at this stage of our project. I think this can easily lead to some inconsistencies (i.e. rule is marked as SafeAutoCorrect but the corner cases keep occurring)

veelenga:
@FnControlOption sorry, just don't want this to be abandoned. If you prefer, I can add a pending test and merge on my own. Please let me know.

FnControlOption:
Oops, sorry for delaying. I’ll try to take a look at this today.

FnControlOption:
Relevant RuboCop issue: [Misleading example for Style/GuardClause](https://github.com/rubocop/rubocop/issues/9227)

> The example in the documentation for `Style/GuardClause` is misleading because mentioned bad and good version don#t match execution.
> 
> It says
> 
> ```
> # bad
> if something
>   foo || raise('exception')
> else
>   ok
> end
> 
> # good
> foo || raise('exception') if something
> ok
> ```
> 
> However both versions behave differently and _bad_ can't be replaced by _good_.
> 
> Imagine `something` being `true` and `foo` returning a truthy value. Bad (original) behavior: `foo` gets called. Good (suggested) behavior: `foo` gets called and `ok` gets called.
> 
> I suggest to remove this example because I don't see a new case that would work which hasn't been mentioned before.

For example, the following is correctable because the if-statement returns the return value:

```rb
# bad
def test1
  if something
    foo || raise('exception')
  else
    ok
  end
end

# good
def test1
  return foo || raise('exception') if something
  ok
end
```

However, the same correction can't be made if the return value is something else:

```rb
def test2
  if something
    foo || raise('exception')
  else
    ok
  end
  bar # this is the return value
end
```

spec/ameba/rule/style/guard_clause_spec.cr
```diff
@@ -0,0 +1,411 @@
...
+
+    pending "does not report an issue if `else` branch is present but empty" do
+      expect_no_issues subject, <<-CRYSTAL
+        def method
+          if bar = foo
+            puts bar
+          else
+            # nothing
+          end
+        end
```
veelenga, line 193:
```suggestion
        end
```

veelenga (approved):

src/ameba/spec/expect_issue.cr
```diff
@@ -161,13 +161,12 @@ module Ameba::Spec::ExpectIssue
...
     lines = code.split('\n') # must preserve trailing newline
     _, actual_annotations = actual_annotations(rules, code, path, lines)
-    unless actual_annotations.to_s == code
-      fail <<-MSG, file, line
+    return if actual_annotations.to_s == code
+    fail <<-MSG, file, line
         Expected no issues, but got:
 
         #{actual_annotations}
         MSG
```
Sija, line 169: Indentation is off:

```suggestion
      Expected no issues, but got:

      #{actual_annotations}
      MSG
```

src/ameba/rule/style/guard_clause.cr
```diff
@@ -0,0 +1,184 @@
...
+      "if"
+    end
+
+    private def accepted_form?(source, node, ending)
+      return true if node.is_a?(Crystal::If) && node.ternary?
+      return true unless cond_loc = node.cond.location
+      return true unless cond_end_loc = node.cond.end_location
+      return true unless cond_loc.line_number == cond_end_loc.line_number
+      return true unless (then_loc = node.then.location).nil? ||
+                         cond_loc < then_loc
```
Sija, line 149:
```suggestion
      return true unless (then_loc = node.then.location).nil? || cond_loc < then_loc
```

Sija (approved):

PR #256, opened 2021-12-31
Add `Lint/DebugCalls` rule

Sija:
Adds support for catching `pp` (et al) calls.


veelenga (commented):
This looks good. However, this is a breaking change for:

1. Integrations (i.e. https://github.com/codacy/codacy-ameba/blob/master/docs/patterns.json#L36)
2. Config files `.ameba.yml`

Alternatively, we could leave the old rule for debugger and add a new one for print-method calls

Sija:
Agreed. I've started like this but went with changing `DebuggerStatement` rule instead.

veelenga (approved):
🎅

Issue #34, opened 2018-01-31
Autocorrect errors


veelenga:
The ability to autocorrect sources using a list of found errors. For example, me as a user seeing this error:

```crystal
def has_user? # PredicateName: Favour method name 'user?' over 'has_user?'
end
```

would love to be able to automatically convert it to:

```crystal
def user?
end
```

Should be doable with `Crystal::Transformer`. Here is a simple example:

https://github.com/crystal-lang/crystal/blob/master/samples/compiler/transformer_example.cr

hugoabonizio:
Awesome! But how to handle the cases where problems comes from third-party shards?

veelenga:
We don't inspect third-party shards. `lib/` directory [is excluded](https://github.com/veelenga/ameba/blob/master/src/ameba/config.cr#L116) from the list of sources.

hugoabonizio:
Actually I was thinking in this case: 
```crystal
if OtherLib.has_something?
  # ...
end
```
But I didn't realised Ameba only checks for methods definitions, not calls. Sorry!

veelenga:
Ah, okay that is an interesting point. For this particular rule (PredicateName) that wouldn't be a case, because it is looking for the method definition, not the method call. So it will not report an error and there will be nothing to autocorrect. But there might be other rules which could do other inspections for third-party calls. Will keep this in mind, thank you.

anicholson:
Oh yes please! I don't want to use a linter that doesn't fix the things it knows how to :)

kostya:
in my project i have ~500 errors: `LargeNumbers: Large numbers should be written with underscores`, replace it by hands is pain.

veelenga:
@kostya @anicholson any help is appreciated :)

veelenga:
Closed by 

https://github.com/crystal-ameba/ameba/pull/248

PR #257, opened 2022-03-05
Disable `Style/LargeNumbers` rule by default

Sija:
Reverts part of e5fb0526e0ead371d50e9cdb2b14a93fcb0ad868

### DIFF
--- a/src/ameba/rule/style/large_numbers.cr
+++ b/src/ameba/rule/style/large_numbers.cr
@@ -28,7 +28,7 @@ module Ameba::Rule::Style
   # ```
   class LargeNumbers < Base
     properties do
-      enabled true
+      enabled false
       description "Disallows usage of large numbers without underscore"
       int_min_digits 5
     end


veelenga (commented):
Could you please elaborate on what is regressed?

Sija:
Setting value. It was disabled before.

veelenga:
This rule was originally disabled because it produced a lot of reports which had to be corrected manually (see https://github.com/crystal-ameba/ameba/issues/34#issuecomment-402743695). But now when it has the automatic correction, I think it makes sense to enable it by default.

Sija:
Main problem is that it produces a lot of false positives (think error codes for instance), therefore I'd keep it disabled by default.

veelenga:
Sorry, could you please provide an example of false positiveness produced by this rule? In my understanding, there should be none (or if there are some, should be easy to correct).

Sija:
Taken from the [mint-lang/mint](https://github.com/mint-lang/mint) repository:

```
Inspecting 1187 files

........................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................F..........................................................................................................................................................................................................................................................................................................................................................................................................

src/lsp/protocol/error_codes.cr:3:28 [Correctable]
[C] Style/LargeNumbers: Large numbers should be written with underscores: -32_700
> ParseError           = -32700
                         ^----^

src/lsp/protocol/error_codes.cr:4:28 [Correctable]
[C] Style/LargeNumbers: Large numbers should be written with underscores: -32_600
> InvalidRequest       = -32600
                         ^----^

src/lsp/protocol/error_codes.cr:5:28 [Correctable]
[C] Style/LargeNumbers: Large numbers should be written with underscores: -32_601
> MethodNotFound       = -32601
                         ^----^

src/lsp/protocol/error_codes.cr:6:28 [Correctable]
[C] Style/LargeNumbers: Large numbers should be written with underscores: -32_602
> InvalidParams        = -32602
                         ^----^

src/lsp/protocol/error_codes.cr:7:28 [Correctable]
[C] Style/LargeNumbers: Large numbers should be written with underscores: -32_603
> InternalError        = -32603
                         ^----^

src/lsp/protocol/error_codes.cr:8:28 [Correctable]
[C] Style/LargeNumbers: Large numbers should be written with underscores: -32_002
> ServerNotInitialized = -32002
                         ^----^

src/lsp/protocol/error_codes.cr:9:28 [Correctable]
[C] Style/LargeNumbers: Large numbers should be written with underscores: -32_099
> ServerErrorStart     = -32099
                         ^----^

src/lsp/protocol/error_codes.cr:10:28 [Correctable]
[C] Style/LargeNumbers: Large numbers should be written with underscores: -32_000
> ServerErrorEnd       = -32000
                         ^----^

src/lsp/protocol/error_codes.cr:11:28 [Correctable]
[C] Style/LargeNumbers: Large numbers should be written with underscores: -32_001
> UnknownErrorCode     = -32001
                         ^----^

src/lsp/protocol/error_codes.cr:12:28 [Correctable]
[C] Style/LargeNumbers: Large numbers should be written with underscores: -32_800
> RequestCancelled     = -32800
                         ^----^

Finished in 269.86 milliseconds
1187 inspected, 10 failures
```

veelenga:
Thi is not false positiveness. It is a suggested way to write large numbers (for better readability). 

However, looking at this now I think it makes more sense to adjust the rule a bit and ignore large numbers with less than 6 digits. The five-digit number still looks readable without any underscores. Your thoughts?

Sija:
> Thi is not false positiveness. It is a suggested way to write large numbers (for better readability).

It is, in a sense that's not a number per se, but an _error code_, so it should be a subject to such formatting rule.

> However, looking at this now I think it makes more sense to adjust the rule a bit and ignore large numbers with less than 6 digits. The five-digit number still looks readable without any underscores. Your thoughts?

We could bump it, sure. There's no guarantee that you won't find 6 digit error codes too, but these are far less common, that's for sure.

veelenga:
Didn't know the error codes are different from numbers 😮 

Ok, let's disable if it brings issues.

veelenga (approved):

PR #258, opened 2022-03-08
Bump default value of `LargeNumbers#int_min_digits` to 6

Sija:
As most of the error codes are smaller than 6 digits
Ref https://github.com/crystal-ameba/ameba/pull/257#issuecomment-1060798162


veelenga (approved):
:+1:

PR #260, opened 2022-04-03
Ignore macro scopes in `ShadowingOuterLocalVar`

### LINKED ISSUE #224: `Lint/ShadowingOuterLocalVar` gives false positives in macro code
Hi,

The following code produce an incorrect warning:
```Crystal
struct Test
  def test
    {% for ivar in @type.instance_vars %}
      {% var_type = ivar %}
    {% end %}

    {% ["a", "b"].map { |ivar| puts ivar } %}
  end
end

Test.new.test
```

Although `ivar` is not defined outside the `for` loop, ameba think it is:

```
Inspecting 1 file

F

t.cr:10:26
[W] Lint/ShadowingOuterLocalVar: Shadowing outer local variable `ivar`
> {% ["a", "b"].map { |ivar| puts ivar } %}
                       ^

Finished in 1.82 milliseconds
1 inspected, 1 failure
```

You can check that `ivar` is actually not defined by adding `{% puts ivar %}` right after the `for` loop: the compiler will complains.
The error from Ameba disappears if we remove the assignation to `var_type`.

Sija:
Closes #224


veelenga (approved):
:+1:

PR #263, opened 2022-04-05
Change `crystal` version constraint to `~> 1.4.0`

Sija:
Uses a more conservative, yet still flexible approach to Crystal version compatibility.

### DIFF
--- a/shard.yml
+++ b/shard.yml
@@ -15,6 +15,6 @@ scripts:
 executables:
   - ameba
 
-crystal: ">= 1.4.0"
+crystal: "~> 1.4.0"
 
 license: MIT


shard.yml
```diff
@@ -15,6 +15,6 @@ scripts:
 executables:
   - ameba
 
-crystal: ">= 1.4.0"
+crystal: "~> 1.4.0"
```
veelenga, line 18: Why do we need to exclude future major releases?

Sija: That's the more conservative approach and one more in line with semver.

straight-shoota: I'm generally not a fan of upper bounded restrictions. But ameba uses internal compiler APIs which have no stability guarantee, so it might be good to be more conservative.
This means of course that every Crystal release needs a new ameba release. And that might not be the most practical solution, considering that breaking changes are actually still quite rare.

Sija: Not _patch_ releases, just the _minor_/_major_ ones.

veelenga (approved):

PR #268, opened 2022-04-21
Fix regression in `AnyInsteadOfEmpty` rule

### LINKED ISSUE #267: `AnyInsteadOfEmpty` rule uses wrong node position
In Lucky, we have a query method `.any?`. This actually runs SQL, so `Performance/AnyInsteadOfEmpty` rule here doesn't apply. 

This code passed on 0.14.4, but fails on 1.0.0
```crystal
TransactionQuery.new
  .receiver(user)
  .any? # ameba:disable Performance/AnyInsteadOfEmpty
```

```
src/serializers/post_serializer.cr:80:24 [Correctable]
[C] Performance/AnyInsteadOfEmpty: Use `!{...}.empty?` instead of `{...}.any?`
> TransactionQuery.new
  ^
```

Maybe solved by https://github.com/crystal-ameba/ameba/pull/265 ?

Sija:
Fixes #267 regression (introduced in [`63a6c73`](https://github.com/crystal-ameba/ameba/commit/63a6c73dc0549d5091016dcc1aa74d4d747269f5#diff-7e35ef9fecf1900c60df38cce289a5265975e26bcbf07f485d84f8af50f87a1dL43-R48) as a part of #253)

### DIFF
--- a/spec/ameba/rule/performance/any_instead_of_empty_spec.cr
+++ b/spec/ameba/rule/performance/any_instead_of_empty_spec.cr
@@ -16,7 +16,7 @@ module Ameba::Rule::Performance
     it "reports if there is any? call without a block nor argument" do
       source = expect_issue subject, <<-CRYSTAL
         [1, 2, 3].any?
-        # ^^^^^^^^^^^^ error: Use `!{...}.empty?` instead of `{...}.any?`
+                # ^^^^ error: Use `!{...}.empty?` instead of `{...}.any?`
         CRYSTAL
 
       expect_correction source, <<-CRYSTAL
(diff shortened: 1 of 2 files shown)


spec/ameba/rule/performance/any_instead_of_empty_spec.cr
```diff
@@ -51,7 +51,7 @@ module Ameba::Rule::Performance
       issue = source.issues.first
 
       issue.rule.should_not be_nil
-      issue.location.to_s.should eq "source.cr:1:1"
+      issue.location.to_s.should eq "source.cr:1:11"
       issue.end_location.to_s.should eq "source.cr:1:14"
       issue.message.should eq "Use `!{...}.empty?` instead of `{...}.any?`"
```
veelenga, line 56: this test can be removed or should be re-worked to the `# ^^^^` notation.

Sija: Nah, I wouldn't touch it since it's the `reports rule, pos and message` test found all over specs, so there's no point in changing just this one without touching all of the others.

veelenga (approved):

PR #281, opened 2022-10-28
Reduce usage of `Object#not_nil!`

Sija:


veelenga (approved):
Lgtm

spec/ameba/ast/variabling/assignment_spec.cr
```diff
@@ -53,8 +53,8 @@ module Ameba::AST
         scope = Scope.new nodes.def_nodes.first
         variable = Variable.new(nodes.var_nodes.first, scope)
         assignment = Assignment.new(nodes.assign_nodes.first, variable, scope)
-        assignment.branch.should_not be_nil
-        assignment.branch.not_nil!.node.class.should eq Crystal::Expressions
+        branch = assignment.branch.should_not be_nil
```
straight-shoota, line 56: While at it, just a minor improvement suggestion: Instead of expecting *not* to be Nil, change the expectation to what *is* expected (I presume `Branch`?):
```suggestion
        branch = assignment.branch.should be_a Branch
```

Sija: I'd leave it as-is, since this specific spec is interested only in non-nilability.

PR #282, opened 2022-10-28
Populate `Rule::Base+.parsed_doc` on compile time

### LINKED ISSUE #270: Missing `--explain` descriptions
When installing ameba via brew on macOS, the output of `--explain` includes the following:

```
...

## DETAILED DESCRIPTION

Error: Error opening file with mode 'r': '/private/tmp/ameba-20220516-71379-vydjhb/ameba-1.0.0/src/ameba/rule/performance/chained_call_with_no_bang.cr': No such file or directory
```

Presumably some [additional files need to be installed](https://docs.brew.sh/Formula-Cookbook#just-moving-some-files) by the [brew formula](https://github.com/veelenga/homebrew-tap/blob/master/ameba.rb#L13).

```
% brew list ameba
/opt/homebrew/Cellar/ameba/1.0.0/bin/ameba
```

Sija:
Fixes #270


veelenga (approved):
Makes sense. Thank you.

PR #288, opened 2022-10-30
Add `Lint/NotNil` rule

### LINKED ISSUE #284: Warning on `#not_nil!` usage
`Object#not_nil!` shouldn't rly be used and is considered as a code smell.
Assigning to a local variable is preferable.
Exceptions as always, can be excluded if desired.

Sija:
Resolves #284

### DIFF
--- a/src/ameba/rule/lint/not_nil.cr
+++ b/src/ameba/rule/lint/not_nil.cr
@@ -1,0 +1,57 @@
+module Ameba::Rule::Lint
+  # This rule is used to identify usages of `not_nil!` calls.
+  #
+  # For example, this is considered a code smell:
+  #
+  # ```
+  # names = %w[Alice Bob]
+  # alice = names.find { |name| name == "Alice" }.not_nil!
+  # ```
+  #
+  # And can be written as this:
+  #
+  # ```
+  # names = %w[Alice Bob]
+  # alice = names.find { |name| name == "Alice" }
+  #
+  # if alice
+  #   # ...
+  # end
+  # ```
+  #
+  # YAML configuration example:
+  #
+  # ```
+  # Lint/NotNil:
+  #   Enabled: true
+  # ```
+  class NotNil < Base
+    include AST::Util
+
+    properties do
+      description "Identifies usage of `not_nil!` calls"
+    end
+
+    NOT_NIL_NAME = "not_nil!"
+    MSG          = "Avoid using `not_nil!`"
+
+    def test(source)
+      AST::NodeVisitor.new self, source, skip: [
+        Crystal::Macro,
+        Crystal::MacroExpression,
+        Crystal::MacroIf,
+        Crystal::MacroFor,
+      ]
+    end
+
+    def test(source, node : Crystal::Call)
+      return unless node.name == NOT_NIL_NAME
+      return unless node.obj && node.args.empty?
+
+      return unless name_location = node.name_location
+      return unless end_location = name_end_location(node)
+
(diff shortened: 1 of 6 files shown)


src/ameba/rule/lint/not_nil.cr
```diff
@@ -0,0 +1,53 @@
...
+  #
+  # ```
+  # Lint/NotNil:
+  #   Enabled: true
+  # ```
+  class NotNil < Base
+    include AST::Util
+
+    properties do
+      description "Identifies usage of `not_nil!` calls."
```
veelenga, line 32: just for consistency, please remove the final dot

```suggestion
      description "Identifies usage of `not_nil!` calls"
```

src/ameba/rule/lint/not_nil.cr
```diff
@@ -0,0 +1,53 @@
...
+      AST::NodeVisitor.new self, source, skip: [
+        Crystal::Macro,
+        Crystal::MacroExpression,
+        Crystal::MacroIf,
+        Crystal::MacroFor,
+      ]
+    end
+
+    def test(source, node : Crystal::Call)
+      return unless node.name == NOT_NIL_NAME && node.obj
```
veelenga, line 48: to ensure this is `Object#not_nil!`, should we check that this call doesn't have any arguments

src/ameba/rule/lint/percent_array.cr
```diff
@@ -42,10 +42,12 @@ module Ameba::Rule::Lint
           start_token = token.dup
         when .string?
           if start_token && issue.nil?
+            # ameba:disable Lint/NotNil
             issue = array_entry_invalid?(token.value, start_token.not_nil!.raw)
           end
```
veelenga, line 47: can this case be handled by adding an extra variable? i.e.

_sorry, hadn't have a chance to give it a try_

```suggestion
          if (start = start_token) && issue.nil?
            issue = array_entry_invalid?(token.value, start.raw)
          end
```

Sija: It could be done, but it changes semantics and makes the code more convoluted. In this case I feel that using `not_nil!` is more apt, kinda exception that proves the rule ;)

Sija: Well, maybe not... I've pushed the changes with local variables.

veelenga (approved):
Lgtm

Blacksmoke16:
@Sija @veelenga Do we think this is really something we should have enabled by default? I don't think a blanket "not_nil! is bad" without taking the context into consideration, which a linter is unable to do, is providing anything real meaningful. 

Maybe for someone new to the language it might sway them to do something different, but then everyone else who understands when/how to use it is forced to either go thru their code and add `ameba:disable Link/NotNil` everywhere, or disable the rule entirely themselves.

Sija:
@Blacksmoke16 I was hesitant myself at the beginning, yet after checking this against several repos (incl. crystal), most of the issues were legitimate refactoring opportunities - and IMO these would be:

- missing assignment to a local var: `@foo.not_nil!`
- missing proper error handling: `foo.not_nil!` vs `raise ... unless foo`

Remaining ones, incl. cases in which using `not_nil!` is the only sensible thing to do, turned out to be a small minority, easily disabled with `# ameba:disable` pragma - and I regard this as a positive thing, since you're giving by-watchers a signal about your intentions.

All in all, using `not_nil!` **is** a code-smell, so all of the cases should be carefully decided upon, IMO even at the cost of potential brief _annoyance_.

PR #292, opened 2022-11-01
Add `Lint/StaticComparison` rule

### LINKED ISSUE #237: New rule: comparison that always fails
As explained in https://github.com/crystal-lang/crystal/issues/10277, Crystal has a default/falback implementation of `==`, which allows any types to be compared, but it just evaluates to false.

Because of this, even comparisons which are useless or, worse, stem from a type error, are allowed and silently fail.

It would be great to add a new check that would be able to identify comparisons which always fail, such as:

```
'a' == "a"
Enum::Member == Int
Enum::Flag & Enum::Flag == Int
13 == GC
...
```

Sija:
Resolves #237

### DIFF
--- a/spec/ameba/rule/lint/static_comparison_spec.cr
+++ b/spec/ameba/rule/lint/static_comparison_spec.cr
@@ -1,0 +1,51 @@
+require "../../../spec_helper"
+
+module Ameba::Rule::Lint
+  subject = StaticComparison.new
+
+  describe StaticComparison do
+    it "passes for valid cases" do
+      expect_no_issues subject, <<-CRYSTAL
+        /foo/ === "foo"
+        "foo" == foo
+        foo == "foo"
+        CRYSTAL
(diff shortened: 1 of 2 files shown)


veelenga (approved):
Looks great 👍

Sija:
One possible edge-case false positive that comes to mind, is a case in which you have dynamic value - usually in containers like array, tuple, hash, etc. Take this for instance:

```cr
a = "foo"
b = "foo"

[a] == [b] # => true
```

This rule will incorrectly flag it (as false, oops), even though this is not a _static_ comparison anymore.

We could either, remove the literals that allows for that, i.e.: range, tuple, named tuple, array, hash and regex.
Or just let it be, since it's pretty uncommon to have 2 literals compared, and usually it signifies a code-smell anyway.

I'm voting for the 2nd option (possibly changing it upon feedback).

PR #297, opened 2022-11-13
Fix the edge case re: free var comparison

### LINKED ISSUE #295: `Lint/LiteralsComparison` false positive on macro generic comparisons
```cr
class Foo(T)
  def initialize(v : T)
    {% raise "T is Nil" if T == Nil %}
  end
end

Foo.new nil
```
Results in:
```
test.cr:3:28
[W] Lint/LiteralsComparison: Comparison always evaluates to false
> {% raise "T is Nil" if T == Nil %}
```

Introduced via #292.

Sija:
Fixes #295

### DIFF
--- a/src/ameba/rule/lint/literals_comparison.cr
+++ b/src/ameba/rule/lint/literals_comparison.cr
@@ -52,23 +52,28 @@ module Ameba::Rule::Lint
     LITERAL_TYPES =
       PRIMITIVE_LITERAL_TYPES + DYNAMIC_LITERAL_TYPES
 
+    # Edge-case: `{{ T == Nil }}`
+    #
+    # Current implementation just skips all macro contexts,
+    # regardless of the free variable being present.
+    #
+    # Ideally we should only check whether either of the sides
+    # is a free var
+    def test(source)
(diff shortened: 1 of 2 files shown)


src/ameba/rule/lint/literals_comparison.cr
```diff
@@ -52,6 +52,22 @@ module Ameba::Rule::Lint
     LITERAL_TYPES =
       PRIMITIVE_LITERAL_TYPES + DYNAMIC_LITERAL_TYPES
 
+    # Edge-case: `{{ T == Nil }}`
+    #
+    # Current implementation just skips all macro contexts,
+    # regardless of the free variable being present.
+    #
+    # Ideally we should only check whether either of the sides
+    # is a free var
```
Sija, line 61: @veelenga any idea how to do that?

veelenga (approved):

src/ameba/rule/lint/literals_comparison.cr
veelenga: Sorry need some time to give it a try.

PR #299, opened 2022-11-13
Refactor `Style/PredicateName` rule

Sija:
- Removes reporting for `has_<name>?` method names, since they're in most cases legit (`is != has`)
- Enabled by default


veelenga (approved):
Makes sense :+1:

PR #301, opened 2022-11-14
Add `Style/RedundantParentheses` rule

### LINKED ISSUE #140: New Rule: Unnecessary Parenthesis
A helpful rule could be to identify unnecessary parenthesis. 

E.x.
```cr
if (1 == 1)
end
```

They are required when assignments are involved, or for order of operations.  

```cr
if (x = @x)
end
```

```cr
if true && (true || false)
end
```

Just a thought.

Sija:
Resolves #140

### DIFF
--- a/spec/ameba/rule/style/redundant_parentheses_spec.cr
+++ b/spec/ameba/rule/style/redundant_parentheses_spec.cr
@@ -1,0 +1,95 @@
+require "../../../spec_helper"
+
+module Ameba::Rule::Style
+  subject = RedundantParentheses.new
+
+  describe RedundantParentheses do
+    {% for keyword in %w(if unless while until) %}
+      context "{{ keyword.id }}" do
+        it "reports if redundant parentheses are found" do
+          source = expect_issue subject, <<-CRYSTAL, keyword: {{ keyword }}
+            %{keyword}   (foo > 10)
+            _{keyword} # ^^^^^^^^^^ error: Redundant parentheses
+              foo
+            end
+            CRYSTAL
+
+          expect_correction source, <<-CRYSTAL
+            {{ keyword.id }}   foo > 10
+              foo
+            end
+            CRYSTAL
+        end
+      end
+    {% end %}
+
+    context "case" do
+      it "reports if redundant parentheses are found" do
+        source = expect_issue subject, <<-CRYSTAL
+          case (foo = @foo)
+             # ^^^^^^^^^^^^ error: Redundant parentheses
+          when String then "string"
+          when Symbol then "symbol"
+          end
+          CRYSTAL
+
+        expect_correction source, <<-CRYSTAL
+          case foo = @foo
+          when String then "string"
+          when Symbol then "symbol"
+          end
+          CRYSTAL
+      end
+    end
+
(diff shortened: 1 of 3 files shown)


src/ameba/rule/style/redundant_parentheses.cr
```diff
@@ -0,0 +1,61 @@
+module Ameba::Rule::Style
+  # A rule that disallows redundant parentheses around control expressions.
+  #
+  # For example, this is considered invalid:
+  #
+  # ```
+  # if (foo = @foo)
+  #   do_something
+  # end
```
veelenga, line 9: could you please elaborate on the decision why this is considered invalid? According to the referenced ticket, parentheses should be required in this case. I believe the idea was to copy the rubocop's check in order to enforce a safe assignment:

https://www.rubydoc.info/gems/rubocop/RuboCop/Cop/Style/ParenthesesAroundCondition

image

Sija: Yeah, I was considering changing the example.

Sija: This PR solves different case, the one you've mentioned is solved by #302.

spec/ameba/rule/style/redundant_parentheses_spec.cr
```diff
@@ -0,0 +1,115 @@
...
+            {{ keyword.id }}   foo > 10
+              foo
+            end
+            CRYSTAL
+        end
+      end
+    {% end %}
+
+    context "case" do
+      it "reports if there is redundant next in expressions" do
```
veelenga, line 27: seems like this description is invalid

spec/ameba/rule/style/redundant_parentheses_spec.cr
```diff
@@ -0,0 +1,115 @@
...
+        end
+      ), "source.cr"
+      subject.catch(s).should_not be_valid
+      s.issues.size.should eq 1
+      issue = s.issues.first
+      issue.rule.should_not be_nil
+      issue.location.to_s.should eq "source.cr:1:4"
+      issue.end_location.to_s.should eq "source.cr:1:14"
+      issue.message.should eq "Redundant parentheses"
+    end
```
veelenga, line 58: this kind of spec was needed just to test the error messages and issue locations. Since we introduced `expect_issue` spec helper this is not relevant anymore. We might need to remove these specs from all other places (which is out of scope of this PR), but for now my suggestion is to not add it to the newly added specs

src/ameba/rule/style/redundant_parentheses.cr
```diff
@@ -0,0 +1,65 @@
...
+    properties do
+      description "Disallows redundant parentheses around control expressions"
+
+      exclude_ternary false
+      exclude_assignments false
+    end
+
+    MSG = "Redundant parentheses"
+
+    # ameba:disable Metrics/CyclomaticComplexity
```
veelenga, line 38: can we reduce the complexity somehow and remove the directive? Maybe the case statement can be moved to a separate method?

veelenga (approved):
LGTM

PR #302, opened 2022-11-14
Refactor `Style/RedundantParentheses` to support enforcing parens around assignments

Sija:
This renames the rule to more fitting `ParenthesesAroundCondition` and replaces `exclude_assignments` option with `allow_safe_assignment`, which if set enforces parens around assignments, and strips them otherwise.

Refs #140


src/ameba/rule/style/redundant_parentheses.cr
```diff
@@ -44,24 +45,34 @@ module Ameba::Rule::Style
...
     def test(source, node : Crystal::If | Crystal::Unless | Crystal::Case | Crystal::While | Crystal::Until)
+      cond = node.cond
+
+      if cond.is_a?(Crystal::Assign) && parenthesized_assignments
+        issue_for cond, MSG_MISSING do |corrector|
+          corrector.insert_before(cond, '(')
+          corrector.insert_after(cond, ')')
+        end
+        return
+      end
```
veelenga, line 63: sorry, i'm a bit confused. The rule is intended to `Disallow redundant parentheses` but it actually enforces them for this case. Maybe better to move this to a separate one?

Sija: Indeed, with this refactor the name of the rule stops fitting, any suggestions?

re: another rule I started with it, but abandoned this approach for couple of reasons:

1. we'd have two -conflicting - rules dealing with the same issue at hand
2. this conflicting rule - even disabled - would trigger CI checks, since CI runs ameba with `--all` switch

Sija: Ok, I've found [`ParenthesesAroundCondition`](https://www.rubydoc.info/github/bbatsov/RuboCop/RuboCop/Cop/Style/ParenthesesAroundCondition) rule in rubocop, which does exactly that. I'm gonna go ahead and just copy the naming.

veelenga (commented):
The names of the files should be changed too. The rest looks good to me. Thanks for dealing with this

veelenga (approved):
Hm actually reviewed it on mobile and haven't noticed that file names are properly changed)) 

LGTM

PR #303, opened 2022-11-15
Add `Lint/LiteralAssignmentsInExpressions` rule

### LINKED ISSUE #294: Identify unintended assignment in condition
This is an idea for a rule that targets the same error type as #163, an unintended assignment in a condition (instead of equality comparison).

I think the only valid use case for an assignment in a condition is for eliminating a `Nil` type from an expression. For that purpose, it's assigned to a local variable and the type restriction excluding `Nil` applies inside the block.

So we could use a couple of indicators which identify this use case:

1. The assignment target is a local variable.
2. The assignment target is used in the block or subsequent parts of the condition.
3. The assigned value is not a constant or literal.
4. The assigned value should not be a local variable (you don't to assign it again, the type restriction would apply on the original already).
5. The conditional expression is not a trailing one (because there's no body with a type restriction; I suppose it *could* be valid if another part of the condition uses the restricted variable, but that's pretty certainly not a good style for a trailing conditional).

There might be more indicators.

The original example from #163 would already violate 2), 3) and 5) 😎 

```cr
var = 1

puts "foo" if var = 1
```

I think this approach would be far superior to using Yoda conditionals for this use case.

Sija:
Resolves (partially) #294

### DIFF
--- a/src/ameba/rule/lint/literal_assignments_in_expressions.cr
+++ b/src/ameba/rule/lint/literal_assignments_in_expressions.cr
@@ -1,0 +1,43 @@
+module Ameba::Rule::Lint
+  # A rule that disallows assignments with literal values
+  # in control expressions.
+  #
+  # For example, this is considered invalid:
+  #
+  # ```
+  # if foo = 42
+  #   do_something
+  # end
+  # ```
+  #
+  # And most likely should be replaced by the following:
+  #
+  # ```
+  # if foo == 42
+  #   do_something
+  # end
+  # ```
+  #
+  # YAML configuration example:
+  #
+  # ```
+  # Lint/LiteralAssignmentsInExpressions:
+  #   Enabled: true
+  # ```
+  class LiteralAssignmentsInExpressions < Base
+    include AST::Util
+
+    properties do
+      description "Disallows assignments with literal values in control expressions"
+    end
+
+    MSG = "Detected assignment with a literal value in control expression"
+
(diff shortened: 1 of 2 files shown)


src/ameba/rule/lint/literal_assignments_in_expressions.cr
```diff
@@ -0,0 +1,64 @@
...
+    }
+
+    DYNAMIC_LITERAL_TYPES = {
+      Crystal::RangeLiteral,
+      Crystal::RegexLiteral,
+      Crystal::TupleLiteral,
+      Crystal::NamedTupleLiteral,
+      Crystal::ArrayLiteral,
+      Crystal::HashLiteral,
+    }
```
veelenga, line 52: can these constants be combined with existing methods? Maybe they could be moved to a single utility/file? 

https://github.com/crystal-ameba/ameba/blob/482597fbf4b5cb18c77aafb634b0f96eca585453/src/ameba/ast/util.cr#L3-L28

straight-shoota:
This only addresses point 3. of #294, so I think it shouldn't resolve that issue.

src/ameba/rule/lint/literal_assignments_in_expressions.cr
Sija: I was thinking about that too, will do.

Sija: I'll rebase this on top #304, once it lands.

veelenga (approved):
LGTM

PR #305, opened 2022-11-15
Few more readability refactors

Sija:
Followup to #300

### DIFF
--- a/src/ameba/formatter/json_formatter.cr
+++ b/src/ameba/formatter/json_formatter.cr
@@ -62,7 +62,6 @@ module Ameba::Formatter
   #   },
   # }
   # ```
-  #
   class JSONFormatter < BaseFormatter
     def initialize(@output = STDOUT)
       @result = AsJSON::Result.new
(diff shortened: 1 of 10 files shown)


src/ameba/formatter/json_formatter.cr
```diff
@@ -115,26 +128,30 @@ module Ameba::Formatter
...
+          message:   message,
+          location:  {
+            line:   location.try &.line_number,
+            column: location.try &.column_number,
+          },
+          end_location: {
+            line:   end_location.try &.line_number,
+            column: end_location.try &.column_number,
+          },
+        }.to_json(json)
```
veelenga, line 143: could you elaborate on the benefits of this change? Do we really need to create an intermediate hash object to make a json object?

veelenga: To be honest #1 is subjective. In my opinion, the previous version is readable. 

For #2 (consistency), the only way to guarantee it is to enforce it through a rule. Otherwise, (as it often happens) a new developer/maintainer comes and introduces discrepancies or tries to enforce other subjective rules re-writing the whole codebase. 

I'm open to do these changes, but just want to be sure the style is preserved in a long term (if we want to follow it).

Sija: I agree with both comments.

> For #2 (consistency), the only way to guarantee it is to enforce it through a rule. Otherwise, (as it often happens) a new developer/maintainer comes and introduces discrepancies or tries to enforce other subjective rules re-writing the whole codebase.

While I do agree that would be the ideal solution, I suggest to leave it for a followup PR.

veelenga (approved):
LGTM

PR #309, opened 2022-11-22
Make boolean rule properties generate query-style methods

Sija:
```cr
class Foo < Ameba::Rule::Base
  properties do
    description "Disallows foos and possibly bars"
    exclude_bar false
  end
end

# This class definition adds 
# - Foo#exclude_bar (before)
# - Foo#exclude_bar? (now)
```


veelenga (approved):
LGTM

PR #310, opened 2022-11-23
Refactor rules’ default group severity handling

Sija:
In addition, this PR changes default severity of `Performance` and `Metrics` groups from `convention` to `warning`.

### DIFF
--- a/src/ameba/config.cr
+++ b/src/ameba/config.cr
@@ -255,9 +255,8 @@ class Ameba::Config
       {% end %}
 
       {% if properties["severity".id] == nil %}
-        {% default = @type.name.starts_with?("Ameba::Rule::Lint") ? "Ameba::Severity::Warning".id : "Ameba::Severity::Convention".id %}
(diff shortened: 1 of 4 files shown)


src/ameba/config.cr
```diff
@@ -267,6 +266,16 @@ class Ameba::Config
     end
 
     macro included
+      GROUP_SEVERITY = {
+        Lint:        Ameba::Severity::Warning,
+        Metrics:     Ameba::Severity::Warning,
+        Performance: Ameba::Severity::Warning,
+      }
```
veelenga, line 273: this is a breaking change that i'm not sure we would like to introduce. The client could rely on `--fail-level Warning` flag and which executed only Lint rules, but now it executes Metrics and Performance in addition

Sija: That's correct, and I believe we should do this BC since these two groups are more serious than a `convention` severity.

veelenga (approved):

PR #311, opened 2022-11-28
Add `Severity#color`

Sija:
This PR aims to improve the DX by colouring issue messages with unique color assigned to issue severity.

### DIFF
--- a/src/ameba/severity.cr
+++ b/src/ameba/severity.cr
@@ -1,3 +1,5 @@
+require "colorize"
+
 module Ameba
   enum Severity
     Error
@@ -14,6 +16,19 @@ module Ameba
       in Error      then 'E'
       in Warning    then 'W'
       in Convention then 'C'
+      end
+    end
+
+    # Returns a color uniquely indicating severity.
+    #
+    # ```
+    # Severity::Warning.color # => Colorize::ColorANSI::Red
+    # ```
+    def color : Colorize::Color
+      case self
+      in Error      then Colorize::ColorANSI::Red
+      in Warning    then Colorize::ColorANSI::Red
+      in Convention then Colorize::ColorANSI::Blue
       end
     end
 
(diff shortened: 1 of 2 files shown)


src/ameba/severity.cr
```diff
@@ -17,6 +19,19 @@ module Ameba
...
 
+    # Returns a color uniquely indicating severity.
+    #
+    # ```
+    # Severity::Warning.color # => Colorize::ColorANSI::Red
+    # ```
+    def color : Colorize::Color
+      case self
+      in Error      then Colorize::ColorANSI::Red
+      in Warning    then Colorize::ColorANSI::Red
```
veelenga, line 30: what about yellow for warning?

Sija: That was also my choice at the beginning, and by this we come to the interesting topic - do our severities are adequate? I mean - we have just a single rule with `:error` severity, and the rest of the rules is using either of the remaining two. So, in majority of the cases we gonna have kinda-Ukrainian flag on the CLI (yellow/blue) :P not that it bothers me, but in my eyes what [Credo](https://raw.github.com/rrrene/credo) does looks way more adequate:

Credo

Refactoring this + the output was going to be (one of) my next PRs actually ;)

veelenga: Okay, that makes sense. My latest suggestion is to do Red for Error and LightRed for Warning. If you still think that's a bad idea, let's follow current implementation.

veelenga (approved):

src/ameba/severity.cr
Sija: Yeah, I still do ;)

- light red looks kinda odd
- we'd use the weaker of the two for majority of the issues
- we'd BC even more
- refactoring is on the way anyway

straight-shoota: Just a small observation: I think ameba shouldn't have an error severity. Everything that's clearly an actual *error* should be rejected by the compiler. A linter doesn't *need* to handle that (except for automatic fixing maybe?).
So `Rule::Lint::Syntax` is the only error severity rule because it catches syntax errors from the parser. I would expect that to stay the only one for said reason.

PR #314, opened 2022-11-30
Add `Style/QueryBoolMethods` rule

Sija:
Reports boolean properties defined with `?`-less variants of `property` / `getter` macros.

```cr
class Person
  property deceased = false
  getter witty = true
end
```

### DIFF
--- a/spec/ameba/rule/style/query_bool_methods_spec.cr
+++ b/spec/ameba/rule/style/query_bool_methods_spec.cr
@@ -1,0 +1,60 @@
+require "../../../spec_helper"
+
+module Ameba::Rule::Style
+  subject = QueryBoolMethods.new
+
+  describe QueryBoolMethods do
+    it "passes for valid cases" do
+      expect_no_issues subject, <<-CRYSTAL
+        class Foo
+          class_property? foo = true
+          property? foo = true
+          property foo2 : Bool? = true
+          setter panda = true
+        end
+
+        module Bar
+          class_getter? bar : Bool = true
+          getter? bar : Bool
+          getter bar2 : Bool? = true
+          setter panda : Bool = true
+
+          def initialize(@bar = true)
+          end
+        end
+        CRYSTAL
+    end
+
+    {% for call in %w[getter class_getter property class_property] %}
+      it "reports `{{ call.id }}` assign with Bool" do
+        expect_issue subject, <<-CRYSTAL, call: {{ call }}
+          class Foo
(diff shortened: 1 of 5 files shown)


src/ameba/ast/util.cr
```diff
@@ -58,6 +58,20 @@ module Ameba::AST::Util
...
+  def path_named?(node, name) : Bool
+    case node
+    when Crystal::Path
+      node_name = node.names.join("::")
+      node_name = "::#{node_name}" if node.global?
+
+      name == node_name
+    else
+      false
+    end
```
veelenga, line 72: case statement seems to be redundant here:

```suggestion
    return false unless node.is_a?(Crystal::Path)

    node_name = node.names.join("::")
    node_name = "::#{node_name}" if node.global?

    name == node_name
```

spec/ameba/rule/style/query_bool_methods_spec.cr
```diff
@@ -0,0 +1,60 @@
...
+          class Foo
+            %{call}   foo : Bool
+            _{call} # ^ error: Consider using '%{call}?' for 'foo'
+
+            def initialize(@foo = true)
+            end
+          end
+          CRYSTAL
+      end
+    {% end %}
```
veelenga, line 58: please add a test that checks if there are two boolean properties and only one is invalid

Sija: I don't think I understand what do you mean.

Sija: I still don't get it... Adjacent properties do not affect each other...

veelenga: > Adjacent properties do not affect each other

That actually highly depends on the implementation. Here is an example:

```diff
diff --git a/src/ameba/rule/style/query_bool_methods.cr b/src/ameba/rule/style/query_bool_methods.cr
index 4bd8c370..0750b143 100644
--- a/src/ameba/rule/style/query_bool_methods.cr
+++ b/src/ameba/rule/style/query_bool_methods.cr
@@ -66,6 +66,7 @@ module Ameba::Rule::Style
             issue_for name_node, MSG % {exp.name, name_node}
           end
         end
+        return
       end
     end
   end
```

if i slightly change the code, the test above fails, but all other pass

veelenga (approved):
LGTM

PR #317, opened 2022-12-08
Further refinements

Sija:


spec/ameba/rule/lint/literals_comparison_spec.cr
```diff
@@ -41,26 +41,13 @@ module Ameba::Rule::Lint
...
-    it "reports if there is a static path comparison evaluating to false" do
-      expect_issue subject, <<-CRYSTAL
-        String == Nil
-        # ^^^^^^^^^^^ error: Comparison always evaluates to false
-        CRYSTAL
-    end
-
     context "macro" do
-      pending "reports in macro scope" do
+      it "reports in macro scope" do
```
veelenga, line 45: could you please add a positive test in the macro scope? i.e. when there are no issues reported

spec/ameba/rule/lint/literals_comparison_spec.cr
```diff
@@ -41,26 +41,13 @@ module Ameba::Rule::Lint
...
           {{ "foo" == "foo" }}
            # ^^^^^^^^^^^^^^ error: Comparison always evaluates to true
           CRYSTAL
       end
-
-      it "passes for free variables comparisons in macro scope" do
-        expect_no_issues subject, <<-CRYSTAL
-          {{ T == Nil }}
-          CRYSTAL
-      end
```
veelenga, line 63: sorry, unclear why is this test removed if that's the catching regression scenario reported in https://github.com/crystal-ameba/ameba/issues/295

src/ameba/ast/variabling/variable.cr
```diff
@@ -174,7 +174,7 @@ module Ameba::AST
         node.accept self
       end
 
-      # @[AlwaysInline]
+      @[AlwaysInline]
```
veelenga, line 177: i have probably missed something, but could you please explain why do we need to ask the compiler to inline this method?

src/ameba/rule/lint/empty_loop.cr
```diff
@@ -58,7 +56,9 @@ module Ameba::Rule::Lint
 
     private def check_node(source, node, loop_body)
       body = loop_body.is_a?(Crystal::Block) ? loop_body.body : loop_body
-      issue_for node, MSG if body.nil? || body.nop?
+      return unless body.nil? || body.nop?
+
+      issue_for node, MSG
```
veelenga, line 61: could you elaborate on why do we favor early return here over the if statement, but do the opposite in the method above 😄

spec/ameba/rule/lint/literals_comparison_spec.cr
Sija: It's because the support for paths has been removed altogether, so all of the `String == Nil` type of comparisons won't be reported at all, neither in macro context nor in regular crystal code.

src/ameba/ast/variabling/variable.cr
Sija: IMO the method body is simple enough to inline it for a small perf improvement.

src/ameba/rule/lint/empty_loop.cr
Sija: Haha, sure. In cases of one simple expression I've opted to attaching them as trailing expressions, whereas in cases there's more statements in the method body I've used the early return. Also, notice there's already `test` method definition with trailing condition.

spec/ameba/rule/lint/literals_comparison_spec.cr
Sija: Sure, done.

veelenga (approved):
LGTM

PR #320, opened 2022-12-13
Add `Lint/UnusedBlockArgument` rule

### LINKED ISSUE #158: New Rule: Block not captured
Anonymous block annotations are present since a bit of time now. Adding `&` shows that the method yields a block, but using `&block` either means that or that the block is captured.

Here is a simple example:
```cr
# Valid
def m(&block)
  block.call 1
end

# Also valid
def m(&block)
  callback = block
end

# Invalid
def m(&block)
  yield
end

# Correct syntax
def m(&)
  yield
end
```

Sija:
Resolves #158

### DIFF
--- a/spec/ameba/rule/lint/unused_block_argument_spec.cr
+++ b/spec/ameba/rule/lint/unused_block_argument_spec.cr
@@ -1,0 +1,110 @@
+require "../../../spec_helper"
+
+module Ameba::Rule::Lint
+  subject = UnusedBlockArgument.new
+
+  describe UnusedBlockArgument do
+    it "doesn't report if it is an instance var argument" do
+      s = Source.new %(
+        class A
+          def initialize(&@callback)
+          end
+        end
+      )
+      subject.catch(s).should be_valid
+    end
+
+    it "doesn't report if anonymous" do
+      s = Source.new %(
+        def method(a, b, c, &)
+        end
+      )
+      subject.catch(s).should be_valid
+    end
+
+    it "doesn't report if argument name starts with a `_`" do
+      s = Source.new %(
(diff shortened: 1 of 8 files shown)


spec/ameba/rule/lint/unused_block_argument_spec.cr
```diff
@@ -0,0 +1,110 @@
...
+      )
+      subject.catch(s).should be_valid
+    end
+
+    it "doesn't report if anonymous" do
+      s = Source.new %(
+        def method(a, b, c, &)
+        end
+      )
+      subject.catch(s).should be_valid
```
veelenga, line 22: shouldn't we use that newer matcher `expect_no_issues`/`expect_issue` instead?

src/ameba/ast/scope.cr
```diff
@@ -143,6 +146,14 @@ module Ameba::AST
       end || variable.used_in_macro?
     end
 
+    # Returns `true` if current scope (or any of inner scopes) references variable,
+    # `false` otherwise.
+    def yields?(check_inner_scopes = true)
```
veelenga, line 151: this description is wrong

spec/ameba/rule/lint/unused_block_argument_spec.cr
Sija: yeah, we should but I was too lazy to rewrite those, xmas & all that jazz, sorry!

Sija: There are still many specs to rewrite anyway...

veelenga (approved):
Lgtm

PR #321, opened 2022-12-17
Add `Lint/MissingBlockArgument` rule

Sija:
Reports yielding method definitions without block argument, as below:

```cr
def foo
  yield 42
end
```

### DIFF
--- a/spec/ameba/rule/lint/missing_block_argument_spec.cr
+++ b/spec/ameba/rule/lint/missing_block_argument_spec.cr
@@ -1,0 +1,42 @@
+require "../../../spec_helper"
+
+module Ameba::Rule::Lint
+  subject = MissingBlockArgument.new
+
+  describe MissingBlockArgument do
+    it "passes if the block argument is defined" do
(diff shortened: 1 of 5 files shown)


spec/ameba/rule/lint/missing_block_argument_spec.cr
```diff
@@ -0,0 +1,28 @@
...
+    end
+
+    it "reports if the block argument is missing" do
+      expect_issue subject, <<-CRYSTAL
+        def foo
+          # ^^^ error: Missing anonymous block argument. Use `&` as an argument name to indicate yielding method.
+          yield 42
+        end
+        CRYSTAL
+    end
```
veelenga, line 26: please add a test for multi-arg example

veelenga, line 26: please add a test for multi-arg example

veelenga: still needs a test for multi-arg method

veelenga (approved):
LGTM

PR #336, opened 2022-12-23
Allow skipping autocorrect from within autocorrect block

FnControlOption:
If an autocorrect block is exited without making any changes, then the issue in question should not be reported as "Correctable". On the one hand, this will reduce the maintenance burden of Ameba rules because `issue_for` will only need to be called once. On the other hand, this may negatively affect performance because each `Issue` temporarily creates an instance of `Corrector` to check if the issue is autocorrectable.

### DIFF
--- a/src/ameba/source/corrector.cr
+++ b/src/ameba/source/corrector.cr
@@ -13,6 +13,11 @@ class Ameba::Source
       @rewriter = Rewriter.new(code)
     end
 
+    # Returns `true` if no (non trivial) update has been recorded
+    def empty?
+      @rewriter.empty?
+    end
+
     # Replaces the code of the given range with *content*.
     def replace(location, end_location, content)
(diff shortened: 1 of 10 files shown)


src/ameba/source/corrector.cr
```diff
@@ -140,58 +145,6 @@ class Ameba::Source
...
-    end
-
-    private def location(node : Crystal::ASTNode)
-      node.location || raise "Missing location"
-    end
-
-    private def end_location(node : Crystal::ASTNode)
-      node.end_location || raise "Missing end location"
-    end
-
```
Sija, line 194: Why? Removing these will increase the amount of boilerplate code.

Sija: Yeah, I was thinking about this the other day. The optimal solution though would be IMO to let `add_issue` figure it out, instead fiddling with it at the call-site.

PR #332, opened 2022-12-21
Update CI definitions + add @Dependabot

Sija:

PR #340, opened 2023-01-03
Fix CI docs workflow

Sija:
Fixes regression introduced in #332

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -25,9 +25,7 @@ jobs:
         run: shards install
 
       - name: Build docs
-        run: |
-          sed -i -e 's:<.*>::g' README.md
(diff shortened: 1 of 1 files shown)


.github/workflows/docs.yml
```diff
@@ -16,32 +20,12 @@ jobs:
...
         uses: actions/checkout@v3
-        with:
-          persist-credentials: false
 
       - name: Install dependencies
         run: shards install
 
       - name: Build docs
-        run: |
-          sed -i -e 's:<.*>::g' README.md
```
veelenga, line 27: sed is still needed in order to remove html tags from README (which are not properly displayed in crystal docs)

Sija: Not anymore, since Crystal got a Markdown parser upgrade.

veelenga (approved):
Lgtm

PR #341, opened 2023-01-04
Remove buggy auto-correction from `Performance/AnyInsteadOfEmpty` rule

Sija:
It was failing in several hard-to-fix edge-cases, so for now I'll remove it in order to avoid producing invalid code.


veelenga (approved):
LGTM

PR #323, opened 2022-12-18
Make `Lint/NotNilAfterNoBang` report calls to `#rindex`

Sija:
Requires Crystal 1.7.0

PR #347, opened 2023-01-28
Use Crystal 1.6.x for ameba v1.4

Sija:
Ameba v1.4 is supposed to work with Crystal < 1.7.
Ameba v1.5 OTOH should support Crystal 1.7+.

This requires re-release.

### DIFF
--- a/shard.yml
+++ b/shard.yml
@@ -15,6 +15,6 @@ scripts:
 executables:
   - ameba
 
-crystal: "~> 1.7.0"
+crystal: "~> 1.6.0"
 
 license: MIT


shard.yml
```diff
@@ -15,6 +15,6 @@ scripts:
 executables:
   - ameba
 
-crystal: "~> 1.7.0"
+crystal: "~> 1.6.0"
```
veelenga, line 18: does this mean Ameba 1.4 will warn when installed on crystal 1.7? Also see

https://github.com/crystal-ameba/ameba/pull/263#discussion_r843114405

Sija: yes, and that's expected, since we **want to** to split the ameba versions based on the crystal version - in this way we'll be able to include features for Crystal 1.7+ only in ameba v1.5.0

veelenga: We have a version that supports Crystal 1.6, it is Ameba v1.3. 

Why do we need another one? Why can't we support Crystal 1.7 in v1.4 now instead of waiting for v1.5 so people can start using it right now? 

Sorry, this looks like a mess and I think we need to formalize the release process.

Sija: What I wanted to say is that the v1.4 should be the **last one** supporting Crystal 1.6.
Then from v1.5 onwards we'll support Crystal 1.7+ with unique features designed to work only there.

Sija: @veelenga would you mind hopping to Twitter to continue this convo?

veelenga: Got it, but are there blockers from supporting Crystal 1.7 right now?

Ameba v1.4 brings a lot of great stuff which i don't want to be blocked on the latest Crystal version `1.7`. Maybe we could extend the range i.e. `>= 1.6.0 and < 1.8` and release a bugfix `v1.4.1` ? On other hand, `v1.4.0` is out, and making more releases with version adjustments may produce more confusion and questions on crystal compatibility.

I would prefer to leave it as is for now unless Ameba v1.4 has compatibility issues with Crystal 1.7. Also, I will prepare a doc for the release process, which will include version bump and compiler compatibility questions, so we will be on the same page in the future.  

PS Sorry for updating your release PR on my own, had to discuss that.

veelenga: @Sija what do you think about expanding the range?

Sija: It doesn't make sense to me, as we need to differentiate between the features available in 1.6 and 1.7+ onwards (#323 for instance).

veelenga: But we should not break the compatibility at this point. I suggest leaving it as is for now. We can release one more minor version (1.5) that supports crystal 1.7 and includes new features (i.e. https://github.com/crystal-ameba/ameba/pull/323)

Sija: Yeah, let it be, we probably should have some version compatibility table (ameba <-> crystal) anyway...

Sija: In this case I'd suggest merging #351 for 1.4.1 release, and then releasing #323 as 1.5.0.

PR #361, opened 2023-03-06
Support hierarchical loading of the config file

### LINKED ISSUE #275: [Feature Request] Search ~/.ameba.yml as a default config for current linux user.
As done like rubocop.

we probably have to merge project config with home config.

Thank you.

Sija:
Files are searched for in the same order as used by [RuboCop](https://docs.rubocop.org/rubocop/configuration.html#config-file-locations).

Resolves #275

### DIFF
--- a/src/ameba/config.cr
+++ b/src/ameba/config.cr
@@ -10,8 +10,30 @@ require "./glob_utils"
 # config.formatter = my_formatter
 # ```
 #
-# By default config loads `.ameba.yml` file in a current directory.
+# By default config loads `.ameba.yml` file located in a current
+# working directory.
 #
+# If it cannot be found until reaching the root directory, then it will be
+# searched for in the user’s global config locations, which consists of a
+# dotfile or a config file inside the XDG Base Directory specification.
+#
+# - `~/.ameba.yml`
+# - `$XDG_CONFIG_HOME/ameba/config.yml` (expands to `~/.config/ameba/config.yml`
+#   if `$XDG_CONFIG_HOME` is not set)
+#
+# If both files exist, the dotfile will be selected.
+#
+# As an example, if Ameba is invoked from inside `/path/to/project/lib/utils`,
+# then it will use the config as specified inside the first of the following files:
+#
+# - `/path/to/project/lib/utils/.ameba.yml`
+# - `/path/to/project/lib/.ameba.yml`
+# - `/path/to/project/.ameba.yml`
+# - `/path/to/.ameba.yml`
+# - `/path/.ameba.yml`
+# - `/.ameba.yml`
+# - `~/.ameba.yml`
+# - `~/.config/ameba/config.yml`
 class Ameba::Config
   include GlobUtils
 
--- a/src/ameba/config.cr
+++ b/src/ameba/config.cr
@@ -24,9 +24,10 @@ class Ameba::Config
     json:     Formatter::JSONFormatter,
   }
 
-  DEFAULT_PATHS = {
-    "~/.ameba.yml",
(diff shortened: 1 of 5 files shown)


src/ameba/config.cr
```diff
@@ -24,7 +24,12 @@ class Ameba::Config
     json:     Formatter::JSONFormatter,
   }
 
-  PATH = ".ameba.yml"
+  DEFAULT_PATHS = {
+    "~/.ameba.yml",
+    "~/.config/ameba/config.yml",
```
veelenga, line 29: i think this must follow [XDG spec](https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html) and use the `${XDG_CONFIG_HOME:-~/.config}/ameba/` folder.

veelenga: something like this?

```cr
XDG_CONFIG_HOME = ENV.fetch("XDG_CONFIG_HOME", "~/.config")

DEFAULT_PATHS = {
  "~/.ameba.yml",
  File.expand_path("#{XDG_CONFIG_HOME}/ameba", home: true)
}
```

Sija: Yeah, no need for `.expand_path` though, since it's being already done l8r on

Sija: Fixed, thanks for spotting! 🙏

veelenga (approved):
LGTM

Sija:
@veelenga Should we create v1.4.3 (I guess the last in 1.4.x branch) to release it? WDYT?

src/ameba/config.cr
```diff
@@ -77,14 +84,33 @@ class Ameba::Config
...
 
+  protected def self.read_config(path) : String?
+    if path
+      return File.exists?(path) ? File.read(path) : nil
+    end
+    path = Path[DEFAULT_PATH].expand(home: true)
+
+    search_paths = path
+      .parents
+      .map! { |search_path| search_path / FILENAME }
```
straight-shoota, line 103: Why not do this lazily while iterating?

straight-shoota (commented):
A more DRY approach for `read_config` would be something like this:

```cr
def read_config(file)
  each_config_path(file) do |path|
    if File.exists?(path)
      return File.read(path)
    end
  end
end

def each_config_path(path)
  yield path if path
  # ...
  search_paths.reverse_each do |search_path|
    yield search_path
  end
  # ...
  DEFAULT_PATHS.each do |default_path|
      yield default_path
    end
end
```

Sija:
@straight-shoota Looks good, I'll do that, thanks!

veelenga (approved):

straight-shoota:
If you want completely different behaviour depending on whether `path` is present, I'd make that more explicit. Probably with two separate overloads, `read_config(String | Path)` and `read_config(Nil)` (or `read_config()`).

src/ameba/config.cr
```diff
@@ -77,14 +106,36 @@ class Ameba::Config
...
-    content = File.exists?(path) ? File.read path : "{}"
+    content = read_config(path) || "{}"
     Config.new YAML.parse(content)
   rescue e
     raise "Config file is invalid: #{e.message}"
   end
 
+  protected def self.read_config(path = nil)
+    if path
+      return File.exists?(path) ? File.read(path) : nil
```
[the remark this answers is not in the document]
veelenga: Yes, I think it should.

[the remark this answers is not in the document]
Sija: I'll create a new issue/pr to handle that.

PR #363, opened 2023-03-08
Skip all config when use with --gen-config.

zw963:


src/ameba/cli/cmd.cr
```diff
@@ -14,7 +14,7 @@ module Ameba::Cli
       raise "Invalid usage: Cannot explain an issue and autocorrect at the same time."
     end
 
-    config = Config.load opts.config, opts.colors?
+    config = Config.load path: opts.config, colors: opts.colors?, no_config: opts.no_config?
```
veelenga, line 17: passing no_config to Config module looks a bit weird. About about renaming no_config to something line "skip_reading_config" or "skip_global_paths"?

Sija: Please avoid adding new CLI switches. This should be working without those.

zw963: > Please avoid adding new CLI switches. This should be working without those.

Okay, i just thought add a new switch will make code more clearly.

spec/ameba/cli/cmd_spec.cr
```diff
@@ -82,6 +82,12 @@ module Ameba::Cli
         c.colors?.should be_true
       end
 
+      it "ignores --config if --gen-config flag passed" do
+        c = Cli.parse_args %w(--gen-config --config my_config.yml)
+        c.formatter.should eq :todo
+        c.skip_reading_config?.should be_true
```
veelenga, line 88: could you please add another test verifying that flag defaults to false?

spec/ameba/cli/cmd_spec.cr
```diff
@@ -47,6 +47,11 @@ module Ameba::Cli
         c.rules?.should eq false
       end
 
+      it "defaults skip_reading_config? flag to false" do
+        c = Cli.parse_args %w(file.cr)
+        c.skip_reading_config?.should eq false
```
veelenga, line 52:
```suggestion
        c.skip_reading_config?.should be_false
```

src/ameba/cli/cmd.cr
```diff
@@ -14,7 +14,7 @@ module Ameba::Cli
       raise "Invalid usage: Cannot explain an issue and autocorrect at the same time."
     end
 
-    config = Config.load opts.config, opts.colors?
+    config = Config.load path: opts.config, colors: opts.colors?, skip_reading_config: opts.skip_reading_config?
```
Sija, line 17: No need to use named arguments here, code is pretty self-explanatory.

src/ameba/cli/cmd.cr
```diff
@@ -51,6 +51,7 @@ module Ameba::Cli
     property except : Array(String)?
     property location_to_explain : NamedTuple(file: String, line: Int32, column: Int32)?
     property fail_level : Severity?
+    property? skip_reading_config = false
```
Sija, line 54: Perhaps sth shorter?

```suggestion
    property? ignore_config = false
```

Sija (commented):
Couple of comments and it's GTG

src/ameba/cli/cmd.cr
zw963: Agreed, self-explanatory after do rename.

src/ameba/cli/cmd.cr
zw963: Consider skip reading config is more accurately for what it does?

spec/ameba/cli/cmd_spec.cr
zw963: Replace all occurs of `eq false` to `be_false`,  don't know if it's appropriate.

Sija (approved):

PR #376, opened 2023-06-09
Fix specs against Crystal nightly

Sija:
Merge once the Crystal nightly release becomes stable.


veelenga:
This looks good. However, it now fails for crystal latest. Maybe we could have some sort of macro, which would allow running tests for specific crystal versions?

veelenga (approved):

Sija:
It's not worth the effort IMO, since specs are run mostly with compatible Crystal versions and cases like this are happening quite rarely.

PR #380, opened 2023-06-15
Add `Lint/DocumentationAdmonition` rule

Sija:


veelenga (commented):
Looking good to me. I would extend the top-level doc explaining the cases when:

1. there is a date 
2. there is a date + time
3. there is an admonition error. Maybe the error message could state what the op must do to fix the error.

src/ameba/rule/lint/documentation_admonition.cr
```diff
@@ -0,0 +1,84 @@
...
+  #   Admonitions: [TODO, FIXME, BUG]
+  #   Timezone: UTC
+  # ```
+  #
+  # ameba:disable Lint/DocumentationAdmonition
+  class DocumentationAdmonition < Base
+    properties do
+      description "..."
+      admonitions %w[TODO FIXME BUG]
+      timezone "UTC"
```
straight-shoota, line 28: I would expect local time zone, not UTC. In the end, it's probably not that relevant really (we're not dealing with a very high granularity of time) but it's odd if the reported time span doesn't correspond to local wall time.

veelenga: Local time will give different results locally vs on CI. I would stick to UTC for consistency.

Sija: Also, timezones might differ from user to user, so it's good to have some consistent point of reference.

straight-shoota: Well I would expect such a date to match the date on my wall calendar. I don't care to correlate it to a universally exact moment in the time line. I would be surprised if ameba told me it wasn't June 15 right now when it is actually June 15 for me.

Practical difference of local vs UTC is probably not very high because it's only a couple of hours and date ranges usually span much bigger time frames.
But, I'd rather have CI and local runs disagree when the CI server has a different time zone than being forced into UTC dates when my mental model is about local days.

Sija: Date on _your_ wall calendar might be different than a date on _someone else's_ wall calendar, mind you.

Sija: Also, since the TZ is configurable it should be easy to grok the reference. Also, that what credo rule does.

veelenga (approved):
LGTM 👏

PR #381, opened 2023-06-15
Add `Lint/Typos` rule

Sija:


src/ameba/rule/lint/typos.cr
```diff
@@ -0,0 +1,97 @@
...
+        }
+
+        # FIXME
+        # if typo.corrections.size == 1
+        #   issue_for typo.location, typo.end_location, message do |corrector|
+        #     corrector.replace(typo.location, typo.end_location, typo.corrections.first)
+        #   end
+        # else
+        #   issue_for typo.location, typo.end_location, message
+        # end
```
veelenga, line 89: what's the problem with corrections? I think it would be very nice having it for this specific rule

Sija: Sure thing, that's why I left it for further work. ATM it doesn't work for some reason - try uncommenting the code and check for yourself. Looks like a corrector bug, or some other strange creature - perhaps you'll be able to spot the error 👀

veelenga: I ended up having `Error: Infinite loop in my_file.cr caused by Lint/Typos`, however, was not able to find the cause. Still looking

Sija: yup, same here

Sija: @FnControlOption Any idea what might be going on?

FnControlOption: Hm… Could it be that there’s an issue being reported that doesn’t get fixed by autocorrect?

Sija: None other then the typos, it seems.

FnControlOption: Right. Do you mind logging the original source at `typo.location` and the expected correction `typo.corrections.first` and pasting the output here?

Sija: Sure, just a sec.

Sija:
```cr
def tpos_frm(userl)
  pp :foo
end
```

```cr
Ameba::Rule::Lint::Typos::Typo(
  @typo="tpos",
  @corrections=["typos"],
  @location={1, 5},
  @end_location={1, 8},
  @path="bug.cr"
)
```

Sija: It seems like off-by-one type of error. I'm not sure where though.

Sija: It magically started to work... 🤯 @veelenga Could you confirm?

veelenga: Hm, yes, it is fixed for me as well.

veelenga (approved):
LGTM 👏

PR #384, opened 2023-06-29
Fix `Style/VerboseBlock` rule to work with binary operations

### LINKED ISSUE #383: correctable error results in invalid crystal
The following line results in an error:

`results.each { |type| type.asset_count = counts[type.id]? || 0_i64 }`
becomes
`results.each(&.asset_count=(counts[type.id]? || 0_i64))`
resulting in a compiler error: `Error: undefined local variable or method 'type'`

```
src/placeos-rest-api/controllers/asset_types.cr:46:15 [Correctable]
[C] Style/VerboseBlock: Use short block notation instead: `each(&.asset_count=(counts[type.id]? || 0_i64))`
> results.each { |type| type.asset_count = counts[type.id]? || 0_i64 }
```
image

Sija:
Resolves #383


veelenga (approved):
LGTM

PR #387, opened 2023-07-10
Add `Performance/MinMaxAfterMap` rule

Sija:
This rule is used to identify usage of `min/max/minmax` calls that follow `map`.

```cr
%w[Alice Bob].map(&.size).min
%w[Alice Bob].map(&.size).max
%w[Alice Bob].map(&.size).minmax
```

And it should be written as this:

```cr
%w[Alice Bob].min_of(&.size)
%w[Alice Bob].max_of(&.size)
%w[Alice Bob].minmax_of(&.size)

### DIFF
--- a/spec/ameba/rule/performance/minmax_after_map_spec.cr
+++ b/spec/ameba/rule/performance/minmax_after_map_spec.cr
@@ -1,0 +1,45 @@
+require "../../../spec_helper"
+
+module Ameba::Rule::Performance
+  subject = MinMaxAfterMap.new
+
+  describe MinMaxAfterMap do
+    it "passes if there is no potential performance improvements" do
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: spec/ameba/rule/performance/minmax_after_map_spec.cr [hunk 1]
```diff
@@ -0,0 +1,45 @@
+require "../../../spec_helper"
+
+module Ameba::Rule::Performance
+  subject = MinMaxAfterMap.new
+
+  describe MinMaxAfterMap do
+    it "passes if there is no potential performance improvements" do
```
veelenga, line 7:
```suggestion
    it "passes if there are no potential performance improvements" do
```

veelenga (approved):
Lgtm

PR #390, opened 2023-07-13
Refactor `--rules` CLI switch output + add `--describe <rule-name>` CLI switch

Sija:
## `--rules` switch refactor

image

## `--describe <rule-name>` switch

image

### DIFF
--- a/src/ameba/cli/cmd.cr
+++ b/src/ameba/cli/cmd.cr
@@ -5,9 +5,6 @@ require "option_parser"
 module Ameba::Cli
   extend self
 
-  private ENABLED_MARK  = "✓".colorize(:green)
-  private DISABLED_MARK = "x".colorize(:red)
-
   def run(args = ARGV)
     opts = parse_args args
     location_to_explain = opts.location_to_explain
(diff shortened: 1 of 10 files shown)


src/ameba/cli/cmd.cr
```diff
@@ -184,13 +202,79 @@ module Ameba::Cli
   end
 
   private def print_rules(config)
```
veelenga, line 204: what about moving this logic to a separate formatter? That way it will be de-coupled from the cli commands and will give an ability to cover it by unit tests

Sija: Thing is, this functionality covers only the rules and not the sources, so it's not rly suited as a formatter. I could do it, although IMO it would be even more clumsy than what we have here.

veelenga: I would try to decouple the "presentation part" from cmd and keep the latter one as simple as possible. 

If it is not suited for the formatter, maybe we could introduce another abstraction or something?

veelenga (approved):
Looks great 👏 
Would be great to have some tests.

Sija:
@veelenga Sure, I've added some.

PR #391, opened 2023-07-15
Make postinstall portable

### LINKED ISSUE #368: Support Windows postinstall
This is a partial resurrection of #230 with a proposal for handling installs on Windows. I believe that until https://github.com/crystal-lang/shards/issues/468 is finally resolved, Ameba's `postinstall` hook should run a Crystal file instead of a Makefile directly, which could then invoke a cmd version of the Makefile for Windows. Something like this:

```crystal
{% if flag?(:win32) %}
  `.\make.cmd build`
{% else %}
  `make bin && make run_file`
{% end %}
```

At least in this way, people attempting to use Ameba on Windows will actually be able to install it instead of getting a confusing error message and having to run the build commands manually.

straight-shoota:
There's no need for copying the executables manually (which happens in both makefile targets `bin` and `run_file`).
Shards' `executables` takes care of that.

The makefile targets could potentially be dropped as well, I don't think there would be other uses case for those.

Using `shards build` directly instead of `make build` improves portability a lot.

The result should be identical to the previous behaviour, except that it works on more platforms (including Windows).
Only practical difference is that the executables are linked instead of copied. This could perhaps be a problem for the source file `ameba.cr`. It should be fine for the binary.

This should basically "just work" almost anywhere because `shards` is always available in the environment. No need for `make` to be available and makefile compatibility doesn't matter either.

Resolves #368

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -30,9 +30,5 @@ bin: build
 	mkdir -p $(SHARD_BIN)
 	cp ./bin/ameba $(SHARD_BIN)
 
-.PHONY: run_file
-run_file:
-	cp -n ./bin/ameba.cr $(SHARD_BIN) || true
-
 .PHONY: test
 test: spec lint
--- a/shard.yml
+++ b/shard.yml
@@ -10,10 +10,11 @@ targets:
 
 scripts:
   # TODO: remove pre-compiled executable in future releases
-  postinstall: make bin && make run_file
+  postinstall: make build
 
 executables:
   - ameba
+  - ameba.cr
 
 crystal: "~> 1.7.0"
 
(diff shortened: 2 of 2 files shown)


straight-shoota:
Actually, installing `crystal.cr` doesn't work on Windows.

But this is a bug in shards: It always appends `.exe` to the filename and then can't find `crystal.cr.exe` 🤦

straight-shoota:
I think we can merge as is. It requires a patched shards on Windows, but postinstall and executables don't currently work there anyway. So this is not making anything worse. And it will start working in the next shards release with crystal-lang/shards#593 merged.

veelenga:
@straight-shoota Wow, thanks for this suggestion. Can we remove `run_file` makefile target as part of this PR? The other targets (including `bin`) could still be useful for local testing.

veelenga (approved):
Tested locally, LGTM.

@Sija please be free to finish the review or merge

Sija (approved):

PR #393, opened 2023-07-24
Raise error when passed invalid config file path

### LINKED ISSUE #362: Providing invalid config file path should raise instead of silently ignoring it
RIght now, it's possible to provide an invalid config file path and ameba just ignores it and uses the default config.
Instead it should raise, and abort the execution altogether.

stufro:
Resolves #362 

I'm new here, let me know if you think this could be done in a different place/better way.


spec/ameba/cli/cmd_spec.cr
```diff
@@ -20,8 +20,14 @@ module Ameba::Cli
 
       %w(-c --config).each do |f|
         it "accepts #{f} flag" do
-          c = Cli.parse_args [f, "config.yml"]
-          c.config.should eq Path["config.yml"]
+          c = Cli.parse_args [f, "shard.yml"]
+          c.config.should eq Path["shard.yml"]
```
stufro, line 24: I had to change the config file we pass to `parse_args` to a file which actually exists so I chose the shard file.

I could create a temp file in the test and throw it away afterwards?

src/ameba/cli/cmd.cr
```diff
@@ -77,6 +77,8 @@ module Ameba::Cli
 
       parser.on("-c", "--config PATH",
         "Specify a configuration file") do |path|
+        raise ArgumentError.new("Unable to find config file #{path}") if !File.exists?(path) && !opts.skip_reading_config?
+
```
veelenga, line 81: Hey and thanks for your contribution. 

can we move this logic to the config itself? 

https://github.com/crystal-ameba/ameba/blob/8c9d234d0b06d945d0cf841dfa0596734acc715c/src/ameba/config.cr#L109-L128

spec/ameba/config_spec.cr
```diff
@@ -84,6 +84,12 @@ module Ameba
         config.formatter.should_not be_nil
       end
 
+      it "raises when custom config file doesn't exist" do
+        expect_raises(Exception, "Config file is invalid: Unable to find config file foo.yml") do
+          Config.load "foo.yml"
+        end
```
veelenga, line 90: can we rephrase the error message re-thrown from `load` method now? Maybe something like

`Config file is invalid` -> `Unable to load config file`

Previously the only reason why it was thrown is invalid YAML. Now the file may not exist here.

src/ameba/config.cr
```diff
@@ -120,7 +120,7 @@ class Ameba::Config
 
   protected def self.read_config(path = nil)
     if path
-      return File.exists?(path) ? File.read(path) : nil
+      return File.exists?(path) ? File.read(path) : raise("Unable to find config file #{path}")
```
veelenga, line 123:
```suggestion
      raise ArgumentError.new("Config file does not exist #{path}") unless File.exists?(path)
      return File.read(path)
```

spec/ameba/config_spec.cr
stufro: sure thing good shout :) 4741c9f

spec/ameba_fixture.yml
```diff
@@ -0,0 +1,2 @@
+Lint/ComparisonToBoolean:
+  Enabled: true
```
veelenga, line 2: could you please move it spec/fixtures/config.yml ?

veelenga (commented):
The last minor comment. The rest of the changes LGTM

veelenga (approved):
LGTM

Sija:
This change forces every glob path under the `Globs` and `Excluded` keys to match, which is a breaking change -IMO unwanted. We need to either address that or revert this PR altogether.

Issue #388, opened 2023-07-11
Providing an invalid file path should raise instead of silently ignoring it


jwoertink:
Sort of related: https://github.com/crystal-ameba/ameba/issues/362

But in this case, if you just tell ameba to check a file that doesn't exist, it returns Inspecting 0 files

```
❯ ./bin/ameba sdfawsdf
Inspecting 0 files



Finished in 10 microseconds
0 inspected, 0 failures
```

I'm thinking of maybe a situation where someone has a CI and types the name of a path wrong, the CI runs and says things are all good, but you don't find out until later that's not the case.

Sija:
Duplicate of #362

veelenga:
@Sija this looks like a separate issue. See the difference between config file path and a file to check.

jwoertink:
Yeah, I originally considered adding to the other issue, but thought maybe they'd be different enough where one is 

```
./bin/ameba -c not_there.yml
```

and this is

```
./bin/ameba not_there.cr
```

I guess another option could be to combine them in to a single issue that's like "passing file paths that don't exist should raise an error in all cases"?

Sija:
You're right, I was trigger happy 😅

stufro:
This is an interesting one to implement because the command line arguments can be a single file or a glob. In [glob_utils.cr#expand](https://github.com/crystal-ameba/ameba/blob/master/src/ameba/glob_utils.cr#L22-L27) the list of arguments are expanded and only file paths which exist are returned from `Dir[glob]`.

This makes it hard to catch non-existent files passed on the command line anywhere else in the code base after the globs have been expanded, which is why I'm thinking to implement this in `GlobUtils#expand`.

My idea is to raise an exception if a particular glob doesn't return any results from `Dir[glob]`. This would raise in both the below examples:

```
> ./bin/ameba not_there.cr
Error: No files found matching not_there.cr

> ./bin/ameba not_there/**/*.cr
Error: No files found matching not_there/**/*..cr
```

I'll make a PR using this approach, but I'm open to feedback on different approaches.

veelenga:
@stufro thanks for taking a look.

This is what I would expect, however, need to understand if there are non-intuitive edge cases.

stufro:
Thanks - #394. I've tried all the edge cases I can think of and it work as I would expect.

PR #394, opened 2023-07-26
Raise error when passed invalid file paths

stufro:
#388 

See [comment](https://github.com/crystal-ameba/ameba/issues/388#issuecomment-1651562708) on #388 for my thought process. 

- I had to change `spec/ameba/cli/cmd_spec.cr` to use real files to prevent the error being thrown there.
  - This also involved changing [line 8](https://github.com/crystal-ameba/ameba/compare/master...stufro:388-raise-on-invalid-file-path?expand=1#diff-94ee7d95bcd546992514ff5718c32bf9b79b665e04feb4c75234b199cb524e8dR8) of that spec to use a config file which disables some rules defined in `spec/spec_helper.cr` which always adds errors.

### DIFF
--- a/spec/ameba/cli/cmd_spec.cr
+++ b/spec/ameba/cli/cmd_spec.cr
@@ -5,7 +5,7 @@ module Ameba::Cli
   describe "Cmd" do
     describe ".run" do
       it "runs ameba" do
-        r = Cli.run %w(-f silent file.cr)
+        r = Cli.run %w(-f silent -c spec/fixtures/config.yml spec/fixtures/passing_ameba.cr)
         r.should be_nil
       end
     end
@@ -43,12 +43,12 @@ module Ameba::Cli
       end
 
       it "defaults rules? flag to false" do
-        c = Cli.parse_args %w(file.cr)
+        c = Cli.parse_args %w(spec/fixtures/passing_ameba.cr)
         c.rules?.should be_false
       end
 
       it "defaults skip_reading_config? flag to false" do
-        c = Cli.parse_args %w(file.cr)
+        c = Cli.parse_args %w(spec/fixtures/passing_ameba.cr)
         c.skip_reading_config?.should be_false
       end
 
@@ -58,7 +58,7 @@ module Ameba::Cli
       end
 
       it "defaults all? flag to false" do
-        c = Cli.parse_args %w(file.cr)
+        c = Cli.parse_args %w(spec/fixtures/passing_ameba.cr)
         c.all?.should be_false
       end
 
@@ -95,35 +95,35 @@ module Ameba::Cli
 
       describe "-e/--explain" do
         it "configures file/line/column" do
-          c = Cli.parse_args %w(--explain src/file.cr:3:5)
+          c = Cli.parse_args %w(--explain spec/fixtures/passing_ameba.cr:3:5)
 
           location_to_explain = c.location_to_explain.should_not be_nil
-          location_to_explain[:file].should eq "src/file.cr"
+          location_to_explain[:file].should eq "spec/fixtures/passing_ameba.cr"
           location_to_explain[:line].should eq 3
           location_to_explain[:column].should eq 5
         end
 
         it "raises an error if location is not valid" do
           expect_raises(Exception, "location should have PATH:line:column") do
-            Cli.parse_args %w(--explain src/file.cr:3)
+            Cli.parse_args %w(--explain spec/fixtures/passing_ameba.cr:3)
           end
         end
 
         it "raises an error if line number is not valid" do
           expect_raises(Exception, "location should have PATH:line:column") do
-            Cli.parse_args %w(--explain src/file.cr:a:3)
+            Cli.parse_args %w(--explain spec/fixtures/passing_ameba.cr:a:3)
           end
         end
 
         it "raises an error if column number is not valid" do
           expect_raises(Exception, "location should have PATH:line:column") do
-            Cli.parse_args %w(--explain src/file.cr:3:&)
+            Cli.parse_args %w(--explain spec/fixtures/passing_ameba.cr:3:&)
           end
         end
 
         it "raises an error if line/column are missing" do
           expect_raises(Exception, "location should have PATH:line:column") do
-            Cli.parse_args %w(--explain src/file.cr)
+            Cli.parse_args %w(--explain spec/fixtures/passing_ameba.cr)
           end
         end
       end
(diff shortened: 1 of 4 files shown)


src/ameba/glob_utils.cr
```diff
@@ -21,6 +21,8 @@ module Ameba
     # ```
     def expand(globs)
       globs.flat_map do |glob|
+        raise ArgumentError.new("No files found matching #{glob}") if Dir[glob].empty?
+
         glob += "/**/*.cr" if File.directory?(glob)
         Dir[glob]
```
veelenga, line 27: the assertion probably needs to be moved one line below to raise on empty folders:

```suggestion
        glob += "/**/*.cr" if File.directory?(glob)
        
        raise ArgumentError.new("No files found matching #{glob}") if Dir[glob].empty?
        
        Dir[glob]
```

veelenga: Also, could you please add a test for this case?

spec/ameba/cli/cmd_spec.cr
```diff
@@ -43,12 +43,12 @@ module Ameba::Cli
       end
 
       it "defaults rules? flag to false" do
-        c = Cli.parse_args %w(file.cr)
+        c = Cli.parse_args %w(spec/fixtures/passing_ameba.cr)
```
veelenga, line 46: let's rename the fixture to `passing_ameba.cr` -> `source.cr`

spec/ameba/glob_utils_spec.cr
```diff
@@ -45,6 +45,12 @@ module Ameba
         subject.expand(["**/#{current_file_basename}", "**/#{current_file_basename}"])
           .should eq [current_file_path]
       end
+
+      it "raises an ArgumentError when the glob doesn't match any files" do
+        expect_raises(ArgumentError, "No files found matching foo/*") do
+          subject.expand(["foo/*"])
+        end
+      end
```
veelenga, line 53: please make sure to add more test cases:

1. missing files in a glob (already added)
2. missing file
3. missing folder
4. empty folder
5. multiple arguments

src/ameba/glob_utils.cr
stufro: I originally had the assertion below `glob += "/**/*.cr" if File.directory?(glob)` however this causes the below error when running `crystal src/cli.cr`.

```
> crystal src/cli.cr                                                                                                                              
globs: ["**/*.cr"]
globs: ["lib"]
Error: No files found matching lib/**/*.cr
```

`expand` is called with `["lib"]` because it is one of the default rejected globs. But since it is empty the exception is raised.

Therefore I went down the route of only raising if the given directory doesn't exists, regardless of whether there are `.cr` files in the directory.

Interested in your thoughts on this.

spec/ameba/glob_utils_spec.cr
stufro: b2069ea - added extra test cases apart from for empty folder, as per above discussion.

src/ameba/glob_utils.cr
veelenga: @stufro I believe now the `!lib` glob has to be added to the default globs conditionally (i.e. excluded if exists). WDYT?

stufro: Hm yes that's a good point. I guess I was assuming that every Crystal project will have lib/ but that's not a safe assumption. 

I'll have a look at conditionally adding !lib to the default globs.

stufro: 7690074 - WDYT..? I'm not super happy with it but it's the best I could come up with

src/ameba/config.cr
```diff
@@ -239,6 +239,13 @@ class Ameba::Config
     end
   end
 
+  private def default_globs
+    DEFAULT_GLOBS.select do |glob|
+      !Dir[glob].empty? ||
+        (glob.starts_with?("!") && !Dir["#{glob[1..-1]}/**/*.cr"].empty?)
+    end
+  end
```
veelenga, line 247: I would probably remove `!lib` part from the DEFAULT_GLOBS and include it conditionally here (as an opposite to excluding it). Also, DEFAULT_GLOBS would probably have to be renamed:

```crystal
MATCHING_SOURCES_GLOB = "**/*.cr"
```

WDYT?

stufro: Thank you! I've pushed eb60b25 using a similar approach to what you described.

Let me know if there's anything else we can improve me merging this :)

spec/ameba/config_spec.cr
```diff
@@ -9,10 +9,20 @@ module Ameba
...
+          config.globs.should eq ["**/*.cr"]
+        end
+
+        it "only sets !lib as a default glob when there are .cr files in lib" do
+          File.touch "lib/shard.cr"
+          yml = YAML.parse "{}"
+          config = Config.new(yml)
+          config.globs.should eq ["**/*.cr", "!lib"]
+          File.delete "lib/shard.cr"
+        end
```
veelenga, line 25: cleanup must be guaranteed in the tests:

```suggestion
        it "only sets !lib as a default glob when there are .cr files in lib" do
          File.touch "lib/shard.cr"
          yml = YAML.parse "{}"
          config = Config.new(yml)
          config.globs.should eq ["**/*.cr", "!lib"]
        ensure
          File.delete "lib/shard.cr"
        end
```

src/ameba/config.cr
```diff
@@ -239,6 +236,12 @@ class Ameba::Config
     end
   end
 
+  private def default_globs
+    [SOURCES_GLOB].tap do |globs|
+      globs.push("!lib") if !Dir["lib/**/*.cr"].empty?
```
veelenga, line 241:
```suggestion
      globs.push("!lib") unless Dir["lib/**/*.cr"].empty?
```

veelenga (commented):
Just a few minor suggestions on my end. Rest LGTM

veelenga (approved):
LGTM

Sija:
This release [broke](https://github.com/crystal-ameba/github-action/actions/runs/7324627583/job/19948387851) ameba's github action :( Also, I believe it's incorrect to raise for invalid paths in `Globs` or `Excluded` - these can be potentially empty or missing and neither of these should be a reason to raise an error and halt the execution of the whole program.

stufro:
Ah sorry about this @Sija. I'm happy to have a look at resolving this but won't likely have time for several weeks so it might be best to revert this PR for now.

Sija:
@stufro No worries, let's revert this PR for now and take care of it l8r.

Issue #406, opened 2023-10-06
Failing spec on crystal master


straight-shoota:
With current master of the compiler, an ameba spec fails:

```
  1) Ameba::AST::Assignment #transformed? returns true if the assignment is transformed by the compiler

       Empty enumerable (Enumerable::EmptyError)
         from /usr/share/crystal/src/enumerable.cr:534:13 in 'first'
         from spec/ameba/ast/variabling/assignment_spec.cr:109:37 in '->'
         from /usr/share/crystal/src/spec/example.cr:45:13 in 'internal_run'
         from /usr/share/crystal/src/spec/example.cr:33:16 in 'run'
         from /usr/share/crystal/src/spec/context.cr:18:23 in 'internal_run'
         from /usr/share/crystal/src/spec/context.cr:342:7 in 'run'
         from /usr/share/crystal/src/spec/context.cr:18:23 in 'internal_run'
         from /usr/share/crystal/src/spec/context.cr:342:7 in 'run'
         from /usr/share/crystal/src/spec/context.cr:18:23 in 'internal_run'
         from /usr/share/crystal/src/spec/context.cr:158:7 in 'run'
         from /usr/share/crystal/src/spec/dsl.cr:212:7 in '->'
         from /usr/share/crystal/src/crystal/at_exit_handlers.cr:14:19 in 'run'
         from /usr/share/crystal/src/crystal/main.cr:64:14 in 'exit'
         from /usr/share/crystal/src/crystal/main.cr:59:5 in 'main'
         from /usr/share/crystal/src/crystal/main.cr:141:3 in 'main'
         from /lib/x86_64-linux-gnu/libc.so.6 in '??'
         from /lib/x86_64-linux-gnu/libc.so.6 in '__libc_start_main'
         from /root/.cache/crystal/crystal-run-spec.tmp in '_start'
         from ???
```

https://github.com/crystal-lang/test-ecosystem/actions/runs/6428749143/job/17456492697

This is most likely caused by https://github.com/crystal-lang/crystal/pull/11597 which changed the AST of packed block parameters.

PR #407, opened 2023-10-06
fix: crystal next compatibility

veelenga:
refs https://github.com/crystal-lang/crystal/pull/11597 
fixes https://github.com/crystal-ameba/ameba/issues/406

cc @straight-shoota

### DIFF
--- a/spec/ameba/ast/variabling/assignment_spec.cr
+++ b/spec/ameba/ast/variabling/assignment_spec.cr
@@ -85,30 +85,5 @@ module Ameba::AST
         assignment.branch.should be_nil
       end
     end
-
-    describe "#transformed?" do
-      it "returns false if the assignment is not transformed by the compiler" do
-        nodes = as_nodes <<-CRYSTAL
-          def method(a)
-            a = 2
-          end
-          CRYSTAL
-        scope = Scope.new nodes.def_nodes.first
-        variable = Variable.new(nodes.var_nodes.first, scope)
-        assignment = Assignment.new(nodes.assign_nodes.first, variable, scope)
-        assignment.transformed?.should be_false
-      end
-
-      it "returns true if the assignment is transformed by the compiler" do
-        nodes = as_nodes <<-CRYSTAL
-          array.each do |(a, b)|
-          end
-          CRYSTAL
-        scope = Scope.new nodes.block_nodes.first
-        variable = Variable.new(nodes.var_nodes.first, scope)
-        assignment = Assignment.new(nodes.assign_nodes.first, variable, scope)
-        assignment.transformed?.should be_true
-      end
-    end
   end
 end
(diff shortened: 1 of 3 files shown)


Sija (commented):
`Lint/UselessAssign` rule is using `transformed?`, just fyi

veelenga:
> `Lint/UselessAssign` rule is using `transformed?`, just fyi

Exactly, I left it for BC. It is not needed anymore for Crystal master and above.

Sija:
@veelenga Could you please explain what do you mean by BC in this particular case? And why it's not longer needed?

veelenga:
@Sija, sure.

In this PR, we've introduced support for Crystal's master version and subsequent versions. The term "BC" refers to the backward compatibility of Crystal (1.9.x) and earlier versions.

We implemented the `transformed?` method to determine if a node represents an assignment transformed by a compiler. To illustrate, in Crystal 1.9.x, the following code:

```crystal
collection.each do |(a, b)|
  puts b
end
```

is transformed before the AST phase to:

```crystal
collection.each do |__arg0|
  a = __arg0[0]
  b = __arg0[1]
  puts(b)
end
```

This transformation leads to false positives since the assignment to the variable `a` becomes redundant. So we had to handle that case "manually" via `#transformed?` method. 

However, in the Crystal master, the block mentioned above remains unaltered, and the resulting AST mirrors the original code:

```crystal
collection.each do |(a, b)|
  puts(b)
end
```

Consequently, for the Crystal master and later versions, the `transformed?` method becomes unnecessary. However, I suggest retaining this method until the next release for the sake of backward compatibility. This ensures that users using both Ameba's master and Crystal 1.9.x will still not encounter these false positives. 

--- 

I hope this revised version provides clarity and a more polished tone.

veelenga:
@Sija, sure.

In this PR, we've introduced support for Crystal's master version and subsequent versions. The term "BC" refers to the backward compatibility of Crystal (1.9.x) and earlier versions.

We implemented the `transformed?` method to determine if a node represents an assignment transformed by a compiler. To illustrate, in Crystal 1.9.x, the following code:

```crystal
collection.each do |(a, b)|
  puts b
end
```

is transformed before the AST phase to:

```crystal
collection.each do |__arg0|
  a = __arg0[0]
  b = __arg0[1]
  puts(b)
end
```

This transformation leads to false positives since the assignment to the variable `a` becomes redundant. So we had to handle that case "manually" via `#transformed?` method. 

However, in the Crystal master, the block mentioned above remains unaltered, and the resulting AST mirrors the original code:

```crystal
collection.each do |(a, b)|
  puts(b)
end
```

Consequently, for the Crystal master and later versions, the `transformed?` method becomes unnecessary. However, I suggest retaining this method until the next release for the sake of backward compatibility. This ensures that users using both Ameba's master and Crystal 1.9.x will still not encounter these false positives.

Sija:
@veelenga Thanks for the thorough explanation ❤️ It all makes sense, let's have this shipped then! :shipit:

Sija (approved):

PR #414, opened 2023-11-09
Add `Naming/AsciiIdentifiers` rule

Sija:


src/ameba/rule/naming/ascii_identifiers.cr
```diff
@@ -0,0 +1,68 @@
...
+  class AsciiIdentifiers < Base
+    include AST::Util
+
+    properties do
+      description "Disallows non-ascii characters in identifiers"
+    end
+
+    MSG = "Identifier contains non-ascii characters"
+
+    def test(source, node : Crystal::Assign)
```
veelenga, line 33: there is also [MultiAssign](https://github.com/crystal-lang/crystal/blob/master/src/compiler/crystal/syntax/ast.cr#L875)

src/ameba/rule/naming/ascii_identifiers.cr
```diff
@@ -0,0 +1,68 @@
...
+
+      if location && end_location
+        check_issue(source, location, end_location, node.name)
+      else
+        check_issue(source, node, node.name)
+      end
+    end
+
+    def test(source, node)
+      if node.responds_to?(:name) && (name = node.name)
```
veelenga, line 51: not sure if that is a good idea to visit all the nodes and invoke responds_to? on them. Can this slow down process?

src/ameba/rule/naming/ascii_identifiers.cr
Sija: Multiple assignment is [not allowed](https://github.com/crystal-lang/crystal/blob/c25640d79bcbf16b015cf9439a862d4626973cf2/src/compiler/crystal/syntax/parser.cr#L150) for constants.

src/ameba/rule/naming/ascii_identifiers.cr
straight-shoota: The meat of `responds_to` is a compile-time operation. At runtime it's just a type filter (like `if x.is_a?(Foo)`).

veelenga (approved):
LGTM

src/ameba/rule/naming/ascii_identifiers.cr
Sija: Actually, I'll extend these checks to all identifiers, not only constants.

PR #419, opened 2023-11-10
Add `Naming/BlockParameterName` rule

Sija:


src/ameba/rule/naming/block_parameter_name.cr
```diff
@@ -0,0 +1,56 @@
...
+  #   AllowNamesEndingInNumbers: true
+  #   AllowedNames: [_, e, i, j, k, v, x, y, ex, io, ws, tx, id, k1, k2, v1, v2]
+  #   ForbiddenNames: []
+  # ```
+  class BlockParameterName < Base
+    include AST::Util
+
+    properties do
+      description "Disallows non-descriptive block parameter names"
+      min_name_length 1
```
veelenga, line 31: just curious if the default value can be increased here to 3. I believe rubocop enforces block args to have a min length of 3, any benefits of this?

Sija: I agree, I've forgot to change it.

Sija: Just ftr, Rubocop has default set to `1`.

veelenga (approved):
LGTM

PR #420, opened 2023-11-10
Add `Lint/SpecFilename` rule

Sija:
As per Crystal guide on [testing](https://crystal-lang.org/reference/1.10/guides/testing.html#running-specs).


veelenga (approved):
LGTM

PR #422, opened 2023-11-12
Make it easier to add issues for nodes with name location preference

Sija:


Sija:
@veelenga This PR fixes CI, btw

straight-shoota:
Have you considered a different method name (such as `add_name_issue`)? I would argue that's a more explicit separation of behaviour and better than a parameter.

Sija:
@straight-shoota I did considered some, yet it was the best I could come up with atm. I prefer named parameter, since it's a preference only, I see no point of having a dedicated method for it.

straight-shoota:
I would say the callsite is more succinct and easier to understand what it means. `prefer_name_location` is named after an internal detail, but `naming_issue_for` describes the purpose.

```cr
naming_issue_for node, MSG
#vs
issue_for node, MSG, prefer_name_location: true
```

veelenga:
I agree with @straight-shoota that `naming_issue_for` looks more understandable. Also, that amount of named params in the `issue_for` method makes it much more complicated.

Sija:
Yeah, I agree, although in this case the parameter indicates a _preference_, not an unconditional behavior, like I've understood you've suggested. Also, neither of the names is properly describing the logic behind it.

Sija:
In this particular context all proposed names: i.e. `add_name_issue` and `naming_issue_for` are IMO too vague and unclear. `name_location` is a _term_ already used, therefore familiar. Also, as I mention before, this is a setting which modifies the original behaviour, and so in my opinion is not suited for extraction to a different method.

src/ameba/rule/lint/missing_block_argument.cr
```diff
@@ -36,10 +34,7 @@ module Ameba::Rule::Lint
     def test(source, node : Crystal::Def, scope : AST::Scope)
       return if !scope.yields? || node.block_arg
 
-      return unless location = node.name_location
-      end_location = name_end_location(node)
-
-      issue_for location, end_location, MSG
+      issue_for node, MSG, prefer_name_location: true
```
veelenga, line 37: does this change the behaviour? Previously if name_location was empty it returned but now it uses a node's location as a fallback?

Sija: Indeed, that was a fix on my part, since the end location is not required for an issue.

veelenga (approved):
LGTM

PR #425, opened 2023-11-14
Prepare release 1.6.0

Sija:


Sija:
@veelenga ping

shard.yml
```diff
@@ -1,21 +1,22 @@
...
-  # TODO: remove pre-compiled executable in future releases
   postinstall: shards build -Dpreview_mt
 
+# TODO: remove pre-compiled executable in future releases
 executables:
   - ameba
   - ameba.cr
 
-crystal: "~> 1.9.0"
+crystal: ~> 1.9
```
veelenga, line 20: since we drop the compatibility with crystal 1.9 below, shouldn't we bump it to 1.10?

Sija: We do? If yes, I missed that.

veelenga: We remove the code that ensures BC for crystal <= 1.9

https://github.com/crystal-ameba/ameba/pull/425/files#diff-3c764e731960f30a76c6f13b0387d5494855d0f52c8fa1a66df17637e3450680L79-L106

veelenga: Which is fine, however, I expected to bump the version here

```suggestion
crystal: ~> 1.10
```

Sija: Done 🧹

veelenga (approved):
LGTM :shipit:

PR #428, opened 2023-11-17
Revert "Fix `Performance/ExcessiveAllocations` to exclude `each` call…

Sija:
…s without a block"

This reverts commit 29e29b8e1d04b9126fb692e2da930a6536d3bceb.

---

This change turned out to be incorrect.


veelenga (commented):
Could you please help me understand the problem to be solved and why this is reverted?

Sija:
Sure, the commit in question restricted the rule to report only calls with blocks, which is incorrect, since the same optimization can be applied to iterators (same calls without a block)

veelenga (approved):

PR #430, opened 2023-12-14
Report unused type declarations in `Lint/UselessAssign` rule

### LINKED ISSUE #429: Recognize unused variable from type declaration
`Lint/UselessAssign` rule reports local variables that are defined per assignment but never used.

```cr
a = "" # Lint/UselessAssign: Useless assignment to variable `b`
```

Type declarations also declare a variable yet do not assign to the variable. This should even more reason to report it.

For the following program, ameba should report a similar issue:

```
a : String
```

This came up in https://github.com/crystal-lang/crystal/issues/14058

Sija:
Resolves #429


Sija:
As type declarations are often used as part of DSLs and within accessor macros, I see no reliable way to mark them as unused assignments.

Sija:
I'm not happy with these few added `# ameba:disable Lint/UselessAssign` pragmas, yet I believe it's a sensible price to pay for having this bug fixed.

@veelenga Any idea how this could be optimized?

Sija:
@veelenga Ready for review.

veelenga:
It's a tough one. Need more time to review.

Sija:
I'm thinking of adding an option to exclude type declarations - in case someone would use 'em in their DSLs, WDYT @veelenga?

veelenga (approved):
LGTM. No comments/suggestions on my end, good work

PR #437, opened 2023-12-28
add error to allowed exception variable names

### LINKED ISSUE #433: exception names are too limited
Currently the default ameba options seem too limited. `%w[e ex exception]`
I feel like this should include version common names like `error` but also it should probably add something like `|| begins_with?("#{allowed}_")` so nested exceptions can be handled like `ex_parse` `ex_io` etc

Also `e` is not very descriptive and breaks other rules like descriptive block naming.

stakach:
resolves #433


veelenga:
@stakach thanks for the PR. There is a test to be fixed.

Sija:
Fixed by #438

PR #438, opened 2023-12-28
Add `error` to the `RescuedExceptionsVariableName#allowed_names`

### LINKED ISSUE #433: exception names are too limited

Sija:
Supersedes #437
Resolves #433


veelenga (approved):
LGTM

PR #443, opened 2024-01-10
Do not report type declarations within `lib` definitions

### LINKED ISSUE #442: `Lint/UselessAssign` reports type declarations within `lib` definitions
Code below triggers an issue:

```cr
lib LibFoo
  struct Foo
    a : String?
  end
end
```

Sija:
Resolves #442


veelenga (approved):
Lgtm

PR #445, opened 2024-01-14
Enhance `Makefile`

straight-shoota:
I figured the `Makefile` could use some love after it failed me in some capacities.

* `install` recipe does not rebuild binary
* Add `help` target and documentation
* Add several config variables
* Add sources as dependencies for (in-)validation

Based on template https://gist.github.com/straight-shoota/275685fcb8187062208c0871318c4a23

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,34 +1,92 @@
+.POSIX:
+all:
+
+# Recipes
+
+## Build ameba
+##   $ make
+## Run tests
+##   $ make test
+## Install ameba
+##   $ sudo make install
+
+-include Makefile.local # for optional local options
+
+BUILD_TARGET ::= bin/ameba
+
+DESTDIR ?=          ## Install destination dir
+PREFIX ?= /usr/local## Install path prefix
+BINDIR ?= $(DESTDIR)$(PREFIX)/bin
+
+# The crystal command to use
 CRYSTAL_BIN ?= crystal
+# The shards command to use
 SHARDS_BIN ?= shards
-PREFIX ?= /usr/local
+# The install command to use
+INSTALL_BIN ?= /usr/bin/install
+
 SHARD_BIN ?= ../../bin
 CRFLAGS ?= -Dpreview_mt
 
+SRC_SOURCES ::= $(shell find src -name '*.cr' 2>/dev/null)
+DOC_SOURCE   ::= src/**
+
+.PHONY: all
+all: build
+
 .PHONY: build
-build:
+build: ## Build the application binary
+build: $(BUILD_TARGET)
+
+$(BUILD_TARGET): $(SRC_SOURCES)
 	$(SHARDS_BIN) build $(CRFLAGS)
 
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -1,34 +1,92 @@
...
+all:
+
+# Recipes
+
+## Build ameba
+##   $ make
+## Run tests
+##   $ make test
+## Install ameba
+##   $ sudo make install
```
Sija, line 11: This list IMO should be either completed to match the targets 1:1, or dropped altogether.

straight-shoota: The idea of the recipe section is to highlight the most common commands and show some examples for how to use them. I suppose the options are very straightforward, so it might not be super necessary (compared to `make help` of other projects like [Crystal](https://github.com/crystal-lang/crystal/blob/master/Makefile) and [Shards](https://github.com/crystal-lang/shards/blob/master/Makefile)).
I think it's nevertheless helpful to have some recipes to start with.
But it certainly doesn't make sense to list *all* targets here.

Sija: Makes sense, thanks :)

Sija (approved):

veelenga (approved):
LGTM 👏

Sija:
@straight-shoota This PR broke my local setup (macOS 12.7) :(

`make` -> `make: Nothing to be done for 'all'.`
`make build` -> `make: Nothing to be done for 'build'.`

straight-shoota:
Are you sure any of the source files changed since you last ran `make`?
This patch added the entire source tree as a dependency of the `build` recipe. So it won't rebuild if the source hasn't changed.
You can force a rebuild with `make -B`.

You could also try `make -d` to get some debug information on which dependencies were considered and what lead to not building.

straight-shoota:
Hm, looks like make doesn't recognize the prerequisites (`SRC_SOURCES`) at all. 

Maybe try this diff to see if there are any error siwht the find command? Would be very strange though...
```diff
@@ -28,7 +28,7 @@ INSTALL_BIN ?= /usr/bin/install
 SHARD_BIN ?= ../../bin
 CRFLAGS ?= -Dpreview_mt

-SRC_SOURCES ::= $(shell find src -name '*.cr' 2>/dev/null)
+SRC_SOURCES ::= $(shell find src -name '*.cr')
 DOC_SOURCE   ::= src/**
```

Sija:
Nope, doesn't work. Also see:

```sh
λ Shabdiz → shell find src -name '*.cr'                                                                                                                           
zsh: command not found: shell
```

straight-shoota:
`$(shell xxx)` is a make function: https://www.gnu.org/software/make/manual/html_node/Shell-Function.html

It means run `xxx` in the shell.

So could you try if `find src -name '*.cr'` works for you? I don't know how that could fail 🤷 It should print the paths of all source files.

Sija:
Yeah, that works just fine.

PR #450, opened 2024-01-17
Exclude reporting type declarations passed as call arguments

### LINKED ISSUE #447: `Lint/UselessAssign` reports macros accepting `Crystal::Macros::TypeDeclaration`
```crystal
macro test(type_decl)
  def {{type_decl.var.id}} : {{type_decl.type}}
    "foo"
  end
end

test name : String
```

```sh
[W] Lint/UselessAssign: Useless assignment to variable `name`
> test name : String
```

This should probably not be flagged since Ameba can't know what the custom macro is doing.

Sija:
Resolves #447


veelenga (approved):
LGTM

PR #460, opened 2024-04-17
Make sure we only return files from `GlobUtils#expand` method

### LINKED ISSUE #459: Ameba trips on folder named `*.cr`
When ameba discovers source files, it considers every path ending in `.cr` as a source file, even if it's not a file but a directory:

```console
$ mkdir -p test/foo.cr
$ ameba test
Error: Error reading file (test/foo.cr): Is a directory
```

This needs a fix similar to https://github.com/crystal-lang/crystal/pull/13760

You'll see this error when running ameba in the Crystal repo which contains a folder ending in `.cr` since fixing that bug.

Sija:
Resolves #459


veelenga (approved):
LGTM. Would be nice to have a test as well

PR #466, opened 2024-08-19
Add cli flag to read from STDIN

nobodywasishere:
This will be useful for IDE stuff as I won't need to write source code to a temp file.

### DIFF
--- a/src/ameba/cli/cmd.cr
+++ b/src/ameba/cli/cmd.cr
@@ -16,6 +16,7 @@ module Ameba::Cli
 
     config = Config.load opts.config, opts.colors?, opts.skip_reading_config?
     config.autocorrect = autocorrect
+    config.stdin_filename = opts.stdin_filename
 
     if globs = opts.globs
       config.globs = globs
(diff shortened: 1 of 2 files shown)


src/ameba/cli/cmd.cr
```diff
@@ -59,6 +60,7 @@ module Ameba::Cli
     property describe_rule : String?
     property location_to_explain : NamedTuple(file: String, line: Int32, column: Int32)?
     property fail_level : Severity?
+    property stdin_filename : String? = nil
```
Sija, line 63:
```suggestion
    property stdin_filename : String?
```

nobodywasishere:
Would help implement https://github.com/crystal-ameba/vscode-crystal-ameba/issues/89

Sija (approved):

straight-shoota:
A common indicator to read from stdin is the filename `-` (for example: `crystal tool format -`).
`--stdin-filename` could only provide an optional name for the stdin file.

nobodywasishere:
All of the crystal tools (except `format`) and `build`/`run` use the `--stdin-filename` flag to read from STDIN, I was matching that functionality.

veelenga (approved):
lgtm

Issue #470, opened 2024-09-26
Restructure Ameba Distribution


devnote-dev:
I recently read an issue involving Ameba; a contributor stated they didn't believe it should be installed and compiled as a development dependency, but rather distributed and installed like any other external tool. I (also recently, coincidentally) encountered an odd issue when running Ameba in one of my projects: I have several folders that link to `PATH` so running `ameba` can call any of them which can be any version and return varying results. And that's exactly what happened 😄 

Now, to remove these binaries wouldn't necessarily solve the issue as different projects use different versions of Ameba, some of which require refactoring to support newer versions. I also can't have multiple versions of Ameba available in `PATH` because of the aforementioned issue. But what if I could have one version of Ameba _with multiple versions?_

I propose that all rules be bound to a version requirement and introduce a `--version` flag (or command, whatever works best) that accepts a valid Ameba version. Running `ameba --version 1.5.0` would only enable rules matching the `< 1.5.0` requirement. Running `ameba` would default to the latest available version, that is, the version of that binary. This process is similar to Rust editions (if I've understood that correctly) and would remove the need to depend on multiple local versions.

Alongside this, I propose that Ameba is released in official distribution channels and discourage the current practice of local development installations with Shards (perhaps a compile time warning, similar to how Crystal's `i_know_what_im_doing` flag works).

straight-shoota:
I completely agree. But the ameba version should generally be recorded in `.ameba.yml`. There's already a discussion about keeping the ruleset consistent in #448.

You're totally right that in order to remove ameba from an implicit `postinstall`, we need to improve and expand its distribution channels. It needs to be easy to get a recent version installed.

There are currently couple of packages available in the distribution package repos of Alpine Linux, AUR and nixpkgs (https://repology.org/project/ameba/versions).

devnote-dev:
> But the ameba version should generally be recorded in `.ameba.yml`.

I hadn't even thought about that but it's perfect! 👍

As for distribution, I believe that providing compiled binaries as part of each release would greatly support this. I'm not sure how practical it would be to provide a couple previous versions' releases, but it's definitely viable going forward.

straight-shoota:
> I'm not sure how practical it would be to provide a couple previous versions' releases, but it's definitely viable going forward.

If binaries are attached to GitHub releases for example, earlier versions would automatically be available.
However I don't think there should be much need for installing older versions if newer ones are able to emulate the configuration of previous versions.

devnote-dev:
[...]

Issue #307, opened 2022-11-18
Add ability to work with a rule subset


veelenga:
More information is here:

https://forum.crystal-lang.org/t/linting-the-crystal-repository/5118/2

Sija:
That was implemented in #471

PR #471, opened 2024-10-07
Implement rule versioning

### LINKED ISSUE #448: Avoid annoyance from new rules
Quoting @Blacksmoke16 from https://github.com/crystal-ameba/ameba/issues/447#issuecomment-1892747306

> It just feels less than ideal that new rules keep being added that cannot be actually robust without semantic analysis, or are more subjective in general. Which ends up forcing me to either disable them inline, globally, or conform to the suggestions on every minor release.
>
> I still think some/most of these should be disabled by default, or made optional via some sort of opinionated style https://github.com/crystal-ameba/ameba/pull/112.

I think this is quite similar to the considerations about the Crystal formatter rolling out new formats (c.f. https://github.com/crystal-lang/crystal/issues/13002).
The formatter is tied to Crystal releases. Ameba on the other hand has more freedom about chosing when to release new rules.
How exactly that can be utilized, is up for discussion.

But in general, I think it would be very helpful to avoid required user interaction every time a new rule is added. That's especially if the rule has a high rate of false positives (such as `Lint/UselessAssign`). That just keeps annoying users.

Maybe for a project with existing ameba configuration, all rules that have been added since the version that generated the ameba config could be disabled by default? I.e. if I have a config from ameba 1.6, ameba will only consider rules from 1.6 until I explicitly upgrade to a new version and ruleset.

Sija:
- Adds `Version` property to `.ameba.yml` (optional)
- Adds `Rule::Base#since_version` (optional)
- Adds `--up-to-version <VERSION>` CLI switch
- Adds `--rule-versions` CLI switch

image

Resolves #448
Refs #470, #307


devnote-dev (approved):

devnote-dev:
`ameba -u` working great! I have a somewhat long list of rules and the versions they were added in if you need it, it goes back as far as 1.3.0. All that remains is implementing the warning message. A compile time flag seems best here, I'm not picky about any particular name for it.

Sija:
@devnote-dev I'll go ahead and finalize this PR, merge it and then let's move onto the next points in the following PRs.

veelenga (approved):
lgtm 👏🏻

src/ameba/presenter/rule_versions_presenter.cr
```diff
@@ -0,0 +1,12 @@
+module Ameba::Presenter
+  class RuleVersionsPresenter < BasePresenter
+    def run(rules)
+      versions =
+        rules.compact_map(&.since_version).sort!.uniq!
+
+      versions.each do |version|
+        output.puts "- %s" % version.to_s.colorize(:green)
```
devnote-dev, line 8:
```suggestion
        output << "- " << version.colorize(:green) << '\n'
```

src/ameba/presenter/rule_presenter.cr
```diff
@@ -1,14 +1,24 @@
...
-      output.puts
       output_title "Rule info"
-      output_paragraph "%s of a %s severity [enabled: %s]" % {
+
+      info = <<-INFO
+        Name:           %s
+        Severity:       %s
+        Enabled:        %s
+        Since version:  %s
+        INFO
```
devnote-dev, line 11: Couldn't we make use of `String.build` here then pass it to `output_paragraph`?

src/ameba/config.cr
```diff
@@ -100,6 +104,9 @@ class Ameba::Config
     @excluded = load_array_section(config, "Excluded")
     @globs = load_array_section(config, "Globs", DEFAULT_GLOBS)
 
+    if version = config["version"]?.try(&.as_s).presence
```
devnote-dev, line 107:
```suggestion
    if version = config["Version"]?.try(&.as_s).presence
```

src/ameba/presenter/rule_presenter.cr
Sija: I see no reason to do so.

devnote-dev (approved):

spec/ameba/presenter/rule_versions_presenter_spec.cr
```diff
@@ -0,0 +1,34 @@
...
+        versions = rules
+          .sort_by { |rule| rule.since_version || missing_version }
+          .group_by(&.since_version)
+
+        versions.each do |version, version_rules|
+          output.should contain (version || "N/A").to_s
+
+          version_rules.map(&.name).each do |name|
+            output.should contain name
+          end
```
straight-shoota, line 29: **thought:** This is close to a re-implementation of `RuleVersionsPresenter`. If both have the same logic error, the testing is pretty useless. Also the tests do not establish any relation between the individual items, so there could be many wrong outputs that still pass the test.
Why not test against a fixed output string?

src/ameba/runner.cr
```diff
@@ -59,18 +62,27 @@ module Ameba
...
-        config.rules
-          .find &.class.==(Rule::Lint::UnneededDisableDirective)
+        rules.find &.class.==(Rule::Lint::UnneededDisableDirective)
     end
 
-    protected def initialize(@rules, @sources, @formatter, @severity, @autocorrect = false)
+    protected def rule_runnable?(rule)
+      rule.enabled? && !rule.special? &&
+        (!(version = @version) || !(since_version = rule.since_version) ||
+          since_version <= version)
```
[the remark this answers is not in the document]
Sija: Agreed, although IMO it still small enough to let it be. I'll happily refactor it once it becomes unwieldy.

devnote-dev:
Is there anything left to be added for this to be merged?

Sija:
@devnote-dev I'm thinking of making another point-release before switching to 1.7, so until then I want to wait with merging this PR.

src/ameba/runner.cr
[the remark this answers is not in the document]
Sija: Addressed in 2e62b5d0b3da2d682a0a94b63f032ca5a620d79b

PR #474, opened 2024-10-08
Guard calls to the `typos` with a mutex

### LINKED ISSUE #473: MT breaks `Lint/Typos` rule
When ameba is built with `-Dpreview_mt`, the `Rule::Lint::Typos` rule hangs indefinitely while calling `Process.run` .

Sija:
Resolves #473


veelenga (approved):
lgtm

PR #476, opened 2024-10-10
Ignore block parameter names starting with `_`

### LINKED ISSUE #457: `Naming/BlockParameterName` with ignored names
Related: https://github.com/crystal-ameba/ameba/issues/453

I have a block parameter that starts with `_` indicating that this value is ignored, but since it still has a single letter, Ameba picks it up as a name that shouldn't be used

image

Maybe for this option, if the block parameter starts with `_` then Ameba could ignore it regardless of what it's called?

Sija:
Resolves #457


veelenga (approved):
lgtm

PR #478, opened 2024-10-10
Fix portable variable assignments in Makefile

### LINKED ISSUE #477: Q: I can't build ameba with make
This might be a very basic issue, and I might be missing something obvious, but I need some advice.

I tried the basic git clone and make process, but it failed:

```
git clone https://github.com/crystal-ameba/ameba
cd ameba
make
```

```
make: Nothing to be done for `all'.
```

What? I haven't built ameba in this directory before, so why am I seeing this message?

Next, I tried to install ameba using Homebrew, but it also failed:

```
brew tap crystal-ameba/ameba
brew install ameba
```

```
Error: An exception occurred within a child process:
  Errno::ENOENT: No such file or directory - bin/ameba
```

The error message says it can't find `bin/ameba`, so I suspect it's related to the same issue.

Since this is such a basic step, I thought it might be something wrong with my environment. So I forked the ameba repository and ran `make` using GitHub Actions CI, but it failed with the same error.

https://github.com/kojix2/ameba/actions/runs/11265560789/job/31327589181
https://github.com/crystal-ameba/ameba/compare/master...kojix2:ameba:u1

image

I might be missing something obvious, but this is the first step most people would try. There are probably others who faced the same issue and gave up on ameba. This might be a silly question, but I'm posting here, believing that there are at least ten people who have encountered the same problem.

straight-shoota:
Resolves #477

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -29,7 +29,7 @@ SHARD_BIN ?= ../../bin
 CRFLAGS ?= -Dpreview_mt
 
 SRC_SOURCES := $(shell find src -name '*.cr' 2>/dev/null)
-DOC_SOURCE   := src/**
+DOC_SOURCE   := src/cli.cr
 
 .PHONY: all
 all: build
--- a/Makefile
+++ b/Makefile
@@ -12,7 +12,7 @@ all:
 
 -include Makefile.local # for optional local options
 
-BUILD_TARGET ::= bin/ameba
+BUILD_TARGET := bin/ameba
 
 DESTDIR ?=          ## Install destination dir
 PREFIX ?= /usr/local## Install path prefix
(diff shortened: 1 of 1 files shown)


straight-shoota:
I'd also suggest using `make` in CI to remove duplication and have an integration test for the build system. And please remove the `branch: [master]` restriction from the CI workflow. That prevents CI from running on feature branches and you need extra hassle to make it work.

Makefile
```diff
@@ -28,8 +28,8 @@ INSTALL_BIN ?= /usr/bin/install
 SHARD_BIN ?= ../../bin
 CRFLAGS ?= -Dpreview_mt
 
-SRC_SOURCES ::= $(shell find src -name '*.cr' 2>/dev/null)
-DOC_SOURCE   ::= src/**
+SRC_SOURCES := $(shell find src -name '*.cr' 2>/dev/null)
+DOC_SOURCE   := src/cli.cr
```
Sija, line 32: Why this change?

straight-shoota: The glob catches all kinds of things, including directories 🤦
And we really only need one entry point. That should cover all code and makes sure it's required in a working order.

Sija: I believe you're correct (didn't tested it myself) 👍

Sija:
> I'd also suggest using `make` in CI to remove duplication and have an integration test for the build system.

Good point.

> And could we please remove the `branches: [master]` restriction from the CI workflow? It prevents CI from running on feature branches and you need extra hassle to make it work.

Why not open a (draft) PR?

Sija (approved):

PR #485, opened 2024-10-14
fix(lint): useless assign with method call within a condition

### LINKED ISSUE #468: False positive `Lint/UselessAssign`
```cr
hash = {"foo" => "bar"}

def call(v)
end

unless v = hash["foo"]?
  call v = "default"
end

v
```
Results in:
```cr
[W] Lint/UselessAssign: Useless assignment to variable `v`
> unless v = hash["foo"]?
```
The assignment is for sure not useless as it handles assigning `v` if that key is present in the hash, otherwise is assigned within the `unless` block.

Creating a new issue as unlike https://github.com/crystal-ameba/ameba/issues/447, this doesn't require any type declarations or macros.

veelenga:
fixes #468

### DIFF
--- a/spec/ameba/rule/lint/useless_assign_spec.cr
+++ b/spec/ameba/rule/lint/useless_assign_spec.cr
@@ -513,6 +513,33 @@ module Ameba::Rule::Lint
             CRYSTAL
         end
 
+        it "doesn't report if assignment is referenced within a method call" do
+          expect_no_issues subject, <<-CRYSTAL
+            if v = rand
+              puts(v = 1)
+            end
+            v
+            CRYSTAL
+
+          expect_no_issues subject, <<-CRYSTAL
+            puts v = 1 unless v = rand
+            v
+            CRYSTAL
+
+          expect_no_issues subject, <<-CRYSTAL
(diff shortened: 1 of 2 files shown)


src/ameba/ast/branch.cr
```diff
@@ -181,6 +181,8 @@ module Ameba::AST
         if loop?(node) && (block = node.block)
           on_branchable_start node, block.body
         end
+
+        true
```
veelenga, line 185: the nodes inside the `Crystal::Call` were ignored before

spec/ameba/rule/lint/useless_assign_spec.cr
```diff
@@ -513,6 +513,33 @@ module Ameba::Rule::Lint
...
+
+            def call(v)
+            end
+
+            unless v = hash["foo"]?
+              call v = "default"
+            end
+
+            v
+            CRYSTAL
```
Sija, line 540: Reduced examples should be fine.

```suggestion
```

Sija (changes requested):

Sija (approved):

PR #497, opened 2024-11-17
Add a JSON Schema for `.ameba.yml` and a CLI tool for automatically generating it

nobodywasishere:
When added to [JSON Schema Store](https://www.schemastore.org/json/), this will allow autocomplete and documentation information for the `.ameba.yml` format. Instead of writing one by-hand like was the case for `shard.yml`, instead I wrote a CLI tool that will automatically generate it and update it as new rules are added and changed.

https://github.com/user-attachments/assets/0e507ba0-e01a-4e04-b185-50945ed7f9f0


nobodywasishere:
How feasible would it be to make running this script / updating the schema a part of CI? That way ppl don't have to remember to run it for it to stay up to date.

Sija:
@nobodywasishere That certainly makes sense. It needs to be automated in order to make it hassle-free for the end-users.

nobodywasishere:
> That certainly makes sense. It needs to be automated in order to make it hassle-free for the end-users.

Didn't feel right about having a CI do commits, so instead added one that will fail if the user needs to run `shards run schema`, lmk if this works.

Sija:
> Didn't feel right about having a CI do commits, so instead added one that will fail if the user needs to run `shards run schema`, lmk if this works.

I'm gonna insist on having this automated (I see nothing bad in CI doing commits, btw) :) The end user IMO shouldn't be concerned with a need to operate tooling like this - except for the case when we're talking about rules development, then it makes sense, yet still more for local development than for our main repo.

straight-shoota:
@Sija IMO that requirement is asking too much at once. It brings a completely new concept into the CI game, which isn't even directly related. This feels like a very unnecessary obstacle.

This PR is a complete feature. It makes sure the schema cannot go out of sync. That's a solid base and should be acceptable.

Automating the synchronization is an enhancement for convenience. It should not be a hard requirement. Everything is safe and sound without it. It just replaces a round of potential manual intervention. But it's not a complex process to run the schema generator and commit its changes if CI tells you it needs an update.

Currently, this repo has not a single CI job that performs any automatic commits. So it seems very reasonable not to ask for introducing a completely new CI functionality on top of a feature change.

nobodywasishere:
> (I see nothing bad in CI doing commits, btw) :)

Maybe it could be fine, but I see it potentially adding to noise if someone's working on a rule, the commit being added and removed, etc, versus if it's just a simple command the dev needs to run once the rule is done being developed to pass CI.

> The end user IMO shouldn't be concerned with a need to operate tooling like this

The end user won't need to run the schema command, it's only meant for developers. The committed version of the schema file on master is all people need to worry about (or the schema file for their version of ameba - if people want to get that granular).

src/ameba/cli/schema.cr
```diff
@@ -0,0 +1,50 @@
...
+    bld.field("description", "Configuration rules for the Crystal lang ameba linter")
+    bld.field("type", "object")
+    bld.field("additionalProperties", false)
+
+    bld.string("properties")
+    bld.object do
+      bld.string("Excluded")
+      bld.object do
+        bld.field("type", "array")
+        bld.field("title", "excluded files and paths")
```
Sija: Use title-case:

```suggestion
        bld.field("title", "Excluded files and paths")
```

ditto elsewhere

src/ameba/config.cr
```diff
@@ -363,6 +363,135 @@ class Ameba::Config
           SemanticVersion.parse(version)
         end
       end
+
+      def self.to_json_schema(bld : JSON::Builder)
```
Sija:
```suggestion
      def self.to_json_schema(builder : JSON::Builder)
```

src/ameba/config.cr
```diff
@@ -363,6 +363,135 @@ class Ameba::Config
...
+            {% end %}
+
+            {% unless properties["excluded".id] %}
+              bld.string("Excluded")
+              bld.object do
+                bld.string("type")
+                bld.array do
+                  bld.string("array")
+                  bld.string("string")
+                end
```
Sija:
```suggestion
                bld.field("type", "array")
```

Sija (changes requested):

Sija (changes requested):

.github/workflows/ci.yml
```diff
@@ -9,6 +9,28 @@ on:
...
+
+    steps:
+      - name: Set timezone to UTC
+        uses: szenius/set-timezone@v2.0
+
+      - name: Install Crystal
+        uses: crystal-lang/install-crystal@v1
+
+      - name: Download source
+        uses: actions/checkout@v4
```
Sija:
```suggestion
        uses: actions/checkout@v5
```

src/ameba/cli/schema.cr
```diff
@@ -0,0 +1,50 @@
...
+    builder.field("type", "object")
+    builder.field("additionalProperties", false)
+
+    builder.string("properties")
+    builder.object do
+      builder.string("Excluded")
+      builder.object do
+        builder.field("type", "array")
+        builder.field("title", "Excluded files and paths")
+        builder.field("description", "an array of wildcards (or paths) to exclude from the source list")
```
Sija: Title-case please, ditto elsewhere.

```suggestion
        builder.field("description", "An array of wildcards (or paths) to exclude from the source list")
```

Sija:
Looks great! One thing, though, that still needs some attention is the size of the resulting json file. I believe it could be cut down by extracting the repetitive parts (common rule properties + severity enum) into their own (sub)schemas.

Sija (changes requested):

src/ameba/cli/schema.cr
```diff
@@ -0,0 +1,99 @@
...
+          builder.string("Enabled")
+          builder.object do
+            builder.field("type", "boolean")
+            builder.field("default", true)
+          end
+
+          builder.string("Severity")
+          builder.object do
+            builder.field("$ref", "#/definitions/Severity")
+            builder.field("default", Ameba::Severity.default_severity.to_s)
```
Sija: I'm sorry, that should have been this:

```suggestion
            builder.field("default", Ameba::Rule::Base.default_severity.to_s)
```

That's also general, since the default changes by the group... but it's better than nothing, I guess...

nobodywasishere:
I think it may be good to move away from force pushes, makes PRs a lot less clear, and we can just squash when merging to master

src/ameba/config.cr
```diff
@@ -363,6 +363,93 @@ class Ameba::Config
...
+
+            {% if !properties["severity".id] %}
+              if default_severity != Ameba::Rule::Base.default_severity
+                builder.string("Severity")
+                builder.object do
+                  builder.field("$ref", "#/definitions/Severity")
+                  builder.field("default", default_severity.to_s)
+                end
+              end
+            {% end %}
```
Sija: Nice having this, thanks for the effort! ❤️ 

```suggestion
            {% unless properties["severity".id] %}
              unless default_severity == Ameba::Rule::Base.default_severity
                builder.string("Severity")
                builder.object do
                  builder.field("$ref", "#/definitions/Severity")
                  builder.field("default", default_severity.to_s)
                end
              end
            {% end %}
```

Sija:
Then, the suggestions might end up incorrect due to the possible version difference between `master` and the locally installed version.

nobodywasishere:
That's completely fine imo. We don't know what version of ameba they're using and master provides the most up-to-date version. More accurate autocompletes n such can be provided by ameba-ls

nobodywasishere:
I haven't seen any if you can link to one

> Do we?

I meant client-side, in vscode or w/e. I don't know how vscode would know which schema to select beyond the most recent, as it doesn't involve ameba binaries at all.

In terms of releasing a v1.6.4 schema, we can just create a new `v1.6.4-schema` tag w/ this commit and point to that in the schema store. Doesn't need to be in the v1.6.4 tag

nobodywasishere:
https://github.com/SchemaStore/schemastore/blob/master/CONTRIBUTING.md#how-to-add-a-json-schema-with-multiple-versions

Sija:
Nice! In such a case, I'd generate the schema for v1.6.4 first and host it in the schemastore repo to see how it works, so if any changes are needed, we'll have it ironed out before v1.7.0, wdyt?

nobodywasishere:
https://github.com/SchemaStore/schemastore/pull/5131

PR #501, opened 2024-11-18
While passing file through `STDIN` we should lint just that

Sija:
Fixes the flawed feature implementation from #466
Essentially reverts f29cffa83b7efc3437419ecac9a6a5c83af4e39a


veelenga (approved):
lgtm

PR #507, opened 2024-11-22
Add `Lint/UnusedComparison` and `Lint/UnusedLiteral`

### LINKED ISSUE #235: Condition in void context
The `==` most like was supposed to be `=` here.

```
x = 23
x == 42
puts x
```

### LINKED ISSUE #508: Add `Lint/UnusedLiteral`
Similar to #235 (and potentially solving #467), I think it would be useful if there was a linting rule `Lint/UnusedLiteral` that reported when there was a literal value in a void context:

```crystal
def something : Bool
  if cond
    true
  # ^^^^ error: Literal value is not used
  end
  false
end
```

There would be _some_ overlap with `Lint/UselessComparison` if a literal is in an unused comparison, but maybe there can be a config option to not warn if inside an unused comparison(?).

nobodywasishere:
Closes #235
Closes #508

This is still a little WIP, I need to go through the code again and write more / better specs. Maybe the visitor can be moved to the main visitors, as it may be useful for writing other `Lint/Useless*` rules. It's basically useful for checking if the return value from a node will be captured or not, which is how I determined whether the comparison operations were useful.

image


Sija:
Nice one ❤️ I'd rename the rule to `UnusedComparison` though, _useless_ IMO brings some negative connotations.

Sija:
I'll do the code review once you mark the PR as _ready for review_, just FYI.

nobodywasishere:
> I'll do the code review once you mark the PR as _ready for review_, just FYI.

Yeah that's why I marked it as draft, so you wouldn't waste time reviewing it only for me to completely change it.

src/ameba/ast/visitors/implicit_return_visitor.cr
```diff
@@ -0,0 +1,216 @@
...
+        node.double_splat.try &.accept(self)
+        node.block_arg.try &.accept(self)
+      }
+
+      return_type = node.return_type
+      case return_type
+      when Crystal::Path
+        # Special case of the return type being nil, meaning the last
+        # line of the method body is ignored
+        if ["::Nil", "Nil"].includes?(return_type.names.join("::"))
```
nobodywasishere, line 96: If this is too hacky / "semantic" I can remove it. Just helps catch the edge case of the last line in def bodies being ignored if the return type is nil.

nobodywasishere:
Running these rules over the ameba, Kagi, and crystal compiler codebases, didn't turn up any violations. Don't know if that's due to a bug in the rules or violations being rare, though I'm leaning towards the latter.

nobodywasishere:
The only literal that isn't supported is `RegexLiteral`, and that's due to them not having a location.

src/ameba/ast/visitors/implicit_return_visitor.cr
```diff
@@ -0,0 +1,216 @@
+module Ameba::AST
+  class ImplicitReturnVisitor < BaseVisitor
+    # When greater than zero, indicates the current node's return value is used
+    @stack : Int32 = 0
+
+    def visit(node : Crystal::Expressions) : Nil
```
Sija, line 6: `visit` methods ought to return `Bool`, not `Nil`.

ditto elsewhere

src/ameba/ast/visitors/implicit_return_visitor.cr
```diff
@@ -0,0 +1,216 @@
...
+    def visit(node : Crystal::Call) : Nil
+      @rule.test(@source, node, @stack > 0)
+
+      if node.block
+        incr_stack { node.obj.try &.accept(self) }
+      else
+        node.obj.try &.accept(self)
+      end
+
+      incr_stack {
```
Sija, line 34: For multiline blocks use `do ... end`.

src/ameba/ast/visitors/implicit_return_visitor.cr
Sija, line 96:
```suggestion
        if return_type.names.join("::").in?("::Nil", "Nil")
```

src/ameba/ast/visitors/implicit_return_visitor.cr
```diff
@@ -0,0 +1,216 @@
...
+      @rule.test(@source, node, @stack > 0)
+
+      incr_stack {
+        node.args.each &.accept(self)
+        node.double_splat.try &.accept(self)
+        node.block_arg.try &.accept(self)
+      }
+
+      return_type = node.return_type
+      case return_type
```
Sija, line 92:
```suggestion
      case return_type = node.return_type
```

src/ameba/ast/visitors/implicit_return_visitor.cr
```diff
@@ -0,0 +1,216 @@
...
+    def visit(node : Crystal::ControlExpression) : Nil
+      @rule.test(@source, node, @stack > 0)
+
+      incr_stack { node.exp.try &.accept(self) }
+    end
+
+    def visit(node : Crystal::RangeLiteral) : Nil
+      @rule.test(@source, node, @stack > 0)
+
+      if !node.from.is_a?(Crystal::NumberLiteral)
```
Sija, line 183:
```suggestion
      unless node.from.is_a?(Crystal::NumberLiteral)
```

src/ameba/ast/visitors/implicit_return_visitor.cr
```diff
@@ -0,0 +1,216 @@
...
+        node.from.accept(self)
+      end
+
+      if !node.to.is_a?(Crystal::NumberLiteral)
```
Sija, line 187:
```suggestion
      unless node.to.is_a?(Crystal::NumberLiteral)
```

src/ameba/rule/lint/unused_comparison.cr
```diff
@@ -0,0 +1,62 @@
...
+      since_version "1.7.0"
+      description "Disallows unused comparison operations"
+    end
+
+    MSG = "Comparison operation is unused"
+
+    COMPARISON_OPERATORS = %w(
+      == != =~ !~ ===
+      < <= > >= <=>
+    )
```
Sija, line 48:
```suggestion
    COMPARISON_OPERATORS = %w[
      == != =~ !~ ===
      < <= > >= <=>
    ]
```

src/ameba/rule/lint/unused_comparison.cr
```diff
@@ -0,0 +1,62 @@
...
+
+    def test(source : Source)
+      AST::ImplicitReturnVisitor.new(self, source)
+    end
+
+    def test(source, node : Crystal::Call, last_is_used : Bool) : Bool
+      if !last_is_used && node.name.in?(COMPARISON_OPERATORS) && node.args.size == 1
+        issue_for node, MSG
+      end
+
```
Sija, line 58:
```suggestion
```

src/ameba/rule/lint/unused_literal.cr
```diff
@@ -0,0 +1,76 @@
...
+             Crystal::TupleLiteral | Crystal::NumberLiteral |
+             Crystal::StringLiteral | Crystal::SymbolLiteral |
+             Crystal::NamedTupleLiteral,
+      last_is_used : Bool
+    ) : Bool
+      if !last_is_used
+        issue_for node, MSG
+      end
+
+      true
```
Sija, line 73:
```suggestion
      issue_for node, MSG unless last_is_used
      true
```

Sija (changes requested):

src/ameba/ast/visitors/implicit_return_visitor.cr
nobodywasishere: IMO this could be another linter rule

src/ameba/ast/visitors/implicit_return_visitor.cr
nobodywasishere: I did this as I didn't want to use the native `accept_children` methods, and this was easier than adding a `false` to the end of every visit method

src/ameba/ast/visitors/implicit_return_visitor.cr
Sija: Agreed, I had it on my radar for a while, feel free to add it.

spec/ameba/rule/lint/unused_literal_spec.cr
```diff
@@ -0,0 +1,256 @@
...
+
+  describe UnusedLiteral do
+    it "passes if literals are used to assign" do
+      expect_no_issues subject, <<-CRYSTAL
+        a = 1
+        b = "string"
+        g = "interp \#{string}"
+        h = <<-HEREDOC
+        this is a heredoc
+        HEREDOC
```
Sija, line 14:
```suggestion
          this is a heredoc
          HEREDOC
```

spec/ameba/rule/lint/unused_literal_spec.cr
```diff
@@ -0,0 +1,256 @@
...
+        1..2
+      # ^^^^ error: Literal value is not used
+        {goodnight: moon}
+      # ^^^^^^^^^^^^^^^^^ error: Literal value is not used
+        {1, 2, 3}
+      # ^^^^^^^^^ error: Literal value is not used
+        <<-HEREDOC
+      # ^^^^^^^^^^ error: Literal value is not used
+        this is a heredoc
+        HEREDOC
```
Sija, line 80:
```suggestion
          this is a heredoc
          HEREDOC
```

spec/ameba/rule/lint/unused_literal_spec.cr
```diff
@@ -0,0 +1,256 @@
...
```
Sija, line 110:
```suggestion
          this is a heredoc
          HEREDOC
```

spec/ameba/rule/lint/unused_literal_spec.cr
```diff
@@ -0,0 +1,256 @@
...
```
Sija, line 142:
```suggestion
          this is a heredoc
          HEREDOC
```

spec/ameba/rule/lint/unused_literal_spec.cr
```diff
@@ -0,0 +1,256 @@
...
```
Sija, line 173:
```suggestion
          this is a heredoc
          HEREDOC
```

spec/ameba/rule/lint/unused_literal_spec.cr
```diff
@@ -0,0 +1,256 @@
...
+            [1, 2, 3, 4, 5]
+          # ^^^^^^^^^^^^^^^ error: Literal value is not used
+            {"hello" => "world"}
+          else
+            '\t'
+          # ^^^ error: Literal value is not used
+            <<-HEREDOC
+          # ^^^^^^^^^^ error: Literal value is not used
+            this is a heredoc
+            HEREDOC
```
Sija, line 246:
```suggestion
              this is a heredoc
              HEREDOC
```

src/ameba/ast/visitors/implicit_return_visitor.cr
```diff
@@ -0,0 +1,269 @@
...
+      @rule.test(@source, node, @stack > 0)
+
+      incr_stack do
+        node.args.each &.accept(self)
+        node.double_splat.try &.accept(self)
+        node.block_arg.try &.accept(self)
+      end
+
+      case return_type = node.return_type
+      when Crystal::Path
```
Sija, line 108:
```suggestion
      if (return_type = node.return_type).is_a?(Crystal::Path)
```

Sija (changes requested):

src/ameba/ast/visitors/implicit_return_visitor.cr
```diff
@@ -0,0 +1,269 @@
...
+      end
+
+      false
+    end
+
+    def visit(
+      node : Crystal::BoolLiteral | Crystal::CharLiteral |
+             Crystal::RegexLiteral | Crystal::NumberLiteral |
+             Crystal::StringLiteral | Crystal::SymbolLiteral |
+             Crystal::ProcLiteral
```
Sija, line 249:
```suggestion
             Crystal::ProcLiteral,
```

src/ameba/rule/lint/unused_literal.cr
```diff
@@ -0,0 +1,73 @@
...
+    end
+
+    def test(
+      source,
+      node : Crystal::BoolLiteral | Crystal::CharLiteral | Crystal::HashLiteral |
+             Crystal::ProcLiteral | Crystal::ArrayLiteral | Crystal::RangeLiteral |
+             Crystal::TupleLiteral | Crystal::NumberLiteral |
+             Crystal::StringLiteral | Crystal::SymbolLiteral |
+             Crystal::NamedTupleLiteral | Crystal::StringInterpolation,
+      last_is_used : Bool
```
Sija, line 67:
```suggestion
      last_is_used : Bool,
```

nobodywasishere:
If desired, I can add support for regex literals gated behind whether Ameba is built with Crystal 1.15.0 or later. This way, support can be added and merged now without requiring ameba to update its minimum version from 1.10

```diff
diff --git a/spec/ameba/rule/lint/unused_literal_spec.cr b/spec/ameba/rule/lint/unused_literal_spec.cr
index 5c22ea7b..7494a0ba 100644
--- a/spec/ameba/rule/lint/unused_literal_spec.cr
+++ b/spec/ameba/rule/lint/unused_literal_spec.cr
@@ -252,5 +252,16 @@ module Ameba::Rule::Lint
           end
       CRYSTAL
     end
+
+    # Locations for Regex literals were added in Crystal v1.15.0
+    {% if compare_versions(Crystal::VERSION, "1.15.0") >= 0 %}
+      it "fails if a regex literal is unused" do
+        expect_issue subject, <<-CRYSTAL
+            a = /hello world/
+            /goodnight moon/
+          # ^^^^^^^^^^^^^^^^ error: Literal value is not used
+          CRYSTAL
+      end
+    {% end %}
   end
 end
diff --git a/src/ameba/rule/lint/unused_literal.cr b/src/ameba/rule/lint/unused_literal.cr
index 50b3fdd6..cdf30201 100644
--- a/src/ameba/rule/lint/unused_literal.cr
+++ b/src/ameba/rule/lint/unused_literal.cr
@@ -57,6 +57,14 @@ module Ameba::Rule::Lint
       AST::ImplicitReturnVisitor.new(self, source)
     end
 
+    # Locations for Regex literals were added in Crystal v1.15.0
+    {% if compare_versions(Crystal::VERSION, "1.15.0") >= 0 %}
+      def test(source, node : Crystal::RegexLiteral, last_is_used : Bool) : Bool
+        issue_for node, MSG unless last_is_used
+        true
+      end
+    {% end %}
+
     def test(
       source,
       node : Crystal::BoolLiteral | Crystal::CharLiteral | Crystal::HashLiteral |
```

Sija:
@nobodywasishere That would be great 🚀

spec/ameba/rule/lint/unused_comparison_spec.cr
```diff
@@ -0,0 +1,172 @@
...
+        # ^^^^^ error: Comparison operation is unused
+        end
+
+        a = if x == 3
+              x > 1
+            # ^^^^^ error: Comparison operation is unused
+              x < 1
+            else
+              x > 1
+            end
```
veelenga, line 157: why the other two are not detected here?

src/ameba/rule/lint/unused_comparison.cr
```diff
@@ -0,0 +1,61 @@
...
+    def test(source : Source)
+      AST::ImplicitReturnVisitor.new(self, source)
+    end
+
+    def test(source, node : Crystal::Call, last_is_used : Bool) : Bool
+      if !last_is_used && node.name.in?(COMPARISON_OPERATORS) && node.args.size == 1
+        issue_for node, MSG
+      end
+      true
+    end
```
veelenga, line 59: why do we return Bool here? This looks uncommon for `test` methods

src/ameba/ast/visitors/implicit_return_visitor.cr
```diff
@@ -0,0 +1,268 @@
...
+      false
+    end
+
+    def visit(node : Crystal::HashLiteral | Crystal::NamedTupleLiteral) : Bool
+      @rule.test(@source, node, @stack > 0)
+
+      incr_stack { node.entries.each &.value.accept(self) }
+
+      false
+    end
```
veelenga, line 173: many of these methods return `false` meaning the nested AST will not expand further. However, what if HashLiteral has nested hashes or similar examples?

spec/ameba/rule/lint/unused_literal_spec.cr
```diff
@@ -0,0 +1,267 @@
...
+    # Locations for Regex literals were added in Crystal v1.15.0
+    {% if compare_versions(Crystal::VERSION, "1.15.0") >= 0 %}
+      it "fails if a regex literal is unused" do
+        expect_issue subject, <<-CRYSTAL
+            a = /hello world/
+            /goodnight moon/
+          # ^^^^^^^^^^^^^^^^ error: Literal value is not used
+          CRYSTAL
+      end
+    {% end %}
```
veelenga, line 265: do we want to add a test for the opposite situation to ensure it works as expected on Crystal < 1.15? Might be subjective.

spec/ameba/rule/lint/unused_comparison_spec.cr
nobodywasishere: Because they're assigned to the value

src/ameba/rule/lint/unused_comparison.cr
nobodywasishere: Leftover from old code, literal and return restr can be removed

src/ameba/ast/visitors/implicit_return_visitor.cr
nobodywasishere: They return false because I don't want to use the built in accept_children methods (which is what this bool indicates). Line 170 there has it visiting all of the entries of the hash, and because they're used in the hash, increments the stack.

spec/ameba/rule/lint/unused_literal_spec.cr
nobodywasishere: That works for me

src/ameba/ast/visitors/implicit_return_visitor.cr
veelenga: But that means we try to manually accept the children instead of delegating this work to the visitor. If a new node would be added to the syntax tree in crystal compiler, we will never visit it without changing this code.

nobodywasishere: Yes, that's true. We can't delegate this to the stdlib visitor though as we need to handle how the child elements are called, as different parts of elements are "used" or not (i.e., the condition of an `if` statement is used by itself, but the body and else are used by its parent).

The stdlib parser is internal code to the Crystal compiler, so using it requires some level of being okay with breaking changes as-is.

spec/ameba/rule/lint/unused_literal_spec.cr
nobodywasishere: Found a small bug because of this, thanks for the catch

veelenga (approved):
LGTM. Great job 👏🏻

carlhoerberg:
Seems to think that `IO.copy(in, out, len) == len || raise IO::EOFError.new`  is an unused comparison, and that `{ error: "abc" }.to_json(context.response)` is an unused literal.

nobodywasishere:
Good catch! I'll do a follow-up PR trying to limit these edge-cases. Since it's impossible to know whether a method has side-effects, I may not catch `||` and `&&` if one of them is a call, and not catch literals if they're the object of a call. I'll also do some tests to see if there are other edge-cases like this that need to be handled.

Issue #489, opened 2024-10-21
The `max` after the `map` is removed


kojix2:
Code:

```cr
list.map { |i| i.size }.max
```

Ameba auto correct

```
ameba --fix
```

The map is not replaced with max_of, but simply max is removed.

```cr
list.map(&.size)
```

kojix2:
The original code when I discovered this behavior was:

```
max = glossary_list.map { |g| g.name.size }.max.not_nil!
```

Sija:
Interestingly it occurs when both `Style/VerboseBlock` and `Performance/MinMaxAfterMap` rules are enabled and trying to auto-correct the code.

Sija:
@FnControlOption Could you take a look at it, please?

FnControlOption:
Hi @Sija, hope you are doing well. Unfortunately, I won’t have much time to look at this until around December 14

Sija:
@FnControlOption Hi, I hope you're doing well. Would you have time to look at this? All the best in the New Year, btw! 🎉

straight-shoota:
Just throwing in an idea: I'm wondering if it would make sense to order and separate rule application by category. For example, `Performance` rules could have higher precedence than `Style`: performance issues should be adressed first, before stylistic ones.

Of course it wouldn't solve the problem in general as it cannot avoid conflicts between rules in the same category. But it would reduce the problem space and fix the particular issue at hand.

PR #510, opened 2024-11-23
Add `Lint/RequireParentheses`

### LINKED ISSUE #313: Warn on expressions that contain logical operators but no braces
A pretty mean (and probably common) trap:

```crystal
$ cat >test.cr <<EOC
foo = ["batz"]

if foo.includes? "bar" || foo.includes? "batz"
  puts "this code is bug-free"
end
EOC
```

```bash
$ crystal test.cr
(no output)
```

```bash
$ ameba test.cr

Inspecting 1 file

.

Finished in 2.21 milliseconds
1 inspected, 0 failures
```

Can easily go unnoticed even with test-coverage.
I think it would be great if ameba would warn about it.

nobodywasishere:
Closes #313


spec/ameba/rule/lint/logical_without_paren_spec.cr
```diff
@@ -0,0 +1,44 @@
...
+        if foo.includes?("bar" || foo.includes? "batz")
+          puts "this code is bug-free"
+        end
+
+        if foo.includes?("bar") || foo.includes?("batz")
+          puts "this code is bug-free"
+        end
+
+        form.add("query", "val_1" || "val_2")
+        form.add "query", ("val_1" || "val_2")
```
Sija, line 18: Added another case which should be handled:

```suggestion
        form.add "query", ("val_1" || "val_2")
        form.add "query", "val_1" || "val_2"
```

src/ameba/rule/lint/logical_without_paren.cr
```diff
@@ -0,0 +1,55 @@
+module Ameba::Rule::Lint
+  # A rule that disallows logical operators in method args without parenthesis.
+  #
+  # For example, this is considered invalid:
+  #
+  # ```
+  # if a.includes? "b" && c.includes? "c"
```
Sija, line 7: I'd prefer the example from the OP, as it's more clear why it might be bug-prone:

```suggestion
  # if foo.includes? "bar" || foo.includes? "batz"
```

ditto other examples

Sija (changes requested):

src/ameba/rule/lint/logical_without_paren.cr
```diff
@@ -0,0 +1,55 @@
...
+  #
+  # ```
+  # if a.includes?("b") && c.includes?("c")
+  # end
+  #
+  # # or
+  #
+  # if a.includes?("b" && c.includes? "c")
+  # end
+  # ```
```
Sija, line 21:
```suggestion
  # ```
  # if foo.includes?("bar") || foo.includes?("batz")
  # end
  # ```
```

src/ameba/rule/lint/logical_without_paren.cr
```diff
@@ -0,0 +1,55 @@
...
+      description "Disallows logical operators in method args without parenthesis"
+    end
+
+    MSG = "Logical operator in method args without parenthesis is not allowed"
+
+    def test(source, node : Crystal::Call)
+      return if node.args.size == 0 ||
+                node.has_parentheses? ||
+                node.name.ends_with?("=") ||
+                node.name.in?(["[]?", "[]"])
```
Sija, line 35:
```suggestion
    MSG = "Logical operator in method args without parenthesis is not allowed"

    ALLOWED_CALL_NAMES = %w{[]? []}

    def test(source, node : Crystal::Call)
      return if node.args.empty? ||
                node.has_parentheses? ||
                node.name.ends_with?('=') ||
                node.name.in?(ALLOWED_CALL_NAMES)
```

spec/ameba/rule/lint/logical_without_paren_spec.cr
```diff
@@ -0,0 +1,45 @@
...
+        form.add "query", "val_1" || "val_2"
+        CRYSTAL
+    end
+
+    it "passes if logical operator in assignment call" do
+      expect_no_issues subject, <<-CRYSTAL
+        hello.there = "world" || method.call
```
Sija, line 25:
```suggestion
        hello.there = "world" || method.call
        hello.there ||= "world" || method.call
```

spec/ameba/rule/lint/logical_without_paren_spec.cr
```diff
@@ -0,0 +1,45 @@
...
+        CRYSTAL
+    end
+
+    it "passes if logical operator in square bracket call" do
+      expect_no_issues subject, <<-CRYSTAL
+        hello["world" || :thing]
```
Sija, line 31:
```suggestion
        hello["world" || :thing]
        hello["world" || :thing]?
```

Sija (changes requested):

spec/ameba/rule/lint/logical_without_paren_spec.cr
```diff
@@ -0,0 +1,47 @@
...
+          puts "this code is bug-free"
+        end
+
+        if foo.includes?("bar") || foo.includes?("batz")
+          puts "this code is bug-free"
+        end
+
+        form.add("query", "val_1" || "val_2")
+        form.add "query", ("val_1" || "val_2")
+        form.add "query", "val_1" || "val_2"
```
Sija, line 19:
```suggestion
        if foo.includes?("bar") || foo.includes?("batz")
          puts "this code is bug-free"
        end

        if foo.includes?("bar" || foo.includes? "batz")
          puts "this code is bug-free"
        end

        form.add("query", "val_1" || "val_2")
        form.add "query", "val_1" || "val_2"
        form.add "query", ("val_1" || "val_2")
```

src/ameba/rule/lint/logical_without_paren.cr
```diff
@@ -0,0 +1,48 @@
...
+  # ```
+  # Lint/LogicalWithoutParenthesis:
+  #   Enabled: true
+  # ```
+  class LogicalWithoutParenthesis < Base
+    properties do
+      description "Disallows logical operators in method args without parenthesis"
+    end
+
+    MSG = "Logical operator in method args without parenthesis is not allowed"
```
Sija, line 29: Rule name and description is not very descriptive nor accurate, let's change it to (following [RuboCop](https://github.com/rubocop/rubocop/blob/ac98713904b3e6f10605f00786e85bc3b3d0aa15/lib/rubocop/cop/lint/require_parentheses.rb) here):

```suggestion
  # ```
  # Lint/RequireParentheses:
  #   Enabled: true
  # ```
  class RequireParentheses < Base
    properties do
      description "Disallows method calls with no parentheses and a logical operator in the argument list"
    end

    MSG = "Use parentheses in the method call to avoid confusion about precedence"
```

This needs to be updated in other places too (incl. the rule filename).

nobodywasishere: Yeah I had trouble coming up with a good name for this, I'll update. Thank you!

Sija (changes requested):

src/ameba/rule/lint/logical_without_paren.cr
```diff
@@ -0,0 +1,48 @@
+module Ameba::Rule::Lint
+  # A rule that disallows logical operators in method args without parenthesis.
```
Sija, line 2: Wording could be improved still, yet I believe it's a bit better than before.

```suggestion
  # A rule that disallows method calls with at least one argument, where no parentheses are used around
  # the argument list, and a logical operator (`&&` or `||`) is used within the argument list.
```

```suggestion
  # A rule that disallows logical operators in method args without parenthesis.
```

Sija (approved):
LGTM

Sija:
I'll merge it once v1.6.4 is out (which gonna be soon).

src/ameba/rule/lint/require_parentheses.cr
```diff
@@ -0,0 +1,50 @@
...
+  # end
+  # ```
+  #
+  # YAML configuration example:
+  #
+  # ```
+  # Lint/RequireParentheses:
+  #   Enabled: true
+  # ```
+  class RequireParentheses < Base
```
veelenga, line 26: would be nice to have a corrector for this one.

veelenga (approved):
LGTM

src/ameba/rule/lint/require_parentheses.cr
Sija: I agree, although it might be added in followup PR. Also, I believe it's better to fix the bugs that popped up recently before adding even more rules that might trip on 'em.

nobodywasishere: What bugs are you referring to? I'm fine either adding it now or later on, though may take some time to figure out the best way to do it.

Sija: @nobodywasishere #489 to be exact. I'd prefer adding this l8r on tbh.

nobodywasishere: Either way works. Don't expect 1.7.0 to come out for a while either. Just don't want this (and other PRs) to bit rot.

src/ameba/rule/lint/require_parentheses.cr
```diff
@@ -0,0 +1,50 @@
...
+    properties do
```
Sija, line 27:
```suggestion
    properties do
      since_version "1.7.0"
```

PR #511, opened 2024-11-23
Add `Lint/ReadInstanceVar`

### LINKED ISSUE #463: Lint against accessing instance variables outside of the instance of a class
It's possible to access instance variables on an instance of a class similarly to `instance_variable_get` in Ruby

```crystal
class MyKlass
  def initialize(@my_instance_var : String); end
end

puts MyKlass.new("hello").@my_instance_var
# => "hello"
```
[Crystal Playground](https://play.crystal-lang.org/#/r/gzsz)

This can be handy for debugging purposes but I think generally should be discouraged as I think it's commonly expected that instance variables are private to the instance of a class.

nobodywasishere:
Closes #463

### DIFF
--- a/src/ameba/rule/lint/read_instance_var.cr
+++ b/src/ameba/rule/lint/read_instance_var.cr
@@ -2,12 +2,33 @@ module Ameba::Rule::Lint
   # A rule that disallows reading instance variables externally from an object
   # via the `object.@ivar` syntax.
   #
+  # This is only allowed when inside a method def in a class, and the object
+  # is typed to be the same as the current class.
+  #
   # For example, this is not allowed:
+  #
+  # ```
+  # class Greeter
+  #   def combine(other)
+  #     @ivar <=> other.@ivar
+  #   end
+  #
+  #   def split(other : Smiler)
+  #     @ivar - other.@ivar
+  #   end
+  # end
+  # ```
+  #
+  # And this is allowed:
   #
   # ```
   # class Greeter
   #   def combine(other : self)
   #     @ivar <=> other.@ivar
+  #   end
+  #
+  #   def split(other : Greeter)
+  #     @ivar - other.@ivar
   #   end
   # end
   # ```
@@ -21,13 +42,65 @@ module Ameba::Rule::Lint
   class ReadInstanceVar < Base
     properties do
       description "Disallows external reading of instance vars"
-      enabled false
     end
 
     MSG = "Reading instance variables externally is not allowed."
 
-    def test(source, node : Crystal::ReadInstanceVar)
-      issue_for node, MSG
+    def test(source)
+      InstanceVarReadVisitor.new(self, source)
+    end
+
+    def test(
+      source, node : Crystal::ReadInstanceVar,
+      allowed_names : Array(String)
+    ) : Nil
+      case obj = node.obj
+      when Crystal::Var
+        unless obj.name.in?(allowed_names)
+          issue_for node, MSG, prefer_name_location: true
+        end
+      else
+        issue_for node, MSG, prefer_name_location: true
+      end
+    end
+
+    private class InstanceVarReadVisitor < AST::NodeVisitor
+      @class_name : String?
+      @param_names : Array(String) = [] of String
+
+      def visit(node : Crystal::ClassDef) : Nil
+        prev_class = @class_name
+        @class_name = node.name.names[-1]?
+
+        node.body.accept(self)
+
+        @class_name = prev_class
+      end
+
+      def visit(node : Crystal::Def)
+        node.args.each do |arg|
+          case restriction = arg.restriction
+          when Crystal::Path
+            if restriction.names[-1]? == @class_name
+              @param_names << arg.name
+            end
+          when Crystal::Self
+            @param_names << arg.name
+          end
+        end
+
+        super
+      end
+
+      def end_visit(node : Crystal::Def)
+        @param_names.clear
+        super
+      end
+
+      def visit(node : Crystal::ReadInstanceVar)
+        @rule.test(@source, node, @param_names)
+        super
+      end
     end
   end
 end
--- a/src/ameba/rule/lint/read_instance_var.cr
+++ b/src/ameba/rule/lint/read_instance_var.cr
@@ -22,13 +22,13 @@ module Ameba::Rule::Lint
   # And this is allowed:
   #
   # ```
-  # class Greeter
-  #   def combine(other : self)
+  # class Foo
+  #   def <=>(other : self)
   #     @ivar <=> other.@ivar
   #   end
   #
-  #   def split(other : Greeter)
-  #     @ivar - other.@ivar
+  #   def <=>(other : Foo)
+  #     @ivar <=> other.@ivar
   #   end
   # end
   # ```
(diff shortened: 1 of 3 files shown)


straight-shoota:
IMHO such a rule is impossible to implement without taking scopes into account. Examples like the one in https://github.com/crystal-ameba/ameba/issues/463#issuecomment-2225517353 where an ivar is accessed from the same class should be fine.

Doing this correctly would require semantic analysis though.

nobodywasishere:
IMO in that case I think the upstream issue should be closed as no-fix.

Sija:
I think it should be doable with a low false-positive rate using a custom visitor, that ignores the access done for the object of the same class.

nobodywasishere:
Hm okay, I'll try that. Do you mean something along the lines of class/struct ClassName > def.args.any? &.restriction == ClassName or self, then reading instance vars is allow to that variable?

Sija:
Sth along these lines, yes. Top-level code as well needs to be taken into account.

src/ameba/rule/lint/read_instance_var.cr
```diff
@@ -0,0 +1,106 @@
...
+  # ```
+  # class Greeter
+  #   def combine(other)
+  #     @ivar <=> other.@ivar
+  #   end
+  #
+  #   def split(other : Smiler)
+  #     @ivar - other.@ivar
+  #   end
+  # end
```
Sija, line 19: Let's make the examples more abstract (like below):

```suggestion
  # class Foo
  #   def <=>(other)
  #     @ivar <=> other.@ivar
  #   end
  #
  #   def <=>(other : Bar)
  #     @ivar <=> other.@ivar
  #   end
  # end
```

src/ameba/rule/lint/read_instance_var.cr
```diff
@@ -0,0 +1,106 @@
...
+  # ```
+  # class Greeter
+  #   def combine(other : self)
+  #     @ivar <=> other.@ivar
+  #   end
+  #
+  #   def split(other : Greeter)
+  #     @ivar - other.@ivar
+  #   end
+  # end
```
Sija, line 33: ditto

```suggestion
  # class Foo
  #   def <=>(other : self)
  #     @ivar <=> other.@ivar
  #   end
  #
  #   def <=>(other : Foo)
  #     @ivar <=> other.@ivar
  #   end
  # end
```

src/ameba/rule/lint/read_instance_var.cr
```diff
@@ -0,0 +1,106 @@
...
+  # ```
+  #
+  # YAML configuration example:
+  #
+  # ```
+  # Lint/ReadInstanceVar:
+  #   Enabled: false
```
Sija, line 40: These should list the default settings:

```suggestion
  #   Enabled: true
```

src/ameba/rule/lint/read_instance_var.cr
```diff
@@ -0,0 +1,106 @@
...
+  # ```
+  class ReadInstanceVar < Base
+    properties do
+      description "Disallows external reading of instance vars"
+    end
+
+    MSG = "Reading instance variables externally is not allowed."
```
Sija, line 47: Let's stick to the already established stylistic convention:

```suggestion
    MSG = "Reading instance variables externally is not allowed"
```

src/ameba/rule/lint/read_instance_var.cr
```diff
@@ -0,0 +1,106 @@
...
+
+    def test(source)
+      InstanceVarReadVisitor.new(self, source)
+    end
+
+    def test(
+      source, node : Crystal::ReadInstanceVar,
+      allowed_names : Array(String)
+    ) : Nil
```
Sija, line 56:
```suggestion
    def test(source, node : Crystal::ReadInstanceVar, allowed_names)
```

src/ameba/rule/lint/read_instance_var.cr
```diff
@@ -0,0 +1,106 @@
...
+          issue_for node, MSG, prefer_name_location: true
+        end
+      else
+        issue_for node, MSG, prefer_name_location: true
+      end
+    end
+
+    private class InstanceVarReadVisitor < AST::NodeVisitor
+      @class_name : String?
+      @param_names : Array(String) = [] of String
```
Sija, line 69: Redundant type declaration

```suggestion
      @param_names = [] of String
```

src/ameba/rule/lint/read_instance_var.cr
```diff
@@ -0,0 +1,106 @@
...
+          case restriction = arg.restriction
+          when Crystal::Path
+            if restriction.names[-1]? == @class_name
+              @param_names << arg.name
+            end
+          when Crystal::Self
+            @param_names << arg.name
+          end
+        end
+
```
Sija, line 91:
```suggestion
```

src/ameba/rule/lint/read_instance_var.cr
```diff
@@ -0,0 +1,106 @@
...
+      @param_names : Array(String) = [] of String
+
+      def visit(node : Crystal::ClassDef) : Nil
+        prev_class = @class_name
+        @class_name = node.name.names[-1]?
+
+        node.body.accept(self)
+
+        @class_name = prev_class
+      end
```
Sija, line 78:
```suggestion
      def visit(node : Crystal::ClassDef)
        prev_class = @class_name
        @class_name = node.name.names[-1]?
        begin
          node.body.accept(self)
        ensure
          @class_name = prev_class
        end
        true
      end
```

Sija (changes requested):

Sija:
Formatter on Crystal nightly is failing, btw.

src/ameba/rule/lint/read_instance_var.cr
```diff
@@ -0,0 +1,103 @@
...
+      description "Disallows external reading of instance vars"
+    end
+
+    MSG = "Reading instance variables externally is not allowed"
+
+    def test(source)
+      InstanceVarReadVisitor.new(self, source)
+    end
+
+    def test(source, node : Crystal::ReadInstanceVar, allowed_names)
```
Sija, line 53: Looks like the type restriction is actually needed here, my bad.

```suggestion
    def test(source, node : Crystal::ReadInstanceVar, allowed_names : Enumerable(String))
```

straight-shoota:
As I mentioned before, it is impossible to have satisfactory type awareness without semantic analysis. 

You won't be able to determine the type of the target object with only syntactical information. The proposed implementation is very poor and limited. IIUC it only allows access to variables which are passed as parameters with explicit type restriction into a def. I'm not sure where the assumption comes from that valid access should only be allowed to parameters. It should be noted that a valid target object ought to be any kind of local variable, call, or even an instance variable read itself. But even there was justification for only considering parameters, there's still lots of issues. For example, the type of the variable at the time of the instance variable read does not need to be the same as in the method signature. Also, purely lexical comparison of paths may miss important details such as inheritance, aliases, generic arguments, ... I'm sure there are many more problems.

As a result, successful detection rate is quite miserable. IMO this is not even worth trying.
It should be better to invest the effort into figuring out how we could evolve ameba to make use of semantical information.

It's also worth noting that both human readers as well as the compiler (through semantic analysis) can easily understand the typing situation correctly. False positives on clear cases would be perceived as quite irritating issues from a linter.

Sija:
@straight-shoota Good points, agreed.

nobodywasishere:
I agree as well. I'll create an issue to start discussion of what integrating more semantic information into ameba would look like.

PR #512, opened 2024-11-24
Add `Lint/TrailingRescueException`

### LINKED ISSUE #242: Warn on single line `rescue`s with an exception value
Single-line `rescue` expressions swallow all exceptions, returning the argument to the `rescue`. 

It's a relatively common mistake to put an exception as the argument, mistakingly assuming that it will rescue from _that_ exception alone, rather than _all_ exceptions.

[See the example below.](https://carc.in/#/r/bwaq)

```crystal
class SpecificError < Exception
end

class OtherError < Exception
end

def raises_other_error
  raise OtherError.new
end

result = raises_other_error rescue SpecificError

puts result # => SpecificError
```

I propose that ameba produces a warning when the argument to a single-line `rescue` is an exception.

nobodywasishere:
This prevents the use of Paths as the default value for a trailing exception, and recommends the user use the block rescue instead to be able to filter by exception type.

Closes #242.

### DIFF
--- a/spec/ameba/rule/lint/trailing_rescue_exception_spec.cr
+++ b/spec/ameba/rule/lint/trailing_rescue_exception_spec.cr
@@ -19,7 +19,7 @@ module Ameba::Rule::Lint
   it "fails if trailing rescue has exception name" do
     expect_issue subject, <<-CRYSTAL
         puts "hello" rescue MyException
-                          # ^^^^^^^^^^^ error: Trailing rescues with a path aren't allowed; use a block rescue instead to filter by exception type
+                          # ^^^^^^^^^^^ error: Trailing rescues with a path aren't allowed, use a block rescue instead to filter by exception type
         CRYSTAL
   end
 end
--- a/spec/ameba/rule/lint/trailing_rescue_exception_spec.cr
+++ b/spec/ameba/rule/lint/trailing_rescue_exception_spec.cr
@@ -1,0 +1,25 @@
+require "../../../spec_helper"
+
+module Ameba::Rule::Lint
+  subject = TrailingRescueException.new
+
+  it "passes for trailing rescue with literal values" do
+    expect_no_issues subject, <<-CRYSTAL
+        puts "hello" rescue "world"
+        puts :meow rescue 1234
+        CRYSTAL
+  end
+
+  it "passes for trailing rescue with class initialization" do
+    expect_no_issues subject, <<-CRYSTAL
+        puts "hello" rescue MyClass.new
+        CRYSTAL
+  end
+
+  it "fails if trailing rescue has exception name" do
+    expect_issue subject, <<-CRYSTAL
+        puts "hello" rescue MyException
+                          # ^^^^^^^^^^^ error: Trailing rescues with a path aren't allowed; use a block rescue instead to filter by exception type
+        CRYSTAL
+  end
+end
--- a/src/ameba/rule/lint/trailing_rescue_exception.cr
+++ b/src/ameba/rule/lint/trailing_rescue_exception.cr
@@ -19,7 +19,7 @@ module Ameba::Rule::Lint
   # end
   # ```
   #
-  # Or to capture all exceptions (instead of just `MyException`):
+  # Or to rescue all exceptions (instead of just `MyException`):
   #
   # ```
   # value = method.call("param") rescue "default value"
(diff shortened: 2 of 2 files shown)


src/ameba/rule/lint/trailing_rescue_exception.cr
```diff
@@ -0,0 +1,50 @@
+module Ameba::Rule::Lint
+  # A rule that prohibits the common misconception about trailing rescue statements,
+  # preventing the use of Paths (exception class names or otherwise) from being
+  # used as the trailing value.
+  #
+  # For example, this is considered invalid:
+  #
+  # ```
+  #  value = method.call('param') rescue MyException
```
Sija, line 9:
```suggestion
  # value = method.call("param") rescue MyException
```

src/ameba/rule/lint/trailing_rescue_exception.cr
```diff
@@ -0,0 +1,50 @@
...
+  # ```
+  # Lint/TrailingRescueException:
+  #   Enabled: true
+  # ```
+  class TrailingRescueException < Base
+    properties do
+      description "Disallows trailing rescue with a path"
+    end
+
+    MSG = "Trailing rescues with a path aren't allowed; use a block rescue instead to filter by exception type"
```
Sija, line 39:
```suggestion
    MSG = "Trailing rescues with a path aren't allowed, use a block rescue instead to filter by exception type"
```

spec/ameba/rule/lint/trailing_rescue_exception_spec.cr
```diff
@@ -0,0 +1,25 @@
...
+  it "passes for trailing rescue with class initialization" do
+    expect_no_issues subject, <<-CRYSTAL
+        puts "hello" rescue MyClass.new
+        CRYSTAL
+  end
+
+  it "fails if trailing rescue has exception name" do
+    expect_issue subject, <<-CRYSTAL
+        puts "hello" rescue MyException
+                          # ^^^^^^^^^^^ error: Trailing rescues with a path aren't allowed; use a block rescue instead to filter by exception type
```
Sija, line 22:
```suggestion
                          # ^^^^^^^^^^^ error: Trailing rescues with a path aren't allowed, use a block rescue instead to filter by exception type
```

src/ameba/rule/lint/trailing_rescue_exception.cr
```diff
@@ -0,0 +1,50 @@
...
+  #
+  # ```
+  # value = begin
+  #   method.call("param")
+  # rescue MyException
+  #   "default value"
+  # end
+  # ```
+  #
+  # Or to capture all exceptions (instead of just `MyException`):
```
Sija, line 22:
```suggestion
  # Or to rescue all exceptions (instead of just `MyException`):
```

Sija (changes requested):

src/ameba/rule/lint/trailing_rescue_exception.cr
nobodywasishere: Haha been writing too much typescript

Sija (approved):
LGTM

Sija:
I'm gonna merge it, once v1.6.4 is finalized and released (which should be pretty soon).

src/ameba/rule/lint/trailing_rescue_exception.cr
```diff
@@ -0,0 +1,50 @@
...
+  # ```
+  #
+  # YAML configuration example:
+  #
+  # ```
+  # Lint/TrailingRescueException:
+  #   Enabled: true
+  # ```
+  class TrailingRescueException < Base
+    properties do
```
Sija, line 35:
```suggestion
    properties do
      since_version "1.7.0"
```

src/ameba/rule/lint/trailing_rescue_exception.cr
```diff
@@ -0,0 +1,51 @@
...
+  #
+  # And should instead be written as:
+  #
+  # ```
+  # value = begin
+  #   method.call("param")
+  # rescue MyException
+  #   "default value"
+  # end
+  # ```
```
veelenga, line 20: Since this doc is shown when using `explain` command, I would suggest to add some better explanation about misconception here. There is an example, which is good, but maybe we could explain what is actually wrong here?

nobodywasishere: I've updated this, let me know if there's more information / clarity that should be added.

veelenga (approved):
LGTM

Issue #502, opened 2024-11-18
Add rules for enforcing type restrictions on vars and parameters


nobodywasishere:
I'm imagining this would be a few rules, and not sure what namespace they would fall under. Something like:

```yml
Typing/InstanceVarTypeRestriction: # and ClassVarTypeRestriction
  Enabled: true

Typing/LocalVarTypeRestriction:
  Enabled: true
  
Typing/MethodParamTypeRestriction:
  Enabled: true
  DefaultValue: false
  PrivateMethods: false # don't require on private methods
  ProtectedMethods: false
  BlockParam: false # whether typing is required for block params

Typing/MethodReturnTypeRestriction:
  Enabled: true
  PrivateMethods: false
  ProtectedMethods: false

Typing/ProcReturnTypeRestriction:
  Enabled: true

# Maybe some others, can't think of any others currently
```

This would be disabled by default. It wouldn't be perfect, but would definitely be good enough for my use-cases. If this feature would be welcome, I can start prototyping an implementation.

nobodywasishere:
For `getter`/`setter`/`record`, there could be a rule like:

```yml
Typing/MacroCallVarTypeRestriction:
  Enabled: true
  MacroNames:
   - record
   - getter
   - setter
   - property
```

straight-shoota:
Instance variables always need to be typed. The compiler is able to infer the type in some cases.
So I'm wondering what `InstanceVarTypeRestriction` with `DefaultValue: false` would complain about other than what is already a compiler error? Same for `ClassVarTypeRestriction`.

Sija:
- `InstanceVarTypeRestriction`: doesn't IMO make much sense, since ivars need to have the type, either defined explicitly or inferred
- `LocalVarTypeRestriction`: I'm against this one, since typing local vars is more a problem than a solution really
- `MethodParamTypeRestriction` and `MethodReturnTypeRestriction`: I'm ok with
- `ProcReturnTypeRestriction`: by _proc_ you meant _block_?
- `MacroCallVarTypeRestriction`: could you provide any use-case?

devnote-dev:
> - `InstanceVarTypeRestriction` and `ClassVarTypeRestriction`: doesn't IMO make much sense, since ivars/cvars need to have the type, either defined explicitly or inferred

That's all good and well from the compiler's perspective, but isn't the slightest bit helpful to users of, for example, a shard which makes public use of otherwise "untyped" instance variables:

```crystal
class Foo
  getter bar

[...]

PR #518, opened 2024-11-26
Add several rules regarding type restrictions

nobodywasishere:
Related to #502.

- Adds `Typing/MacroCallVarTypeRestriction`
- Adds `Typing/MethodParamTypeRestriction`
- Adds `Typing/MethodReturnTypeRestriction`
- Adds `Typing/ProcReturnTypeRestriction`

Proc return type is enabled by default as it can lead to issues with passing procs to C, and proc params otherwise need to be typed. The other rules are off by default, and have several configuration options for those who want to use them in different ways (for instance, I can see the stdlib / library shards using the method param / return type rules for public / documented methods, to ensure there isn't breakage when changes are made).

I can add the `since_version` configs once a review is done.

### DIFF
--- a/spec/ameba/base_spec.cr
+++ b/spec/ameba/base_spec.cr
@@ -19,6 +19,7 @@ module Ameba::Rule
           Naming
           Performance
           Style
+          Typing
         ]
       end
     end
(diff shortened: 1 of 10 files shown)


nobodywasishere (commented):
Should probably also have a config for splat / double splat params

Sija (changes requested):
Please open a separate PRs for each rule. It will make the review process easier and be more visible in the changelog.

PR #528, opened 2024-12-06
Sort sources and issues before passing them to the formatter

Sija:
This will ensure a consistent experience for the end-user.

### DIFF
--- a/src/ameba/runner.cr
+++ b/src/ameba/runner.cr
@@ -72,7 +72,8 @@ module Ameba
       )
     end
 
-    protected def initialize(rules, @sources, @formatter, @severity, @autocorrect = false, @version = nil)
+    protected def initialize(rules, sources, @formatter, @severity, @autocorrect = false, @version = nil)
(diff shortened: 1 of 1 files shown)


src/ameba/runner.cr
```diff
@@ -159,6 +160,10 @@ module Ameba
 
       File.write(source.path, source.code) unless corrected_issues.empty?
     ensure
+      missing_location = Crystal::Location.new(nil, 0, 0)
+      source.issues.sort_by! do |issue|
+        issue.location || missing_location
+      end
```
veelenga, line 166: didn't test, but how will this work in parallel env?

Sija: Should be fine, since `run_source` is being run per fiber:

https://github.com/crystal-ameba/ameba/blob/29989b06c8fb5543d18221914a188cf684d5774c/src/ameba/runner.cr#L106-L114

veelenga (approved):
lgtm

PR #533, opened 2024-12-10
Fix edge cases with `Lint/UnusedComparison` and `Lint/UnusedLiteral`

nobodywasishere:
Follow up to #507 to fix some things I missed.

- Comparisons with a regex using `===` and `=~` have side effects, so they should not be counted.
- If the right of an And/Or is a call, the left should be counted, as it potentially has side effects / will control whether the right is called.
- If a call has args, count its object as being used, as it could have side effects.
- Fix yield and literals in generics having their expressions counted as unused.

See the specs for examples, mostly pulled from running this over the Crystal stdlib/compiler and seeing what it catches.

### DIFF
--- a/spec/ameba/rule/lint/unused_comparison_spec.cr
+++ b/spec/ameba/rule/lint/unused_comparison_spec.cr
@@ -40,7 +40,7 @@ module Ameba::Rule::Lint
     it "passes for unused comparisons with `===`, `=~`, and `!~`" do
       expect_no_issues subject, <<-CRYSTAL
         /foo(bar)?/ =~ baz
-        "foobar" !~ /(o+)ba(r?)/
+        /foo(bar)?/ !~ baz
         "hello" === world
         CRYSTAL
     end
--- a/spec/ameba/rule/lint/unused_comparison_spec.cr
+++ b/spec/ameba/rule/lint/unused_comparison_spec.cr
@@ -30,6 +30,21 @@ module Ameba::Rule::Lint
         CRYSTAL
     end
 
+    it "passes for comparisons inside '||' and '&&' where the other arg is a call" do
+      expect_no_issues subject, <<-CRYSTAL
+          IO.copy(in_var, out_var, len) == len || raise IO::EOFError.new
+        CRYSTAL
+    end
+
+    it "passes if a regex literal is part of a === or =~ comparison" do
+      expect_no_issues subject, <<-CRYSTAL
+        /f(o+)(bar?)/ === "fooba"
+        puts $~
+        "foobar" =~ /(o+)ba(r?)/
+        puts $1
+      CRYSTAL
+    end
+
     it "fails for all comparison operators" do
       expect_issue subject, <<-CRYSTAL
           x == 2
--- a/spec/ameba/rule/lint/unused_comparison_spec.cr
+++ b/spec/ameba/rule/lint/unused_comparison_spec.cr
@@ -37,7 +37,7 @@ module Ameba::Rule::Lint
         CRYSTAL
     end
 
-    it "passes if a regex literal is part of a === or =~ comparison" do
+    it "passes if a regex literal is part of a comparison" do
       expect_no_issues subject, <<-CRYSTAL
         /f(o+)(bar?)/ === "fooba"
         puts $~
--- a/spec/ameba/rule/lint/unused_comparison_spec.cr
+++ b/spec/ameba/rule/lint/unused_comparison_spec.cr
@@ -32,7 +32,8 @@ module Ameba::Rule::Lint
 
     it "passes for comparisons inside '||' and '&&' where the other arg is a call" do
       expect_no_issues subject, <<-CRYSTAL
-          IO.copy(in_var, out_var, len) == len || raise IO::EOFError.new
+        foo(bar) == baz || raise "bat"
+        foo(bar) == baz && raise "bat"
         CRYSTAL
     end
 
--- a/spec/ameba/rule/lint/unused_comparison_spec.cr
+++ b/spec/ameba/rule/lint/unused_comparison_spec.cr
@@ -37,13 +37,12 @@ module Ameba::Rule::Lint
         CRYSTAL
     end
 
-    it "passes if a regex literal is part of a comparison" do
+    it "passes for unused comparisons with `===`, `=~`, and `!~`" do
       expect_no_issues subject, <<-CRYSTAL
-        /f(o+)(bar?)/ === "fooba"
-        puts $~
-        "foobar" =~ /(o+)ba(r?)/
-        puts $1
-      CRYSTAL
+        /foo(bar)?/ =~ baz
+        "foobar" !~ /(o+)ba(r?)/
+        "hello" === world
+        CRYSTAL
     end
 
     it "fails for all comparison operators" do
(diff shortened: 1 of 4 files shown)


src/ameba/ast/visitors/implicit_return_visitor.cr
```diff
@@ -24,10 +24,24 @@ module Ameba::AST
       false
     end
 
+    def visit(node : Crystal::And | Crystal::Or) : Bool
```
Sija, line 27:
```suggestion
    def visit(node : Crystal::BinaryOp) : Bool
```

src/ameba/ast/visitors/implicit_return_visitor.cr
```diff
@@ -24,10 +24,24 @@ module Ameba::AST
...
+      node.right.accept(self)
+
+      false
+    end
+
     def visit(node : Crystal::Call) : Bool
       @rule.test(@source, node, @stack > 0)
 
-      if node.block
+      if node.block || node.args.size > 0
```
Sija, line 44:
```suggestion
      if node.block || !node.args.empty?
```

spec/ameba/rule/lint/unused_comparison_spec.cr
```diff
@@ -30,6 +30,21 @@ module Ameba::Rule::Lint
...
+    end
+
+    it "passes if a regex literal is part of a === or =~ comparison" do
+      expect_no_issues subject, <<-CRYSTAL
+        /f(o+)(bar?)/ === "fooba"
+        puts $~
+        "foobar" =~ /(o+)ba(r?)/
+        puts $1
+      CRYSTAL
+    end
```
Sija, line 46: This doesn't make much sense, since regex literals always have side-effects (setting `$N` variables and alike).

I think `RegexLiteral`-support should be removed altogether.

spec/ameba/rule/lint/unused_comparison_spec.cr
```diff
@@ -30,6 +30,21 @@ module Ameba::Rule::Lint
         CRYSTAL
     end
 
+    it "passes for comparisons inside '||' and '&&' where the other arg is a call" do
+      expect_no_issues subject, <<-CRYSTAL
+          IO.copy(in_var, out_var, len) == len || raise IO::EOFError.new
```
Sija, line 35: Keep the code in tests generic, please.

```suggestion
        foo(bar) == baz || raise "bat"
        foo(bar) == baz && raise "bat"
```

spec/ameba/rule/lint/unused_literal_spec.cr
```diff
@@ -52,6 +52,18 @@ module Ameba::Rule::Lint
         CRYSTAL
     end
 
+    it "passes if a literal is the object of a call with args" do
+      expect_no_issues subject, <<-CRYSTAL
+          { error: "abc" }.to_json(context.response)
```
Sija, line 57:
```suggestion
        { foo: "bar" }.to_json(io)
```

spec/ameba/rule/lint/unused_literal_spec.cr
```diff
@@ -52,6 +52,18 @@ module Ameba::Rule::Lint
...
+        CRYSTAL
+    end
+
+    it "passes for a literal in a generic type" do
+      expect_no_issues subject, <<-CRYSTAL
+          StaticArray(1, 5)
```
Sija, line 63:
```suggestion
        StaticArray(Int32, 3)
        Int32[3]
```

Sija (changes requested):

spec/ameba/rule/lint/unused_comparison_spec.cr
nobodywasishere: From what I can see looking at the stdlib, `$~` is only set when using these two operators (or the `match` methods), and `$1` is just a shorthand for `$~[1]`.

nobodywasishere:
I don't think the typos lint should check RegexLiterals, feel like that's just more prone to errors.

spec/ameba/rule/lint/unused_comparison_spec.cr
straight-shoota: Ameba probably should not make any assumptions about implementations in stdlib. And without semantic analysis there's no way to know whether any call even targets the stdlib implementation.

Sija:
@nobodywasishere It doesn't matter since the support for them should be removed anyway.

src/ameba/ast/visitors/implicit_return_visitor.cr
```diff
@@ -258,6 +274,18 @@ module Ameba::AST
       true
     end
 
+    def visit(node : Crystal::Yield) : Bool
```
Sija, line 277: Is there any test for that?

nobodywasishere:
@Sija I'm referring to the failing CI, which is failing irregardless of this PR / the unused comparison/literal lints

spec/ameba/rule/lint/unused_comparison_spec.cr
nobodywasishere: That's true, so for unused comparison we should check to see if either the obj or the arg is a regex, and not apply it if so. But in the case of unused literal, there isn't a situation in which a regex literal by itself in a void context will have side effects, right?

straight-shoota: > That's true, so for unused comparison we should check to see if either the obj or the arg is a regex, and not apply it if so.

That would still cause false positive for non-literal instances of `Regex`, such as `re = /(.)oo/; re === "foo"`.
Without semantical analysis this cannot be solved correctly.
The best course of action is probably to drop `===` and `~=` from the list of tested comparison operators.

Other operators should generally not be affected, at least not in stdlib. The semantics defined in the language itself don't suggest any realistic use case with side effects, so they should probably be acceptable.

straight-shoota: > But in the case of unused literal, there isn't a situation in which a regex literal by itself in a void context will have side effects, right?

Probably not. A literal expands to a call to `Regex.literal` which compiles the regex. That *might* cause some side effect inside the regex engine, but that shouldn't be critical.

src/ameba/rule/lint/unused_comparison.cr
```diff
@@ -53,6 +53,9 @@ module Ameba::Rule::Lint
 
     def test(source, node : Crystal::Call, last_is_used : Bool)
       if !last_is_used && node.name.in?(COMPARISON_OPERATORS) && node.args.size == 1
+        return if node.obj.is_a?(Crystal::RegexLiteral) ||
+                  node.args.first.is_a?(Crystal::RegexLiteral)
```
Sija, line 57: Shouldn't it check all arguments?

```suggestion
                  node.args.any?(Crystal::RegexLiteral)
```

spec/ameba/rule/lint/unused_comparison_spec.cr
```diff
@@ -30,18 +30,27 @@ module Ameba::Rule::Lint
...
+        foo(bar) == baz || raise "bat"
+        foo(bar) == baz && raise "bat"
+        CRYSTAL
+    end
+
+    it "passes for unused comparisons with `===`, `=~`, and `!~`" do
+      expect_no_issues subject, <<-CRYSTAL
+        /foo(bar)?/ =~ baz
+        "foobar" !~ /(o+)ba(r?)/
+        "hello" === world
```
Sija, line 44:
```suggestion
        /foo(bar)?/ =~ baz
        /foo(bar)?/ !~ baz
        "hello" === world
```

veelenga (approved):
LGTM, thanks

Sija (approved):

PR #534, opened 2025-01-05
Add `Rule::Base.autocorrect_incompatible_with`

### LINKED ISSUE #489: The `max` after the `map` is removed

FnControlOption:
Fixes #489

The issue here is that the autocorrections for `Style/VerboseBlock` and `Performance/MinMaxAfterMap` are overlapping.

For `Performance/MinMaxAfterMap`, we (1) replace `map` with `max_of` and (2) remove `.max`.

```
list.map { |i| i.size }.max
     ^^^                ^^^
     (1)                (2)
```

(1) is overwritten by `Style/VerboseBlock`, which replaces `map { |i| i.size }` with `map(&.size)`.

```
list.map { |i| i.size }.max
     ^^^^^^^^^^^^^^^^^^ ^^^
        map(&.size)     (2)
```

As a result, we get

```
list.map(&.size)
```

My proposed fix is to run incompatible autocorrections separately instead of all at once.

So first we'll correct `Performance/MinMaxAfterMap`

```
list.map { |i| i.size }.max
     ^^^                ^^^
     (1)                (2)
```

to get

```
list.max_of { |i| i.size }
```

Then, we'll correct `Style/VerboseBlock` to get

```
list.max_of(&:size)
```


Sija:
Thanks for the quick fix! ❤️ Regarding the implementation I'm afraid it's not scalable, since it'd require analyzing all of the rules in order to _compile_ a matrix of dependent (and incompatible) rules. I was thinking of doing that automatically, as it should be possible given that the corrector knows the positions and sizes of the applied corrections.

PR #536, opened 2025-01-12
Add support for linting ECR files

### LINKED ISSUE #509: Add support for linting ECR files
This should be a simple process of checking if the filename ends with `.ecr`, then calling `ECR.process_file` or `ECR.process_string` (with a clean rescue in case an error occurs), and running the linter over the outputted code.

nobodywasishere:
Requires Crystal >= 1.15.0. Lints ECR files by default. Closes #509

### DIFF
--- a/src/ameba.cr
+++ b/src/ameba.cr
@@ -40,4 +40,8 @@ module Ameba
   def run(config = Config.load)
     Runner.new(config).run
   end
+
+  def self.ecr_supported? : Bool
+    {{ compare_versions(Crystal::VERSION, "1.15.0") >= 0 }}
+  end
 end
--- a/src/ameba/config.cr
+++ b/src/ameba/config.cr
@@ -1,5 +1,6 @@
 require "semantic_version"
 require "yaml"
+require "ecr/processor"
 require "./glob_utils"
 
 # A configuration entry for `Ameba::Runner`.
@@ -57,10 +58,18 @@ class Ameba::Config
     Path[XDG_CONFIG_HOME] / "ameba/config.yml",
   }
 
-  DEFAULT_GLOBS = %w(
-    **/*.cr
-    !lib
-  )
+  {% if compare_versions(Crystal::VERSION, "1.15.0") >= 0 %}
+    DEFAULT_GLOBS = %w(
+      **/*.cr
+      **/*.ecr
+      !lib
+    )
+  {% else %}
+    DEFAULT_GLOBS = %w(
+      **/*.cr
+      !lib
+    )
+  {% end %}
 
   getter rules : Array(Rule::Base)
   property severity = Severity::Convention
@@ -167,7 +176,7 @@ class Ameba::Config
   # ```
   # config = Ameba::Config.load
   # config.sources # => list of default sources
-  # config.globs = ["**/*.cr"]
+  # config.globs = ["**/*.cr", "**/*.ecr"]
   # config.excluded = ["spec"]
   # config.sources # => list of sources pointing to files found by the wildcards
   # ```
--- a/src/ameba/config.cr
+++ b/src/ameba/config.cr
@@ -58,18 +58,14 @@ class Ameba::Config
     Path[XDG_CONFIG_HOME] / "ameba/config.yml",
   }
 
-  {% if compare_versions(Crystal::VERSION, "1.15.0") >= 0 %}
-    DEFAULT_GLOBS = %w(
-      **/*.cr
-      **/*.ecr
-      !lib
-    )
-  {% else %}
-    DEFAULT_GLOBS = %w(
-      **/*.cr
-      !lib
-    )
-  {% end %}
+  DEFAULT_GLOBS = %w(
+    **/*.cr
+    !lib
+  )
+
+  if Ameba.ecr_supported?
+    DEFAULT_GLOBS << "**/*.ecr"
+  end
 
   getter rules : Array(Rule::Base)
   property severity = Severity::Convention
--- a/src/ameba/glob_utils.cr
+++ b/src/ameba/glob_utils.cr
@@ -24,13 +24,13 @@ module Ameba
     def expand(globs)
       globs
         .flat_map do |glob|
-        if File.directory?(glob)
-          glob += "/**/*.cr"
+          if File.directory?(glob)
+            glob += "/**/*.cr"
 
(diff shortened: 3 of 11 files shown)


src/ameba/source.cr
```diff
@@ -57,6 +57,24 @@ module Ameba
     # source.ast
     # ```
     getter ast : Crystal::ASTNode do
+      code = @code
+
+      {% if compare_versions(Crystal::VERSION, "1.15.0") >= 0 %}
+        if @path.ends_with?(".ecr")
+          begin
+            code = ECR.process_string(code, @path)
+          rescue ex : Crystal::SyntaxException
```
Sija, line 66:
```suggestion
          rescue ex : ECR::Lexer::SyntaxException
```

Sija (changes requested):

Sija:
Specs would be good to have 🙏

src/ameba/glob_utils.cr
```diff
@@ -25,6 +25,11 @@ module Ameba
       globs
         .flat_map do |glob|
           glob += "/**/*.cr" if File.directory?(glob)
+
+          {% if compare_versions(Crystal::VERSION, "1.15.0") >= 0 %}
+            glob += "/**/*.ecr" if File.directory?(glob)
+          {% end %}
```
Sija, line 31:
```suggestion
        if File.directory?(glob)
          glob += "/**/*.cr"

          {% if compare_versions(Crystal::VERSION, "1.15.0") >= 0 %}
            glob += "/**/*.ecr"
          {% end %}
        end
```

Sija (changes requested):

veelenga:
@nobodywasishere I may have missed some conversation, but does this conflict with `Lint/UselessAssign`?

nobodywasishere:
I don't know if there's been a conversation about that, but I do see how this could conflict with `Lint/UselessAssign`. As ECR files are expanded as Crystal code in-place, this can lead to stuff like this:

```crystal
def render_page
  my_var = 1
  # ^^^^ error: Useless assignment to `my_var`
  render("path/to/page.ecr)
end
```

```erb
<%= my_var %>
```

This is already an issue though without this PR.

Another one specific to ECR though is something like:

```erb
<%
  my_var ||= 2
  # ^^^^ Lint/Syntax: Cannot use `my_var` in assignment to `my_var`
%>
<%= my_var %>
```

Which means templates just need to be written in a way where they're parsable outside the context of where they're rendered.

---

One cool thing I found while testing is that `Lint/UnusedLiteral` works:

```erb
<% 
  "hello world" 
  # ^^^^^^^^^^^ error: unused literal
%>
```

So it will give a warning if you accidentally leave off the `=`.

nobodywasishere:
I've gone through all the rules and all the current auto corrections should be fine. Anything that changes stuff over the `<% / %>` boundary though would be a problem. For example, if there was a rule that commented out the contents of an `if` for some reason:

Original ECR:
```erb
<% if true %>
  <%= "hello" %>
<% end %>
```

Parsed ECR (without loc pragmas):
```cr
if true
  __str__ << "\n  "
  ("hello").to_s(__str__)
  __str__ << "\n"
end
```

"Corrected" source:
```cr
if true
  # __str__ << "\n  "
  # ("hello").to_s(__str__)
  # __str__ << "\n"
end
```

Could completely break or mangle the original ECR. This shouldn't be a problem with the current rules though.

Sija:
By specs I meant testing the ECR support, not adding tests to the rule specs.

src/ameba/source.cr
```diff
@@ -57,6 +57,24 @@ module Ameba
     # source.ast
     # ```
     getter ast : Crystal::ASTNode do
+      code = @code
+
+      {% if compare_versions(Crystal::VERSION, "1.15.0") >= 0 %}
```
veelenga, line 62: since this is copied in multiple places, what do you thing about following more declarative way and defining the helper method. Something like:

```crystal
def ecr_supported?
  {{ compare_versions(Crystal::VERSION, "1.15.0") >= 0 }}
end
```

nobodywasishere:
I thought having some tests that ensure functionality within an ECR file would be useful, but I can remove them if desired.

src/ameba.cr
```diff
@@ -40,4 +40,8 @@ module Ameba
   def run(config = Config.load)
     Runner.new(config).run
   end
+
+  def self.ecr_supported? : Bool
+    {{ compare_versions(Crystal::VERSION, "1.15.0") >= 0 }}
+  end
```
veelenga, line 46: I suggest moving this to config, but may be subjective. @Sija wdyt ?

nobodywasishere: My thought was if they didn't want to lint ECR files, they can add `**/*.ecr` to `Excluded`

veelenga (approved):
lgtm

src/ameba/config.cr
```diff
@@ -57,10 +58,18 @@ class Ameba::Config
     Path[XDG_CONFIG_HOME] / "ameba/config.yml",
   }
 
-  DEFAULT_GLOBS = %w(
-    **/*.cr
-    !lib
-  )
+  {% if compare_versions(Crystal::VERSION, "1.15.0") >= 0 %}
```
Sija, line 61: This could use the helper method as well, right?

Sija (changes requested):

src/ameba/config.cr
nobodywasishere: Yeah if I change it to appending to the array after it's created

nobodywasishere: If we want to use `Ameba.ecr_supported?` here, I can replace this with:
```cr
  DEFAULT_GLOBS = %w(
    **/*.cr
    !lib
  )

  if Ameba.ecr_supported?
    DEFAULT_GLOBS << "**/*.ecr"
  end
```

Sija (approved):

Sija:
This needs another go-around, see https://github.com/crystal-ameba/ameba/actions/runs/12782021496/job/35630866247#step:7:1

nobodywasishere:
I've no idea what caused that that wouldn't also work locally... I can try moving the location of the `require "ecr/processor"` to see if that'd work, but other than that unless it's building using an older version of Crystal (<1.15), I've no idea.

nobodywasishere:
Yeah alpine is still using Crystal 1.14 https://pkgs.alpinelinux.org/package/edge/community/x86_64/crystal

PR #537, opened 2025-01-15
Add `Lint/UnusedGenericOrUnion`

nobodywasishere:
Similar to `Lint/UnusedLiteral` and `Lint/UnusedComparison`. Not too fond of the name so any suggestions are welcome.

### DIFF
--- a/spec/ameba/rule/lint/unused_generic_or_union_spec.cr
+++ b/spec/ameba/rule/lint/unused_generic_or_union_spec.cr
@@ -3,6 +3,7 @@ require "../../../spec_helper"
 module Ameba::Rule::Lint
   describe UnusedGenericOrUnion do
     subject = UnusedGenericOrUnion.new
+
     it "passes if generics and unions are used for assign and method calls" do
       expect_no_issues subject, <<-CRYSTAL
         my_var : String? = EMPTY_STRING
@@ -47,11 +48,11 @@ module Ameba::Rule::Lint
     it "fails if generics or unions are unused at top-level" do
       expect_issue subject, <<-CRYSTAL
         String?
-        # ^^^^^ error: Generic or union is not used
+        # ^^^^^ error: Generic is not used
         Int32 | Float64 | Nil
-        # ^^^^^^^^^^^^^^^^^^^ error: Generic or union is not used
+        # ^^^^^^^^^^^^^^^^^^^ error: Union is not used
         StaticArray(Int32, 10)
-        # ^^^^^^^^^^^^^^^^^^^^ error: Generic or union is not used
+        # ^^^^^^^^^^^^^^^^^^^^ error: Generic is not used
         CRYSTAL
     end
 
@@ -59,13 +60,13 @@ module Ameba::Rule::Lint
       expect_issue subject, <<-CRYSTAL
         def hello
           Float64?
-        # ^^^^^^^^ error: Generic or union is not used
+        # ^^^^^^^^ error: Generic is not used
           0.1
         end
 
         fun fun_name : Int32
           Array(String)
-        # ^^^^^^^^^^^^^ error: Generic or union is not used
+        # ^^^^^^^^^^^^^ error: Generic is not used
           1234
         end
         CRYSTAL
@@ -75,24 +76,24 @@ module Ameba::Rule::Lint
       expect_issue subject, <<-CRYSTAL
         class MyClass
           String?
-        # ^^^^^^^ error: Generic or union is not used
+        # ^^^^^^^ error: Generic is not used
           Array(self)
-        # ^^^^^^^^^^^ error: Generic or union is not used
+        # ^^^^^^^^^^^ error: Generic is not used
           Array(typeof(1))
-        # ^^^^^^^^^^^^^^^^ error: Generic or union is not used
+        # ^^^^^^^^^^^^^^^^ error: Generic is not used
 
           def hello
             self | Nil
-          # ^^^^^^^^^^ error: Generic or union is not used
+          # ^^^^^^^^^^ error: Union is not used
             typeof(1) | Nil | _
-          # ^^^^^^^^^^^^^^^^^^^ error: Generic or union is not used
+          # ^^^^^^^^^^^^^^^^^^^ error: Union is not used
             "Hello, Gordon!"
           end
         end
 
         module MyModule
           Array(Int32)
-        # ^^^^^^^^^^^^ error: Generic or union is not used
+        # ^^^^^^^^^^^^ error: Generic is not used
         end
         CRYSTAL
     end
--- a/spec/ameba/rule/lint/unused_generic_or_union_spec.cr
+++ b/spec/ameba/rule/lint/unused_generic_or_union_spec.cr
@@ -1,26 +1,25 @@
 require "../../../spec_helper"
 
 module Ameba::Rule::Lint
-  subject = UnusedTypeOrConstant.new
+  subject = UnusedGenericOrUnion.new
 
-  describe UnusedTypeOrConstant do
-    it "passes if paths and generic types are used for assign and method calls" do
+  describe UnusedGenericOrUnion do
+    it "passes if generics and unions are used for assign and method calls" do
       expect_no_issues subject, <<-CRYSTAL
-        MyConst = 1
-
         my_var : String? = EMPTY_STRING
 
-        a : Int32 = 10
+        a : Int32? = 10
 
         klass = String?
 
-        my_var.as(String)
+        my_var.as(Array(Char))
 
         puts StaticArray(Int32, 10)
 
+        # Not a union
         Int32 | "Float64"
 
-        MyClass.run
+        MyClass(String).new.run
         CRYSTAL
     end
 
--- a/spec/ameba/rule/lint/unused_generic_or_union_spec.cr
+++ b/spec/ameba/rule/lint/unused_generic_or_union_spec.cr
@@ -23,6 +23,16 @@ module Ameba::Rule::Lint
         CRYSTAL
     end
 
+    it "passes plain pseudo methods, self, and paths" do
+      expect_no_issues subject, <<-CRYSTAL
+        _
+        self
+        self?
+        typeof(1)
+        Int32
+        CRYSTAL
+    end
+
     it "passes if generics and unions are used for methods" do
       expect_no_issues subject, <<-CRYSTAL
         def size(a : Float64?) : Float64?
@@ -75,6 +85,8 @@ module Ameba::Rule::Lint
           def hello
             self | Nil
           # ^^^^^^^^^^ error: Generic or union is not used
+            typeof(1) | Nil | _
+          # ^^^^^^^^^^^^^^^^^^^ error: Generic or union is not used
             "Hello, Gordon!"
           end
         end
--- a/spec/ameba/rule/lint/unused_generic_or_union_spec.cr
+++ b/spec/ameba/rule/lint/unused_generic_or_union_spec.cr
@@ -1,9 +1,8 @@
 require "../../../spec_helper"
 
 module Ameba::Rule::Lint
-  subject = UnusedGenericOrUnion.new
-
   describe UnusedGenericOrUnion do
+    subject = UnusedGenericOrUnion.new
     it "passes if generics and unions are used for assign and method calls" do
       expect_no_issues subject, <<-CRYSTAL
         my_var : String? = EMPTY_STRING
@@ -45,7 +44,7 @@ module Ameba::Rule::Lint
         CRYSTAL
     end
 
-    it "fails if generics or unions are unused top-level" do
+    it "fails if generics or unions are unused at top-level" do
       expect_issue subject, <<-CRYSTAL
         String?
         # ^^^^^ error: Generic or union is not used
--- a/spec/ameba/rule/lint/unused_generic_or_union_spec.cr
+++ b/spec/ameba/rule/lint/unused_generic_or_union_spec.cr
@@ -22,7 +22,7 @@ module Ameba::Rule::Lint
         CRYSTAL
     end
 
-    it "passes plain pseudo methods, self, and paths" do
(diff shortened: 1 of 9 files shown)


nobodywasishere:
`Lint/UnusedTypeOrConstant` may be a better name

spec/ameba/rule/lint/unused_type_or_constant_spec.cr
```diff
@@ -0,0 +1,80 @@
...
+        fun fun_name : Int32
+          Int32
+        # ^^^^^ error: Path or generic type is not used
+          1234
+        end
+
+        class MyClass
+          MyModule
+        # ^^^^^^^^ error: Path or generic type is not used
+        end
```
veelenga, line 76: these are not "top level" as mentioned in the spec name. Can we move it to a separate testcase?

spec/ameba/rule/lint/unused_type_or_constant_spec.cr
```diff
@@ -0,0 +1,80 @@
...
+
+        fun fun_name : Int32
+          1234
+        end
+
+        Int32 | "Float64"
+
+        MyClass.run
+        CRYSTAL
+    end
```
veelenga, line 48: i would suggest to slipt this into smaller testcases as well. Otherwise, new examples will just be added to the same big testcase in the future.

src/ameba/ast/visitors/implicit_return_visitor.cr
```diff
@@ -282,7 +319,30 @@ module Ameba::AST
...
+    def visit(node : Crystal::Alias)
+      @rule.test(@source, node, @stack > 0)
+
+      incr_stack do
+        node.name.accept(self)
+        node.value.accept(self)
+      end
+
+      false
+    end
```
veelenga, line 337: Could you recall how this visitor functions, particularly why we increment a stack and the meaning of `last_is_used` in the rules? It might be helpful to document this visitor for better clarity, as it's currently challenging to understand it quickly.

nobodywasishere: I've added some documentation, let me know if it makes sense.

nobodywasishere:
There's no way to differentiate paths that point to class names vs constants without semantic analysis, so we'll need to drop support for paths in general then. I'll rename to `Lint/UnusedGenericOrUnion`

src/ameba/rule/lint/unused_generic_or_union.cr
```diff
@@ -0,0 +1,87 @@
...
+      when Crystal::Var
+        return false unless rhs.name == "self"
+      when Crystal::Call
+        return false unless (rhs.name == "self") || path_or_generic_union?(rhs)
+      else
+        return false
+      end
+
+      true
+    end
```
veelenga, line 85: can we avoid duplication here? Maybe something like this?

```crystal
def path_or_generic_union?(node : Crystal::Call) : Bool
  return false unless node.name == "|" && node.args.size == 1

  valid_type_node?(node.obj) && valid_type_node?(node.args.first?)
end

private def valid_type_node?(node)
  case node
  when Crystal::Path, Crystal::Generic, Crystal::Self, Crystal::TypeOf, Crystal::Underscore
    true
  when Crystal::Var
    node.name == "self"
  when Crystal::Call
    node.name == "self" || path_or_generic_union?(node)
  else
    false
  end
end

``` 

PS: didn't test, just suggesting

src/ameba/rule/lint/unused_generic_or_union.cr
```diff
@@ -0,0 +1,87 @@
...
+      return if node_is_used || !path_or_generic_union?(node)
+
+      issue_for node, MSG
+    end
+
+    def test(source, node : Crystal::Generic | Crystal::Union, node_is_used : Bool)
+      issue_for node, MSG unless node_is_used
+    end
+
+    def path_or_generic_union?(node : Crystal::Call) : Bool
```
veelenga, line 59: should be a private method i guess

src/ameba/rule/lint/unused_generic_or_union.cr
```diff
@@ -0,0 +1,79 @@
...
+    end
+
+    private def valid_type_node?(node : Crystal::ASTNode?) : Bool
+      case node
+      when Crystal::Path, Crystal::Generic, Crystal::Self, Crystal::TypeOf, Crystal::Underscore
+        true
+      when Crystal::Var
+        node.name == "self"
+      when Crystal::Call
+        (node.name == "self") || path_or_generic_union?(node)
```
Sija, line 73: This would target `foo.self` which we don't want.

```suggestion
        path_or_generic_union?(node)
```

Sija (changes requested):

spec/ameba/rule/lint/unused_generic_or_union_spec.cr
```diff
@@ -0,0 +1,101 @@
...
+      expect_no_issues subject, <<-CRYSTAL
+        def size(a : Float64?) : Float64?
+          0.1.try(&.+(a))
+        end
+
+        def append(a : Array(String)) : Array(String)
+          a << "hello"
+        end
+        CRYSTAL
+    end
```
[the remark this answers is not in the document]
nobodywasishere: Was necessary when I was writing this as I was ending up with cases where paths / generics were used but the implicit return visitor wasn't counting them as used.

veelenga (approved):
lgtm

spec/ameba/rule/lint/unused_generic_or_union_spec.cr
```diff
@@ -0,0 +1,100 @@
+require "../../../spec_helper"
+
+module Ameba::Rule::Lint
+  describe UnusedGenericOrUnion do
+    subject = UnusedGenericOrUnion.new
```
Sija, line 5:
```suggestion
    subject = UnusedGenericOrUnion.new

```

spec/ameba/rule/lint/unused_generic_or_union_spec.cr
```diff
@@ -0,0 +1,100 @@
...
+      expect_no_issues subject, <<-CRYSTAL
+        _
+        self
+        self?
+        typeof(1)
+        Int32
+        CRYSTAL
+    end
+
+    it "passes if generics and unions are used for methods" do
```
Sija, line 35:
```suggestion
    it "passes if generics and unions are used for method arguments and return type" do
```

Sija (changes requested):

spec/ameba/rule/lint/unused_generic_or_union_spec.cr
```diff
@@ -0,0 +1,100 @@
...
+        puts StaticArray(Int32, 10)
+
+        # Not a union
+        Int32 | "Float64"
+
+        MyClass(String).new.run
+        CRYSTAL
+    end
+
+    it "passes plain pseudo methods, self, and paths" do
```
Sija, line 25:
```suggestion
    it "passes for plain pseudo methods, self, and paths" do
```

src/ameba/rule/lint/unused_generic_or_union.cr
```diff
@@ -0,0 +1,77 @@
...
+  # Lint/UnusedGenericOrUnion:
+  #   Enabled: true
+  # ```
+  class UnusedGenericOrUnion < Base
+    properties do
+      since_version "1.7.0"
+      description "Disallows unused generics or unions"
+    end
+
+    MSG = "Generic or union is not used"
```
Sija, line 43: I'd prefer to customize the message depending on the case.

src/ameba/rule/lint/unused_generic_or_union.cr
```diff
@@ -0,0 +1,78 @@
...
+      issue_for node, UNION_MSG
+    end
+
+    def test(source, node : Crystal::Generic, node_is_used : Bool)
+      issue_for node, GENERIC_MSG unless node_is_used
+    end
+
+    private def path_or_generic_union?(node : Crystal::Call) : Bool
+      node.name == "|" && node.args.size == 1 &&
+        valid_type_node?(node.obj) && valid_type_node?(node.args.first?)
```
Sija, line 62:
```suggestion
        valid_type_node?(node.obj) && valid_type_node?(node.args.first)
```

Sija:
```suggestion
      node.name == "|" && node.args.size == 1 && (obj = node.obj) &&
        valid_type_node?(obj) && valid_type_node?(node.args.first)
```

src/ameba/rule/lint/unused_generic_or_union.cr
```diff
@@ -0,0 +1,78 @@
...
+    end
+
+    private def valid_type_node?(node : Crystal::ASTNode?) : Bool
```
Sija, line 65:
```suggestion
    private def valid_type_node?(node : Crystal::ASTNode) : Bool
```

src/ameba/rule/lint/unused_generic_or_union.cr
```diff
@@ -0,0 +1,78 @@
...
+      issue_for node, UNION_MSG
+    end
+
+    def test(source, node : Crystal::Generic, node_is_used : Bool)
+      issue_for node, GENERIC_MSG unless node_is_used
+    end
+
+    private def path_or_generic_union?(node : Crystal::Call) : Bool
+      node.name == "|" && node.args.size == 1 &&
+        valid_type_node?(node.obj) && valid_type_node?(node.args.first)
```
Sija, line 62:
```suggestion
      node.name == "|" && node.args.size == 1 && (obj = node.obj)
        valid_type_node?(obj) && valid_type_node?(node.args.first)
```

src/ameba/rule/lint/unused_generic_or_union.cr
```diff
@@ -0,0 +1,78 @@
...
+
+      issue_for node, UNION_MSG
+    end
+
+    def test(source, node : Crystal::Generic, node_is_used : Bool)
+      issue_for node, GENERIC_MSG unless node_is_used
+    end
+
+    private def path_or_generic_union?(node : Crystal::Call) : Bool
+      node.name == "|" && node.args.size == 1 && (obj = node.obj)
```
Sija, line 61:
```suggestion
      node.name == "|" && node.args.size == 1 && (obj = node.obj) &&
```

src/ameba/rule/lint/unused_generic_or_union.cr
```diff
@@ -0,0 +1,78 @@
...
+      return if node_is_used || !path_or_generic_union?(node)
+
+      issue_for node, UNION_MSG
+    end
+
+    def test(source, node : Crystal::Generic, node_is_used : Bool)
+      issue_for node, GENERIC_MSG unless node_is_used
+    end
+
+    private def path_or_generic_union?(node : Crystal::Call) : Bool
```
Sija, line 60:
```suggestion
    private def path_or_generic_union?(node : Crystal::Call)
```

src/ameba/rule/lint/unused_generic_or_union.cr
```diff
@@ -0,0 +1,78 @@
...
+    end
+
+    def test(source, node : Crystal::Generic, node_is_used : Bool)
+      issue_for node, GENERIC_MSG unless node_is_used
+    end
+
+    private def path_or_generic_union?(node : Crystal::Call)
+      node.name == "|" && node.args.size == 1 && (obj = node.obj) &&
+        valid_type_node?(obj) && valid_type_node?(node.args.first)
+    end
```
Sija, line 63:
```suggestion
    private def path_or_generic_union?(node : Crystal::Call) : Bool
      node.name == "|" && node.args.size == 1 && !!(obj = node.obj) &&
        valid_type_node?(obj) && valid_type_node?(node.args.first)
    end
```

Sija: This method actually only checks for unions, not generics nor paths.

src/ameba/rule/lint/unused_generic_or_union.cr
```diff
@@ -0,0 +1,78 @@
...
+
+    private def path_or_generic_union?(node : Crystal::Call) : Bool
+      node.name == "|" && node.args.size == 1 && !!(obj = node.obj) &&
+        valid_type_node?(obj) && valid_type_node?(node.args.first)
+    end
+
+    private def valid_type_node?(node : Crystal::ASTNode) : Bool
+      case node
+      when Crystal::Path, Crystal::Generic, Crystal::Self, Crystal::TypeOf, Crystal::Underscore
```
Sija, line 67: Shouldn't (`Crystal::...`) `SizeOf`, `InstanceSizeOf`, `AlignOf`, `InstanceAlignOf`, `OffsetOf`, `PointerOf` be included here as well?

nobodywasishere: I didn't as those aren't allowed in type restrictions (which this is supposed to check). Those would fall under `Lint/UnusedPseudoMethod` imo

Sija: Unless I'm missing something this rule ought to check for unused usages of generic types and unions, not type restrictions?

nobodywasishere: That's correct - if i'm remembering correctly unions are only allowed inside type restrictions

nobodywasishere: https://carc.in/#/r/hmnp

`String | Int32` outside of a type restriction is a method call on String that creates a union, that's what we're checking for here

nobodywasishere: Those language constructs don't create unions though (see the carcin link above), so they should not be checked

Sija: You're right. In such a case `TypeOf` shouldn't be included there either.

Sija: My bad, that's correct.

PR #541, opened 2025-01-15
Add support for linting ECR files (pt. 2)

nobodywasishere:
Rework of #536

### DIFF
--- a/src/ameba.cr
+++ b/src/ameba.cr
@@ -1,11 +1,3 @@
-require "./ameba/*"
-require "./ameba/ast/**"
-require "./ameba/ext/**"
-require "./ameba/rule/**"
-require "./ameba/formatter/*"
-require "./ameba/presenter/*"
-require "./ameba/source/**"
-
 # Ameba's entry module.
 #
 # To run the linter with default parameters:
@@ -40,4 +32,18 @@ module Ameba
   def run(config = Config.load)
     Runner.new(config).run
   end
+
+  macro ecr_supported?(&)
+    {% if compare_versions(Crystal::VERSION, "1.15.0") >= 0 %}
+      {{ yield }}
+    {% end %}
+  end
 end
+
+require "./ameba/*"
+require "./ameba/ast/**"
+require "./ameba/ext/**"
+require "./ameba/rule/**"
+require "./ameba/formatter/*"
+require "./ameba/presenter/*"
+require "./ameba/source/**"
--- a/src/ameba/config.cr
+++ b/src/ameba/config.cr
@@ -1,5 +1,6 @@
 require "semantic_version"
 require "yaml"
+require "ecr/processor"
 require "./glob_utils"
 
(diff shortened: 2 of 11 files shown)


src/ameba.cr
```diff
@@ -40,4 +40,8 @@ module Ameba
   def run(config = Config.load)
     Runner.new(config).run
   end
+
+  def self.ecr_supported? : Bool
```
Sija, line 44: This needs to be a macro.

src/ameba/config.cr
```diff
@@ -1,5 +1,6 @@
 require "semantic_version"
 require "yaml"
+require "ecr/processor"
```
Sija, line 3: Why this is added _here_?

src/ameba/source.cr
```diff
@@ -57,6 +57,22 @@ module Ameba
     # source.ast
     # ```
     getter ast : Crystal::ASTNode do
+      code = @code
+
+      if Ameba.ecr_supported? && @path.ends_with?(".ecr")
+        begin
+          code = ECR.process_string(code, @path)
+        rescue ex : ECR::Lexer::SyntaxException
```
Sija, line 65: Looks like there's a `require` missing, covering that path.

Sija (changes requested):

src/ameba/source.cr
nobodywasishere: There is not, the problem is alpine is still using Crystal 1.14, which this requires 1.15

src/ameba.cr
nobodywasishere: How so?

src/ameba/config.cr
nobodywasishere: As this is where other requires for external things go and I tend to like having them all in the same place.

src/ameba/source.cr
Sija: Missing `require` here is not a problem indeed, yet the reason that builds fail on Crystal 1.14 is due to the fact, that the check for ecr support moved from compile-time to runtime, which breaks builds on Crystal < 1.15, that's why the `ecr_supported?` method needs to turned into a macro.

Sija (changes requested):

src/ameba.cr
```diff
@@ -40,4 +32,18 @@ module Ameba
   def run(config = Config.load)
     Runner.new(config).run
   end
+
+  macro ecr_supported?(&)
+    {% if compare_versions(Crystal::VERSION, "1.15.0") >= 0 %}
+      {{ yield }}
+    {% end %}
+  end
```
Sija, line 40: Why not simply?

```suggestion
  macro ecr_supported?
    {{ compare_versions(Crystal::VERSION, "1.15.0") >= 0 }}
  end
```

... followed by `if Ameba.ecr_supported?` in the call-sites?

Sija: My bad, nevermind that.

PR #544, opened 2025-01-24
In case scope queue is empty call the rule with the current scope

Sija:
This PR changes the behavior of `AST::ScopeVisitor` to call the affected rule with the current scope in case scope queue is empty.

### DIFF
--- a/spec/ameba/rule/lint/useless_assign_spec.cr
+++ b/spec/ameba/rule/lint/useless_assign_spec.cr
@@ -1011,6 +1011,18 @@ module Ameba::Rule::Lint
           CRYSTAL
       end
 
+      it "reports if it's not referenced at a top level + in a method" do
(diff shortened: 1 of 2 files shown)


veelenga (commented):
Could you please explain the reason for the change and the use cases it impacts?

Sija:
While reading the source of `AST::ScopeVisitor` I've stumbled upon multiple calls to `on_scope_end` in `end_visit` methods invoked for assignments, which seemed kinda out-of-place, since assignments are not creating any new scopes, at least not in the visitor. These calls seem to be necessary for top-level scoped cases in `Lint/UselessAssign` rule to work though, since the top-level scope is not tested by the visitor's underlying rule. Therefore I've changed the logic to do so in case there are no other scopes found.

veelenga (approved):
LGTM

PR #545, opened 2025-01-24
`Typing/MethodParameterTypeRestriction`: require types for double splat params

nobodywasishere:
For some reason thought this wasn't possible on my first pass. Should definitely be covered when this rule is enabled.

### DIFF
--- a/src/ameba/rule/typing/method_parameter_type_restriction.cr
+++ b/src/ameba/rule/typing/method_parameter_type_restriction.cr
@@ -48,6 +48,7 @@ module Ameba::Rule::Typing
   #   DefaultValue: false
   #   PrivateMethods: false
   #   ProtectedMethods: false
+  #   DoubleSplatParameters: false
   #   BlockParameters: false
   # ```
   class MethodParameterTypeRestriction < Base
@@ -58,6 +59,7 @@ module Ameba::Rule::Typing
       default_value false
       private_methods false
       protected_methods false
+      double_splat_parameters false
       block_parameters false
     end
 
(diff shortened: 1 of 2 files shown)


Sija:
Since it's quite unlikely you'd like to have such restriction (double splatted args are usually options, which include multiple types), I'm fine with this as long as it'll be guarded behind additional rule option.

src/ameba/rule/typing/method_parameter_type_restriction.cr
```diff
@@ -72,6 +72,10 @@ module Ameba::Rule::Typing
         issue_for arg, MSG
       end
 
+      if (arg = node.double_splat) && !(arg.restriction)
```
Sija, line 75:
```suggestion
      if (arg = node.double_splat) && !arg.restriction
```

Sija (changes requested):
Since it's quite unlikely you'd like to have such restriction (double splatted args are usually options, which include multiple types), I'm fine with this as long as it'll be guarded behind additional rule option.

nobodywasishere:
Have no idea how to reduce this without it feeling unnecessary

straight-shoota:
There are only three instance of a type restriction for double splat paramters that I can find in my locally available Crystal code bases. They're all related to `NamedTuple` operations: stdlib's `NamedTuple.new` and `#merge` and [`ResultSet#internal_read` from `crystal-db`](https://github.com/crystal-lang/crystal-db/blob/023dc5de90c11927656fc747601c5f08ea3c906f/src/db/result_set.cr#L136-L144), which essentially implements a `NamedTuple#transform_values`.
In all cases the type restriction is a free variable and used for inspection (and subsequently constructing an appropriate `NamedTuple` instance).

Contrasting this I found over a hundred instances of double splat parameters without type restrictions. And I suppose in the vast majority there's not much reason to add any restrictions because it would just be a free variable without any consequence.

So it feels this addition is rather unnecessary. Type restrictions on double splat parameters are rarely meaningful, so there's little reason for enforcing it.

nobodywasishere:
Sounds good. My only reasoning for wanting it is completeness but I'm fine not having it.

PR #546, opened 2025-01-24
Fix regression re: incorrect glob pattern for folders

nobodywasishere:


nobodywasishere:
There's currently a bug where the default `!lib` (and potentially any folder glob) isn't working correctly and causing it to report issues for files in `lib/` - this is because the original code is appending `/**/*.cr/**/*.ecr` to the folder name, instead of `/**/*.{cr,ecr}`.

Sija:
So from what I gather the case in question is rather incorrect glob, and not issue with ignoring `lib` directories?

nobodywasishere:
Yes technically, the former causing the latter

spec/ameba/glob_utils_spec.cr
```diff
@@ -23,6 +23,12 @@ module Ameba
           .should_not contain current_file_path
       end
 
+      it "doesn't return rejected folders" do
```
Sija, line 26: The method in question is `expand` and not `find_files_by_globs`.

Sija (changes requested):

veelenga (approved):
lgtm

Sija (approved):

PR #547, opened 2025-01-24
Add `Style/RedundantSelf` rule

### LINKED ISSUE #535: Rule `Style/RedundantSelf`
See Rubocop's [`Style/RedundantSelf`](https://github.com/rubocop/rubocop/blob/master/lib/rubocop/cop/style/redundant_self.rb) cop for reference.

Sija:
Resolves #535

### DIFF
--- a/src/ameba/ast/visitors/scope_calls_with_self_receiver_visitor.cr
+++ b/src/ameba/ast/visitors/scope_calls_with_self_receiver_visitor.cr
@@ -1,0 +1,129 @@
+module Ameba::AST
+  class ScopeCallsWithSelfReceiverVisitor < Crystal::Visitor
+    @current_scope : AST::Scope
+    @current_assign : Crystal::ASTNode?
+
+    getter scope_call_queue = {} of AST::Scope => Array(Crystal::Call)
+
+    def initialize(rule, source)
+      @current_scope = AST::Scope.new(source.ast) # top level scope
+
+      source.ast.accept self
+
+      scope_call_queue.each do |scope, calls|
+        calls.each do |call|
+          rule.test source, call, scope
+        end
+      end
+    end
+
+    private def on_scope_enter(node)
+      scope = AST::Scope.new(node, @current_scope)
+
+      @current_scope = scope
+    end
+
+    private def on_scope_end(node)
+      # go up if this is not a top level scope
+      if outer_scope = @current_scope.outer_scope
+        @current_scope = outer_scope
+      end
+    end
+
+    private def on_assign_end(target, node)
+      target.is_a?(Crystal::Var) &&
+        @current_scope.assign_variable(target.name, node)
+    end
+
+    # A main visit method that accepts `Crystal::ASTNode`.
+    # Returns `true`, meaning all child nodes will be traversed.
+    def visit(node : Crystal::ASTNode)
+      true
+    end
+
+    def end_visit(node : Crystal::ASTNode)
+      on_scope_end(node) if @current_scope.eql?(node)
+    end
+
+    def visit(node : Crystal::Def)
+      node.name == "->" || on_scope_enter(node)
+    end
+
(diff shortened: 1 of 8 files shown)


src/ameba/rule/style/redundant_self.cr
```diff
@@ -0,0 +1,239 @@
...
+      if location = obj.location
+        issue_for obj, MSG do |corrector|
+          corrector.remove(location, location.adjust(column_number: {{ "self".size }}))
+        end
+      else
+        issue_for obj, MSG
+      end
+    end
+
+    private class ScopeCallsWithSelfReceiverVisitor < Crystal::Visitor
```
Sija, line 111: There's a lot of duplication with `AST::ScopeVisitor`, thus an opportunity for a refactor.

nobodywasishere (approved):

src/ameba/rule/style/redundant_self.cr
veelenga: I recommend maintaining the original file structure by relocating this component to the visitors directory. Although it may seem that this class is currently used only by the existing rule, similar rules may leverage it in the future.

Sija: This one is pretty specific, that's why I would like to keep it next to the rule. In case it would be needed we can always move it.

Sija: Actually, I've thought this over and I've moved it aside, since it helps with the testing of the visitor class itself, apart of file organization.

src/ameba/ast/visitors/scope_calls_with_self_receiver_visitor.cr
```diff
@@ -0,0 +1,129 @@
+module Ameba::AST
+  class ScopeCallsWithSelfReceiverVisitor < Crystal::Visitor
```
veelenga, line 2: why don't we inherit from BaseVisitor and follow the existing visitor structure?

src/ameba/rule/style/redundant_self.cr
```diff
@@ -0,0 +1,110 @@
+require "compiler/crystal/syntax/token"
```
veelenga, line 1: why do we need to have this here?

src/ameba/rule/style/redundant_self.cr
```diff
@@ -0,0 +1,110 @@
...
+
+module Ameba::Rule::Style
```
veelenga, line 3: I’m wondering if this would be better placed under Lint group, since it’s a code smell rather than a style issue.

src/ameba/rule/style/redundant_self.cr
```diff
@@ -0,0 +1,110 @@
...
+
+      return if name.in?(vars)
+
+      if location = obj.location
+        issue_for obj, MSG do |corrector|
+          corrector.remove(location, location.adjust(column_number: {{ "self".size }}))
+        end
+      else
+        issue_for obj, MSG
+      end
```
veelenga, line 107: do we know the examples when location could be missing here?

veelenga (commented):
Nice one. Added some unblocking comments

src/ameba/rule/style/redundant_self.cr
Sija: Because of `CRYSTAL_KEYWORDS = Crystal::Keyword.values.map(&.to_s)` declaration

src/ameba/rule/style/redundant_self.cr
Sija: Not that I can think of, but it's a nilable property anyway.

src/ameba/rule/style/redundant_self.cr
Sija: Personally I would lean on the side of `Style` group, also because of how RuboCop handles it.

veelenga (approved):
LGTM

PR #549, opened 2025-01-25
ExceptionHandler implicit return fixes

nobodywasishere:
Last line of `ensure` isn't returned and last line of body isn't returned if there's an `else`.


veelenga (approved):
Good catch 👏🏻

PR #552, opened 2025-01-29
Add `Lint/Unused{Local,Instance,Class}VariableAccess`

### LINKED ISSUE #542: Add `Lint/Unused{Local,Instance,Class}VarAccess`
As local, instance, and class vars are known and cannot point to methods, we can add rules that will report if a local / instance / class var is in a void context:

```crystal
a = 1

begin
  a
# ^ error: local variable access in void context
  puts a
end

class MyClass
  @world : Int32 = 0

  def hello
    @world
    # ^^^^ error: instance var access in void context
    @world + 1
  end
end
```

nobodywasishere:
Closes #542

### DIFF
--- a/spec/ameba/rule/lint/unused_class_variable_access_spec.cr
+++ b/spec/ameba/rule/lint/unused_class_variable_access_spec.cr
@@ -1,0 +1,44 @@
+require "../../../spec_helper"
+
+module Ameba::Rule::Lint
+  subject = UnusedClassVariableAccess.new
+
+  describe UnusedClassVariableAccess do
+    it "passes if class variables are used" do
+      expect_no_issues subject, <<-CRYSTAL
+        class MyClass
+          a = @@ivar
+          B = @@ivar
+        end
+        CRYSTAL
+    end
+
+    it "passes if an class variable is used as a target in multi-assignment" do
+      expect_no_issues subject, <<-CRYSTAL
+        class MyClass
+          @@a, @@b = 1, 2
+        end
+        CRYSTAL
+    end
+
+    it "fails if class variables are unused" do
+      expect_issue subject, <<-CRYSTAL
+        class Actor
+          @@name : String = "George"
+
+          @@name
+        # ^^^^^^ error: Value from class variable access is unused
+
+          puts @@name
+
+          def hello : String
+            @@name
+          # ^^^^^^ error: Value from class variable access is unused
+
+            "Hello, \#{@@name}!"
+          end
+        end
+        CRYSTAL
+    end
+  end
+end
--- a/spec/ameba/rule/lint/unused_instance_variable_access_spec.cr
+++ b/spec/ameba/rule/lint/unused_instance_variable_access_spec.cr
@@ -1,0 +1,44 @@
+require "../../../spec_helper"
+
+module Ameba::Rule::Lint
+  subject = UnusedInstanceVariableAccess.new
+
+  describe UnusedInstanceVariableAccess do
+    it "passes if instance variables are used" do
+      expect_no_issues subject, <<-CRYSTAL
+        class MyClass
+          a = @ivar
+          B = @ivar
+        end
+        CRYSTAL
+    end
+
+    it "passes if an instance variable is used as a target in multi-assignment" do
+      expect_no_issues subject, <<-CRYSTAL
+        class MyClass
+          @a, @b = 1, 2
+        end
+        CRYSTAL
+    end
+
+    it "fails if instance variables are unused" do
+      expect_issue subject, <<-CRYSTAL
+        class Actor
+          @name : String = "George"
+
+          @name
+        # ^^^^^ error: Value from instance variable access is unused
+
+          puts @name
+
+          def hello : String
+            @name
+          # ^^^^^ error: Value from instance variable access is unused
+
+            "Hello, \#{@name}!"
+          end
+        end
+        CRYSTAL
+    end
+  end
+end
--- a/spec/ameba/rule/lint/unused_local_variable_access_spec.cr
+++ b/spec/ameba/rule/lint/unused_local_variable_access_spec.cr
@@ -1,0 +1,46 @@
+require "../../../spec_helper"
+
+module Ameba::Rule::Lint
+  subject = UnusedLocalVariableAccess.new
+
+  describe UnusedLocalVariableAccess do
+    it "passes if local variables are used" do
+      expect_no_issues subject, <<-CRYSTAL
+        a = 1
+
+        puts a
+
+        a += 1
+        a, b = 2, 3
+        CRYSTAL
+    end
+
+    it "passes if local variable on left side of a comparison" do
+      expect_no_issues subject, <<-CRYSTAL
+        def hello
+          a = 1
+          a || (puts "a is falsey")
+          a
+        end
+        CRYSTAL
+    end
+
+    it "passes if skip_file exists in a macro" do
+      expect_no_issues subject, <<-CRYSTAL
+        {% skip_file %}
+        CRYSTAL
+    end
+
+    it "fails if local variables are unused" do
+      expect_issue subject, <<-CRYSTAL
+        a = 1
+
+        begin
+          a
+        # ^ error: Value from local variable access is unused
+          puts a
+        end
+        CRYSTAL
+    end
+  end
+end
--- a/src/ameba/ast/visitors/implicit_return_visitor.cr
+++ b/src/ameba/ast/visitors/implicit_return_visitor.cr
@@ -91,7 +91,6 @@ module Ameba::AST
     def visit(node : Crystal::MultiAssign) : Bool
       @rule.test(@source, node, @stack.positive?)
 
-      node.targets.each &.accept(self)
       incr_stack { node.values.each &.accept(self) }
 
       false
--- a/src/ameba/ast/visitors/implicit_return_visitor.cr
+++ b/src/ameba/ast/visitors/implicit_return_visitor.cr
@@ -39,7 +39,9 @@ module Ameba::AST
     def visit(node : Crystal::BinaryOp) : Bool
       @rule.test(@source, node, @stack.positive?)
 
-      if node.right.is_a?(Crystal::Call)
+      if node.right.is_a?(Crystal::Call) ||
+         node.right.is_a?(Crystal::Expressions) ||
+         node.right.is_a?(Crystal::ControlExpression)
         incr_stack { node.left.accept(self) }
       else
         node.left.accept(self)
@@ -262,6 +264,14 @@ module Ameba::AST
 
       # Last line of ensure isn't implicitly returned
       swap_stack { node.ensure.try &.accept(self) }
+
+      false
+    end
+
+    def visit(node : Crystal::Block) : Bool
+      @rule.test(@source, node, @stack.positive?)
+
+      node.body.accept(self)
 
       false
     end
--- a/src/ameba/rule/lint/unused_instance_variable_access.cr
+++ b/src/ameba/rule/lint/unused_instance_variable_access.cr
@@ -1,0 +1,60 @@
+module Ameba::Rule::Lint
+  # A rule that disallows unused instance variable access.
+  #
+  # For example, this is considered invalid:
+  #
+  # ```
+  # class MyClass
+  #   @my_var : String = "hello"
+  #
+  #   @my_var
+  #
+  #   def hello : String
+  #     @my_var
+  #
+  #     "hello, world!"
+  #   end
+  # end
+  # ```
+  #
+  # And these are considered valid:
+  #
+  # ```
+  # class MyClass
+  #   @my_var : String = "hello"
+  #
+  #   @my_other_var = @my_var
+  #
+  #   def hello : String
+  #     return @my_var if @my_var == "hello"
+  #
+  #     "hello, world!"
+  #   end
+  # end
+  # ```
+  #
+  # YAML configuration example:
+  #
+  # ```
+  # Lint/UnusedInstanceVariableAccess:
+  #   Enabled: true
+  # ```
+  class UnusedInstanceVariableAccess < Base
+    properties do
+      since_version "1.7.0"
+      description "Disallows unused access to instance variables"
+    end
+
+    MSG = "Value from instance variable access is unused"
+
+    def test(source : Source)
+      AST::ImplicitReturnVisitor.new(self, source)
+    end
+
+    def test(source, node : Crystal::InstanceVar, node_is_used : Bool)
+      return if node_is_used
+
+      issue_for node, MSG
+    end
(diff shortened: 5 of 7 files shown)


spec/ameba/rule/lint/unused_class_variable_access_spec.cr
```diff
@@ -0,0 +1,44 @@
+require "../../../spec_helper"
+
+module Ameba::Rule::Lint
+  subject = UnusedClassVariableAccess.new
+
+  describe UnusedClassVariableAccess do
```
Sija, line 6:
```suggestion
  describe UnusedClassVariableAccess do
    subject = UnusedClassVariableAccess.new

```

spec/ameba/rule/lint/unused_instance_variable_access_spec.cr
```diff
@@ -0,0 +1,44 @@
+require "../../../spec_helper"
+
+module Ameba::Rule::Lint
+  subject = UnusedInstanceVariableAccess.new
+
+  describe UnusedInstanceVariableAccess do
```
Sija, line 6:
```suggestion
  describe UnusedInstanceVariableAccess do
    subject = UnusedInstanceVariableAccess.new

```

spec/ameba/rule/lint/unused_local_variable_access_spec.cr
```diff
@@ -0,0 +1,46 @@
+require "../../../spec_helper"
+
+module Ameba::Rule::Lint
+  subject = UnusedLocalVariableAccess.new
+
+  describe UnusedLocalVariableAccess do
```
Sija, line 6:
```suggestion
  describe UnusedLocalVariableAccess do
    subject = UnusedLocalVariableAccess.new

```

src/ameba/rule/lint/unused_local_variable_access.cr
```diff
@@ -0,0 +1,78 @@
...
+    end
+
+    MSG = "Value from local variable access is unused"
+
+    def test(source : Source)
+      AST::ImplicitReturnVisitor.new(self, source)
+    end
+
+    def test(source, node : Crystal::Var, node_is_used : Bool)
+      return if node_is_used || node.name.in?("skip_file", "self")
```
Sija, line 73: Why `skip_file` variable name has a special treatment?

Sija (changes requested):

src/ameba/rule/lint/unused_local_variable_access.cr
nobodywasishere: For some reason the parser treats it as a var even though it's a call. This should probably be updated to only skip that (and probably `debug` if that's also an issue) when inside macros. I have another branch waiting to improve macro support in the implicit return visitor, I can remove this until then.

https://crystaldoc.info/github/crystal-lang/crystal/1.15.0/Crystal/Macros.html#skip_file:Nop-instance-method

Sija: I'd prefer doing so. Btw, if possible it would be nice to merge the refactors re: macro support first.

spec/ameba/rule/lint/unused_class_variable_access_spec.cr
nobodywasishere: I swear there must be another spec file in the codebase with this issue as I keep just copy / pasting other spec files as templates and changing the class name lol.

Sija: Quite likely, I'm gonna post the PR with fixes which will deal with that once and for all ;)

nobodywasishere:
Just need to update the specs

src/ameba/ast/visitors/implicit_return_visitor.cr
```diff
@@ -280,16 +285,24 @@ module Ameba::AST
       false
     end
 
+    def visit(node : Crystal::Block) : Bool
+      @rule.test(@source, node, @stack.positive?, @in_macro)
```
veelenga, line 289: don't like an idea of expanding the list of args we pass here, because every rule that uses this visitor will have to be updated. I would suggest to following the existed pattern and wrap the params into a struct or something, similar to what we do for scope visitor and flow expression. Examples:

https://github.com/crystal-ameba/ameba/blob/504017bd7588980ac47992cb3631f885718c73bc/src/ameba/rule/lint/useless_assign.cr#L42

https://github.com/crystal-ameba/ameba/blob/504017bd7588980ac47992cb3631f885718c73bc/src/ameba/rule/lint/unreachable_code.cr#L56

Sija:
I'm wondering, what's the point of `node_is_used` argument if all of the usages of `ImplicitReturnVisitor` rely on the fact of it being `true`?

Sija:
@nobodywasishere Perhaps, yet I see no point of having this baked in.

spec/ameba/rule/lint/unused_class_variable_access_spec.cr
```diff
@@ -0,0 +1,46 @@
...
+
+    it "fails if class variables are unused in void context of method" do
+      expect_issue subject, <<-CRYSTAL
+        def hello : String
+          @@name
+        # ^^^^^^ error: Value from class variable access is unused
+
+          "Hello, \#{@@name}!"
+        end
+        CRYSTAL
```
Sija, line 43:
```suggestion
      expect_issue subject, <<-'CRYSTAL'
        def hello : String
          @@name
        # ^^^^^^ error: Value from class variable access is unused

          "Hello, #{@@name}!"
        end
        CRYSTAL
```

ditto elsewhere

src/ameba/rule/lint/unused_instance_variable_access.cr
```diff
@@ -0,0 +1,60 @@
...
+    MSG = "Value from instance variable access is unused"
+
+    def test(source : Source)
+      AST::ImplicitReturnVisitor.new(self, source)
+    end
+
+    def test(source, node : Crystal::InstanceVar, in_macro : Bool)
+      return if in_macro && node.name.in?("@type")
+
+      issue_for node, MSG
```
Sija, line 57: Is this required? `{% @type %}` expression should be flagged too, ain't that right?

```suggestion
      issue_for node, MSG
```

src/ameba/rule/lint/unused_local_variable_access.cr
```diff
@@ -0,0 +1,78 @@
...
+    end
+
+    MSG = "Value from local variable access is unused"
+
+    def test(source : Source)
+      AST::ImplicitReturnVisitor.new(self, source)
+    end
+
+    def test(source, node : Crystal::Var, in_macro : Bool)
+      return if node.name.in?("self") || (in_macro && node.name.in?("skip_file", "debug"))
```
Sija, line 73: Lonely `self` should be flagged too I reckon.

```suggestion
      return if in_macro && node.name.in?("skip_file", "debug")
```

Sija (changes requested):

src/ameba/rule/lint/unused_local_variable_access.cr
nobodywasishere: That will be marked by `Lint/UnusedSelf`

Sija: In such a case I'd make it clear:

```suggestion
      return if node.name == "self" # This case will be reported by `Lint/UnusedSelf` rule
      return if in_macro && node.name.in?("skip_file", "debug")
```

src/ameba/rule/lint/unused_instance_variable_access.cr
Sija: Ah yes, I forgot about it, good catch!
I wonder is that's the case with `@def` as well, or just the `@type`...

nobodywasishere: For now maybe it's worth having the ivar / cvar rules not care about vars in macros altogether

Sija: Idk, it seems that the current implementation holds well enough.
One thing I might change is:

```suggestion
      return if in_macro && node.name == "@type"

      issue_for node, MSG
```

spec/ameba/rule/lint/unused_class_variable_access_spec.cr
Sija: You're on the roll! 🔥 There are some edge-cases though, since there might be cases where you'd have both (`#{foo}` and `\#{foo}`) - or you'd actually want to use the special character (`\n` and alike) - in the same heredoc, so you wouldn't want to use the interpolation-less variant.

.vscode/settings.json
Sija, line 1: This shouldn't be here.

spec/ameba/rule/lint/unused_local_variable_access_spec.cr
```diff
@@ -0,0 +1,69 @@
+require "../../../spec_helper"
+
+module Ameba::Rule::Lint
+  describe UnusedLocalVariableAccess do
+    subject = UnusedLocalVariableAccess.new
+
+    it "passes if local variables are used in assign" do
+      expect_no_issues subject, <<-CRYSTAL
+        a = 1
```
Sija, line 9: Use `foo`, `bar`, ... naming convention to make it clear that errors cover whole name and not just the first character.

ditto elsewhere

Sija (changes requested):

.vscode/settings.json
nobodywasishere: Accident 🙃 been trying to use Zed, so missed it when using git from cli

spec/ameba/rule/lint/unused_class_variable_access_spec.cr
```diff
@@ -0,0 +1,46 @@
...
+
+module Ameba::Rule::Lint
+  describe UnusedClassVariableAccess do
+    subject = UnusedClassVariableAccess.new
+
+    it "passes if class variables are used for assignment" do
+      expect_no_issues subject, <<-CRYSTAL
+        class MyClass
+          foo = @@ivar
+          Bar = @@ivar
```
Sija, line 11:
```suggestion
```

spec/ameba/rule/lint/unused_instance_variable_access_spec.cr
```diff
@@ -0,0 +1,55 @@
...
+
+module Ameba::Rule::Lint
+  describe UnusedInstanceVariableAccess do
+    subject = UnusedInstanceVariableAccess.new
+
+    it "passes if instance variables are used for assignment" do
+      expect_no_issues subject, <<-CRYSTAL
+        class MyClass
+          foo = @ivar
+          Bar = @ivar
```
Sija, line 11:
```suggestion
```

spec/ameba/rule/lint/unused_class_variable_access_spec.cr
Sija, line 1: Specs for macro expressions are missing.

spec/ameba/rule/lint/unused_instance_variable_access_spec.cr
```diff
@@ -0,0 +1,55 @@
...
+        end
+        CRYSTAL
+    end
+
+    it "passes if @type is unused within a macro expression" do
```
Sija, line 16: Failing spec for macro expression is missing.

spec/ameba/rule/lint/unused_local_variable_access_spec.cr
```diff
@@ -0,0 +1,69 @@
...
+        CRYSTAL
+    end
+
+    it "fails if a local variable is in a void context" do
+      expect_issue subject, <<-CRYSTAL
+        foo = 1
+
+        begin
+          foo
+        # ^ error: Value from local variable access is unused
```
Sija, line 50:
```suggestion
        # ^^^ error: Value from local variable access is unused
```

Sija (changes requested):

spec/ameba/rule/lint/unused_class_variable_access_spec.cr
nobodywasishere: Support for them within macros should be removed:
```
icr:1> {% begin %}
icr:2> {{ @@cvar }}
icr:3> {% end %}
error in line 1
Error: expanding macro


error in line 2
Error: can't execute ClassVar in a macro
```

spec/ameba/rule/lint/unused_instance_variable_access_spec.cr
```diff
@@ -0,0 +1,64 @@
...
+
+    it "fails if @def is unused within a macro expression" do
+      expect_issue subject, <<-CRYSTAL
+        def foo
+          {% @def %}
+           # ^^^^ error: Value from instance variable access is unused
+          :bar
+        end
+        CRYSTAL
+    end
```
Sija, line 62:
```suggestion
    it "fails if instance variable is unused within a macro expression" do
      expect_issue subject, <<-CRYSTAL
        def foo
          {% @bar %}
           # ^^^^ error: Value from instance variable access is unused
          :baz
        end
        CRYSTAL
    end
```

also, please group the related test cases together

Sija (changes requested):

spec/ameba/rule/lint/unused_instance_variable_access_spec.cr
Sija: Actually, after a second thought I think that you can't really use instance variables within macros.

nobodywasishere: You can use these ones https://github.com/crystal-lang/crystal/blob/9b7fd8dc19288f6281814a9bae4b81e75aac32f8/src/compiler/crystal/macros/interpreter.cr#L541

src/ameba/ast/visitors/implicit_return_visitor.cr
```diff
@@ -91,7 +94,6 @@ module Ameba::AST
     def visit(node : Crystal::MultiAssign) : Bool
       report_implicit_return(node)
 
-      node.targets.each &.accept(self)
```
veelenga, line 94: could you elaborate on this change? Why the targets are not accepted anymore?

src/ameba/ast/visitors/implicit_return_visitor.cr
```diff
@@ -280,6 +285,14 @@ module Ameba::AST
       false
     end
 
+    def visit(node : Crystal::Block) : Bool
+      report_implicit_return(node)
+
+      node.body.accept(self)
```
veelenga, line 291: what about other parts? args/call?

nobodywasishere: Unless I missed something, only the body can contain nodes that could be unused.

veelenga (approved):
lgtm

Sija (approved):

PR #554, opened 2025-01-29
Add `Style/HeredocIndent`

nobodywasishere:
Enforces that heredoc bodies are indented consistently.

### DIFF
--- a/spec/ameba/rule/documentation/documentation_spec.cr
+++ b/spec/ameba/rule/documentation/documentation_spec.cr
@@ -29,7 +29,7 @@ module Ameba::Rule::Documentation
 
         private macro bag
         end
-      CRYSTAL
+        CRYSTAL
     end
 
     it "passes for documented public types" do
@@ -59,31 +59,31 @@ module Ameba::Rule::Documentation
         # bag
         macro bag
         end
-      CRYSTAL
+        CRYSTAL
     end
 
     it "fails if there is an undocumented public type" do
       expect_issue subject, <<-CRYSTAL
         class Foo
-      # ^^^^^^^^^ error: Missing documentation
+        # ^^^^^^^ error: Missing documentation
         end
 
         module Bar
-      # ^^^^^^^^^^ error: Missing documentation
+        # ^^^^^^^^ error: Missing documentation
         end
 
         enum Baz
-      # ^^^^^^^^ error: Missing documentation
+        # ^^^^^^ error: Missing documentation
         end
 
         def bat
-      # ^^^^^^^ error: Missing documentation
+        # ^^^^^ error: Missing documentation
         end
 
         macro bag
-      # ^^^^^^^^^ error: Missing documentation
+        # ^^^^^^^ error: Missing documentation
         end
-      CRYSTAL
+        CRYSTAL
     end
 
     context "properties" do
--- a/spec/ameba/rule/style/heredoc_indent_spec.cr
+++ b/spec/ameba/rule/style/heredoc_indent_spec.cr
@@ -1,0 +1,81 @@
+require "../../../spec_helper"
+
+module Ameba::Rule::Style
+  describe HeredocIndent do
+    subject = HeredocIndent.new
+
+    it "passes if heredoc body indented one level" do
+      expect_no_issues subject, <<-CRYSTAL
+        <<-HEREDOC
+          hello world
+          HEREDOC
+
+          <<-HEREDOC
+            hello world
+            HEREDOC
+        CRYSTAL
+    end
+
+    it "fails if the heredoc body is indented incorrectly" do
+      expect_issue subject, <<-CRYSTAL
+        <<-ONE
+        # ^^^^ error: Heredoc body should be indented by 2 spaces
+        hello world
+        ONE
+
+          <<-TWO
+        # ^^^^^^ error: Heredoc body should be indented by 2 spaces
+          hello world
+          TWO
+
+          <<-THREE
+        # ^^^^^^^^ error: Heredoc body should be indented by 2 spaces
+             hello world
+             THREE
+
+          <<-FOUR
+        # ^^^^^^^ error: Heredoc body should be indented by 2 spaces
+        hello world
+        FOUR
+        CRYSTAL
+    end
+
+    context "properties" do
+      context "#indent_by" do
+        rule = HeredocIndent.new
+        rule.indent_by = 0
+
+        it "passes if heredoc body has the same indent level" do
+          expect_no_issues rule, <<-CRYSTAL
+            <<-HEREDOC
+            hello world
+            HEREDOC
+
+              <<-HEREDOC
+              hello world
+              HEREDOC
+            CRYSTAL
+        end
+
+        it "fails if the heredoc body is indented incorrectly" do
+          expect_issue rule, <<-CRYSTAL
+            <<-ONE
+            # ^^^^ error: Heredoc body should be indented by 0 spaces
+              hello world
+              ONE
+
+              <<-TWO
+            # ^^^^^^ error: Heredoc body should be indented by 0 spaces
+                hello world
+                TWO
+
+              <<-FOUR
+            # ^^^^^^^ error: Heredoc body should be indented by 0 spaces
+            hello world
+            FOUR
+            CRYSTAL
+        end
+      end
+    end
+  end
+end
--- a/src/ameba/rule/style/heredoc_indent.cr
+++ b/src/ameba/rule/style/heredoc_indent.cr
@@ -1,0 +1,66 @@
+module Ameba::Rule::Style
+  # A rule that enforces _Heredoc_ bodies be indented one level above the indentation of the
+  # line they're used on.
+  #
+  # For example, this is considered invalid:
+  #
+  # ```
+  # <<-HERERDOC
+  #   hello world
+  # HEREDOC
+  #
+  #   <<-HERERDOC
+  # hello world
+  # HEREDOC
+  # ```
+  #
+  # And should be written as:
+  #
+  # ```
+  # <<-HERERDOC
+  #     hello world
+  #   HEREDOC
+  #
+  #   <<-HERERDOC
+  #     hello world
+  #     HEREDOC
+  # ```
+  #
+  # The `IndentBy` configuration option changes the enforced indentation level of the _heredoc_.
+  #
+  # ```
+  # Style/HeredocIndent:
+  #   Enabled: true
+  #   IndentBy: 2
+  # ```
+  class HeredocIndent < Base
+    properties do
+      since_version "1.7.0"
+      description "Recommends heredoc bodies are indented consistently"
+      indent_by 2
+    end
+
+    MSG = "Heredoc body should be indented by %s spaces"
+
+    def test(source, node : Crystal::StringInterpolation)
+      return unless start_location = node.location
+
+      start_location_pos = source.pos(start_location)
+      return unless source.code[start_location_pos..(start_location_pos + 2)]? == "<<-"
+
+      correct_indent = line_indent(source, start_location) + indent_by
+
+      unless node.heredoc_indent == correct_indent
+        issue_for node, MSG % indent_by
+      end
+    end
+
+    private def line_indent(source, start_location) : Int32
+      line_location = Crystal::Location.new(nil, start_location.line_number, 1)
+      line_location_pos = source.pos(line_location)
+      line = source.code[line_location_pos..(line_location_pos + start_location.column_number)]
+
+      line.size - line.lstrip.size
+    end
+  end
+end
--- a/spec/ameba/rule/lint/empty_ensure_spec.cr
+++ b/spec/ameba/rule/lint/empty_ensure_spec.cr
@@ -23,7 +23,7 @@ module Ameba::Rule::Lint
         ensure
           nil
         end
-      CRYSTAL
+        CRYSTAL
     end
 
     it "fails if there is an empty ensure in method" do
@@ -31,9 +31,9 @@ module Ameba::Rule::Lint
         def method
           do_some_stuff
         ensure
-      # ^^^^^^ error: Empty `ensure` block detected
+        # ^^^^ error: Empty `ensure` block detected
         end
-      CRYSTAL
+        CRYSTAL
     end
 
     it "fails if there is an empty ensure in a block" do
@@ -43,10 +43,10 @@ module Ameba::Rule::Lint
         rescue
           do_some_other_stuff
         ensure
-      # ^^^^^^ error: Empty `ensure` block detected
+        # ^^^^ error: Empty `ensure` block detected
           # nothing here
         end
-      CRYSTAL
+        CRYSTAL
     end
   end
 end
--- a/spec/ameba/rule/lint/shadowed_exception_spec.cr
+++ b/spec/ameba/rule/lint/shadowed_exception_spec.cr
@@ -25,7 +25,7 @@ module Ameba::Rule::Lint
         rescue e : Exception
           handle_exception
         end
-      CRYSTAL
+        CRYSTAL
     end
 
     it "fails if there is a shadowed exception" do
@@ -38,7 +38,7 @@ module Ameba::Rule::Lint
              # ^^^^^^^^^^^^^ error: Shadowed exception found: `ArgumentError`
           handle_argument_error_exception
         end
-      CRYSTAL
+        CRYSTAL
     end
 
     it "fails if there is a custom shadowed exceptions" do
@@ -51,7 +51,7 @@ module Ameba::Rule::Lint
              # ^^^^^^^^^^^^^^^^ error: Shadowed exception found: `MySuperException`
           3
         end
-      CRYSTAL
+        CRYSTAL
     end
 
     it "fails if there is a shadowed exception in a type list" do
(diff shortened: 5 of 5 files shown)


spec/ameba/rule/style/heredoc_indent_spec.cr
```diff
@@ -0,0 +1,81 @@
+require "../../../spec_helper"
+
+module Ameba::Rule::Style
+  describe HeredocIndent do
+    subject = HeredocIndent.new
+
+    it "passes if heredoc body indented one level" do
+      expect_no_issues subject, <<-CRYSTAL
+        call <<-HEREDOC
```
Sija, line 9: We're testing just the heredoc strings, no need for additional constructs.

```suggestion
        <<-HEREDOC
```

ditto elsewhere

src/ameba/rule/style/heredoc_indent.cr
```diff
@@ -0,0 +1,69 @@
+module Ameba::Rule::Style
+  # A rule that enforces heredoc bodies be indented one level above the indentation of the
+  # line they're used on.
+  #
+  # For example, this is considered invalid:
+  #
+  # ```
+  # call <<-HERERDOC
```
Sija, line 8: ditto in comments

src/ameba/rule/style/heredoc_indent.cr
```diff
@@ -0,0 +1,69 @@
+module Ameba::Rule::Style
+  # A rule that enforces heredoc bodies be indented one level above the indentation of the
```
Sija, line 2:
```suggestion
  # A rule that enforces _Heredoc_ bodies be indented one level above the indentation of the
```

src/ameba/rule/style/heredoc_indent.cr
```diff
@@ -0,0 +1,69 @@
...
+  #   HEREDOC
+  #
+  # begin
+  #   call <<-HERERDOC
+  #     hello world
+  #     HEREDOC
+  # end
+  # ```
+  #
+  # The `SameLine` configuration option enforces that the heredoc body have the same indent as the
```
Sija, line 33:
```suggestion
  # The `SameLine` configuration option enforces that the _heredoc_ body have the same indent as the
```

src/ameba/rule/style/heredoc_indent.cr
```diff
@@ -0,0 +1,69 @@
...
+  #   SameLine: true
+  # ```
+  class HeredocIndent < Base
+    properties do
+      since_version "1.7.0"
+      description "Recommends heredoc bodies be indented exactly one level above the line they're used on"
+      same_line false
+    end
+
+    MSG = "Heredoc body should be indented by %s space(s)"
```
Sija, line 48:
```suggestion
    MSG = "Heredoc body should be indented by %s spaces"
```

src/ameba/rule/style/heredoc_indent.cr
```diff
@@ -0,0 +1,69 @@
...
+
+    def test(source, node : Crystal::StringInterpolation)
+      return unless start_location = node.location
+      return unless source.code[source.pos(start_location)..(source.pos(start_location) + 2)]? == "<<-"
```
Sija, line 52:
```suggestion
      start_location_pos = source.pos(start_location)
      return unless source.code[start_location_pos..(start_location_pos + 2)]? == "<<-"
```

src/ameba/rule/style/heredoc_indent.cr
```diff
@@ -0,0 +1,69 @@
...
+  # ```
+  # Style/HeredocIndent:
+  #   Enabled: true
+  #   SameLine: true
+  # ```
+  class HeredocIndent < Base
+    properties do
+      since_version "1.7.0"
+      description "Recommends heredoc bodies be indented exactly one level above the line they're used on"
+      same_line false
```
Sija, line 45: I'd instead use an integer value to make the level configurable.

```suggestion
      indent_by 2
```

src/ameba/rule/style/heredoc_indent.cr
```diff
@@ -0,0 +1,69 @@
...
+
+    MSG = "Heredoc body should be indented by %s space(s)"
+
+    def test(source, node : Crystal::StringInterpolation)
+      return unless start_location = node.location
+      return unless source.code[source.pos(start_location)..(source.pos(start_location) + 2)]? == "<<-"
+
+      correct_indent = line_indent(source, start_location) + (same_line? ? 0 : 2)
+
+      if correct_indent != node.heredoc_indent
```
Sija, line 56:
```suggestion
      unless node.heredoc_indent == indent_by
```

Sija (changes requested):

src/ameba/rule/style/heredoc_indent.cr
nobodywasishere: Why bold / capitalized?

spec/ameba/rule/style/heredoc_indent_spec.cr
nobodywasishere: The `call ` is needed to make sure the error `# ^^^^ ` is within the heredoc body indentation-wise

src/ameba/rule/style/heredoc_indent.cr
Sija: That's italic, not bold.

spec/ameba/rule/style/heredoc_indent_spec.cr
Sija: That's handled by the `^{}` marker.

src/ameba/rule/style/heredoc_indent.cr
```diff
@@ -0,0 +1,65 @@
...
+    def test(source, node : Crystal::StringInterpolation)
+      return unless start_location = node.location
+      start_location_pos = source.pos(start_location)
+      return unless source.code[start_location_pos..(start_location_pos + 2)]? == "<<-"
+
+      correct_indent = line_indent(source, start_location) + indent_by
+
+      unless node.heredoc_indent == correct_indent
+        issue_for node, MSG % correct_indent
+      end
```
Sija, line 54: Indentation should measured relatively to the Heredoc string.

```suggestion
      unless node.heredoc_indent == indent_by
        issue_for node, MSG % indent_by
      end
```

src/ameba/rule/style/heredoc_indent.cr
```diff
@@ -0,0 +1,65 @@
...
+      since_version "1.7.0"
+      description "Recommends heredoc bodies are indented consistently"
+      indent_by 2
+    end
+
+    MSG = "Heredoc body should be indented by %s spaces"
+
+    def test(source, node : Crystal::StringInterpolation)
+      return unless start_location = node.location
+      start_location_pos = source.pos(start_location)
```
Sija, line 47:
```suggestion

      start_location_pos = source.pos(start_location)
```

Sija (changes requested):

spec/ameba/rule/style/heredoc_indent_spec.cr
nobodywasishere: What I mean is if the body is indented, the marker can't point to the beginning of the heredoc. This will give a parsing error:

```crystal
<<-HEREDOC
# ^^^^^^^^ error: Heredoc indent ....
  HEREDOC
```

src/ameba/rule/style/heredoc_indent.cr
nobodywasishere: I'm not sure I follow. `heredoc_indent` is the absolute indent of the heredoc ending identifier (not the relative indent), which is used to strip the extra whitespace from the beginning of the heredoc body. We therefore need to compare it against the desired absolute indent. With that, the heredoc body would always have to be indented by 2 from the beginning of the line.

spec/ameba/rule/style/heredoc_indent_spec.cr
nobodywasishere: Also we're testing the indent relative to the indent of w/e is on the line where the heredoc start is, not to the indent of the heredoc start itself

src/ameba/rule/style/heredoc_indent.cr
Sija: Desired indent should always be relative, not absolute.

nobodywasishere: With that code, a heredoc would have to be indented like this to pass:

```crystal
class MyClass
  def my_method
    var = <<-HEREDOC
  This is the value for `var`
  HEREDOC
  end
end
```

Sija: `StringInterpolation#heredoc_indent` is counted from the start of the heredoc line, not start of _the_ line, thus I think you're mistaken.

nobodywasishere:
```crystal
  call <<-HEREDOC
    body of heredoc
    HEREDOC
^^^^
# This is the heredoc_indent
```

With the proposed change, the heredoc body fails unless it's indented with two spaces, regardless of the indentation of the line the heredoc is used on:

image

Sija: That's exactly the point. Line indentation shouldn't matter here.

nobodywasishere: I completely disagree. That way of formatting heredocs does not match what is used in basically all of the codebases I work with, including this one. Having the body indented two lines while the rest of the code could be 6/8/10 lines indented does not look good.

Sija: Perhaps it doesn't, but that has nothing to do with the HEREDOC itself - which this rule ought to be about, and more with the general indentation.

nobodywasishere: I intended this rule to be about the general indentation, nothing to do with the contents of the heredoc. To standardize how much indentation the end tag of the heredoc has in comparison to the indentation of the line the heredoc start tag is on.

To help me better understand, can you provide me some examples of what you think this rule should allow, and what it should disallow?

Sija: Re: heredoc indentation this rule should only act upon the `StringInterpolation#heredoc_indent` property and nothing more. Mixed line indentation should be handled by the Crystal formatter, so this should be already covered by the `Lint/Formatting` rule.

nobodywasishere: > Mixed line indentation should be handled by the Crystal formatter

It is not currently, so maybe this would be better there.

> heredoc indentation this rule should only act upon the StringInterpolation#heredoc_indent property

There's nothing for this rule to do then if you think this should be handled by the formatter itself.

Sija: > It is not currently, so maybe this would be better there.

I'm pretty sure the formatter handles mixed indentation, if some are missed I believe that's a bug in the formatter.

> There's nothing for this rule to do then if you think this should be handled by the formatter itself.

Handling the mixed indentation level for heredoc strings is what I thought this rule is about - and since the beginning I still support this case.

nobodywasishere: I am confused and think we aren't talking about the same things. 

The formatter will move the heredoc start tag (`<<-HEREDOC`), but will not do anything to change the indenting of the body (both the contents and the end tag `HEREDOC`). This rule is intended to have a consistent indentation for the latter, for things like this:

```crystal
call <<-HEREDOC
HEREDOC
```

Which should be this:

```crystal
call <<-HEREDOC
  HEREDOC
```

I got annoyed at changing the indentation of the specs and missing that I needed to update the indentation of the heredocs too, so I thought this rule would be useful, especially since there seem to be two preferences for indentation level.

Are you talking about the indentation of the contents in comparison to the indentation of the final tag? Like:

```crystal
call <<-HEREDOC
contents
HEREDOC
```

vs

```crystal
call <<-HEREDOC
  contents
HEREDOC
```

If so that's not what I originally intended this rule to be, and should probably be called `Style/HeredocContentIndent` instead if you think that rule is worthwhile (I personally don't think it is).

Sija: Same here. In the 2nd example the `heredoc_indent` would be `19` if I'm counting correctly.

nobodywasishere:
```crystal
class MyClass
  def method
    result = db.exec(<<-SQL, arg1, arg2, arg3, as: Int32)
      SELECT *
      FROM table
      WHERE 
        arg1 = $1 AND
        arg2 = $2 AND
        arg3 = $3
      RETURNING id;
      SQL
  end
end
```

In this example, `heredoc_indent` is 6

Sija: I'm almost sure that's 2, not 6. Could you please show a reproduction confirming your number?

nobodywasishere: Printing all the `heredoc_indent`s for each of the heredocs in the tests

image

src/ameba/rule/style/heredoc_indent.cr
```diff
@@ -0,0 +1,66 @@
...
+    def test(source, node : Crystal::StringInterpolation)
+      return unless start_location = node.location
+
+      start_location_pos = source.pos(start_location)
+      return unless source.code[start_location_pos..(start_location_pos + 2)]? == "<<-"
+
+      correct_indent = line_indent(source, start_location) + indent_by
+
+      unless node.heredoc_indent == correct_indent
+        issue_for node, MSG % correct_indent
```
Sija, line 54:
```suggestion
        issue_for node, MSG % indent_by
```

src/ameba/rule/style/heredoc_indent.cr
```diff
@@ -0,0 +1,66 @@
...
+      end
+    end
+
+    def line_indent(source, start_location) : Int32
+      line_location = Crystal::Location.new(filename: nil, line_number: start_location.line_number, column_number: 1)
+
+      line = source.code[source.pos(line_location)..(source.pos(line_location) + start_location.column_number)]
+
+      line.size - line.lstrip.size
+    end
```
Sija, line 64:
```suggestion
    private def line_indent(source, start_location) : Int32
      line_location = Crystal::Location.new(nil, start_location.line_number, 1)
      line_location_pos = source.pos(line_location)
      line = source.code[line_location_pos..(line_location_pos + start_location.column_number)]

      line.size - line.lstrip.size
    end
```

Sija (changes requested):

spec/ameba/rule/style/heredoc_indent_spec.cr
```diff
@@ -0,0 +1,81 @@
...
+            HEREDOC
+        CRYSTAL
+    end
+
+    it "fails if the heredoc body is indented incorrectly" do
+      expect_issue subject, <<-CRYSTAL
+        call <<-ONE
+           # ^^^^^^ error: Heredoc body should be indented by 2 spaces
+        hello world
+        ONE
```
Sija, line 24:
```suggestion
        <<-ONE
        # ^^^^ error: Heredoc body should be indented by 2 spaces
        hello world
        ONE
```

ditto elsewhere

nobodywasishere: That only works if the heredoc isn't indented at all. I'd have to write this, which fails to parse:

```crystal
  <<-HEREDOC
# ^^^^^^^^^^ error: ...
   HEREDOC
```

Sija: Following wouldn't work?

```cr
<<-ONE
# ^^^^ error: Heredoc body should be indented by 2 spaces
hello world
ONE
```

nobodywasishere: Forgot that those annotations are removed before parsing, disregard.

Sija (approved):

spec/ameba/rule/documentation/documentation_spec.cr
```diff
@@ -59,31 +59,31 @@ module Ameba::Rule::Documentation
...
         end
-      CRYSTAL
+        CRYSTAL
     end
 
     it "fails if there is an undocumented public type" do
       expect_issue subject, <<-CRYSTAL
         class Foo
-      # ^^^^^^^^^ error: Missing documentation
+        # ^^^^^^^ error: Missing documentation
```
veelenga, line 68: does the `^^^...` error marker style work correctly for both cases? I was pretty sure there is a single way to write it.

veelenga (approved):
lgtm

spec/ameba/rule/documentation/documentation_spec.cr
nobodywasishere: There is a single way to write it, these needed to be changed as they've been de-indented

veelenga: Why did the tests pass before if it was de-intended?

nobodywasishere: The test worked before as it was indented 2 spaces, now it's indented 0

veelenga: Exactly, since the logic on error marker hasn't been changed, I assume it works with both cases, when it is indented 2 spaces or 0. Is that correct?

nobodywasishere: I'm not sure what you're saying. I needed to change the error hint as I moved the heredoc end tag. IIRC the error annotations have to be an exact match, so there's only one error annotation form that can exist for a line.

veelenga: Sorry, i missed that the heredoc end tag was moved for this case.

Sija (approved):

PR #555, opened 2025-01-29
Refactors before 1.7 release

Sija:


nobodywasishere (approved):

spec/ameba/rule/documentation/documentation_spec.cr
```diff
@@ -1,14 +1,14 @@
...
 module Ameba::Rule::Documentation
-  subject = Documentation.new
-    .tap(&.ignore_classes = false)
-    .tap(&.ignore_modules = false)
-    .tap(&.ignore_enums = false)
-    .tap(&.ignore_defs = false)
-    .tap(&.ignore_macros = false)
-
   describe Documentation do
+    subject = Documentation.new
```
veelenga, line 5: only now noticed that the rule name is kinda off:

`Ameba::Rule::Documentation::Documentation`

veelenga (commented):
Honestly, I always feel uncomfortable with these massive refactorings across the codebase. In most big projects, contributions that involve refactoring are often discouraged because many suggestions tend to be subjective and, in some cases, introduce hidden issues.

I’m okay with this one, however, wanna add that refactoring just for the sake of cleaner code can sometimes create unnecessary churn and increasing the risk of introducing subtle bugs, especially before the release.

veelenga (approved):

Sija:
@veelenga I understand your sentiment, and in general I'm against such big PRs, yet once in a while I tend to gather all smaller (and thus with no or very slight chances of regressions) changes into one batch to save us time and energy that would go into reviewing these independently.

PR #560, opened 2025-02-03
`Lint/UnusedLiteral` marks unused nil literals as unused

nobodywasishere:
Missed this for some reason


nobodywasishere:
The unused nil _is_ actually unused, but I'm not sure if there's a better way of writing that to work around it, besides expanding it to a full rescue block.

Sija:
That's a customary way to write calls ignoring the errors with trailing rescue (there's no other way to do so otherwise), so I reckon these should be ignored.

nobodywasishere:
Closing this for now. Unused `nil`s don't really have an impact (unlike unused arrays or other literals), so this isn't as big of a deal.

PR #567, opened 2025-02-15
Add `Lint/SpecEqWithBoolOrNilLiteral` rule

Sija:


nobodywasishere (approved):

spec/ameba/rule/lint/spec_eq_with_bool_or_nil_literal_spec.cr
```diff
@@ -0,0 +1,45 @@
...
+        foo.is_a?(String).should be_true
+        foo.is_a?(Int32).should be_false
+        CRYSTAL
+    end
+
+    it "does not report if `be_nil` expectation is used" do
+      expect_no_issues subject, <<-CRYSTAL, path: "source_spec.cr"
+        foo.as?(Symbol).should be_nil
+        CRYSTAL
+    end
```
veelenga, line 18: i would add a case when `should` word is not used, but `be true/false` is

Sija: I can't imagine such a situation tbh, since expectations ought to be used exclusively with `should`.

veelenga: Easy: a custom method could be defined `be true/false`. Anyway, we explicitly check for the `should` call which has to be covered by the spec IMO:

https://github.com/crystal-ameba/ameba/blob/5f99d07195e3f525a828e05c120679c8431a3ab5/src/ameba/rule/lint/spec_eq_with_bool_or_nil_literal.cr#L43

veelenga (approved):
lgtm

spec/ameba/rule/lint/spec_eq_with_bool_or_nil_literal_spec.cr
Sija: Good point, will add some moar test-cases covering that.

PR #568, opened 2025-02-16
Add `Lint/DuplicateWhenCondition` rule

Sija:


src/ameba/rule/lint/duplicated_case_condition.cr
```diff
@@ -0,0 +1,51 @@
...
+  # end
+  # ```
+  #
+  # YAML configuration example:
+  #
+  # ```
+  # Lint/DuplicatedCaseCondition:
+  #   Enabled: true
+  # ```
+  class DuplicatedCaseCondition < Base
```
nobodywasishere, line 32: I think this should be called `DuplicateWhenCondition`

spec/ameba/rule/lint/duplicated_case_condition_spec.cr
```diff
@@ -0,0 +1,32 @@
...
+        when Symbol
+          do_something_else
+        end
+        CRYSTAL
+    end
+
+    it "reports if there are a duplicated `when` conditions" do
+      source = expect_issue subject, <<-CRYSTAL
+        case x
+        when .nil?
```
nobodywasishere, line 21: Does this work for this case:

```crystal
case x
when .foo?, .bar?
  do_something
when .bar?
  do_something_else
end
```

Sija: Not at the moment.

Sija: Good catch! I've added support for checking independent conditions.

veelenga (approved):
lgtm

src/ameba/rule/lint/duplicate_when_condition.cr
```diff
@@ -0,0 +1,53 @@
...
+  # ```
+  class DuplicateWhenCondition < Base
+    properties do
+      since_version "1.7.0"
+      description "Reports repeated conditions used in case `when` expressions"
+    end
+
+    MSG = "Duplicate `when` condition detected"
+
+    def test(source, node : Crystal::Case)
```
nobodywasishere, line 40: Should this also support `Crystal::Select`?

spec/ameba/rule/lint/duplicate_when_condition_spec.cr
```diff
@@ -0,0 +1,45 @@
+require "../../../spec_helper"
+
+module Ameba::Rule::Lint
+  describe DuplicateWhenCondition do
+    subject = DuplicateWhenCondition.new
+
+    it "passes if there are no duplicated `when` conditions" do
```
nobodywasishere, line 7: nitpick: Could be nice to have a test with `in`

Sija: That's not necessary since `in` branches can't be duplicates in exhaustive `case` variant.

PR #576, opened 2025-02-18
Fix issue sorting

Sija:
Fixup of #528

### DIFF
--- a/src/ameba/runner.cr
+++ b/src/ameba/runner.cr
@@ -162,7 +162,12 @@ module Ameba
     ensure
       missing_location = Crystal::Location.new(nil, 0, 0)
       source.issues.sort_by! do |issue|
-        issue.location || missing_location
+        location = issue.location || missing_location
+        {
+          location.filename.to_s,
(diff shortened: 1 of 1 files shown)


src/ameba/runner.cr
```diff
@@ -160,9 +160,9 @@ module Ameba
 
       File.write(source.path, source.code) unless corrected_issues.empty?
     ensure
-      missing_location = Crystal::Location.new(nil, 0, 0)
+      missing_location = Crystal::Location.new(nil, 0, 0).to_s
       source.issues.sort_by! do |issue|
-        issue.location || missing_location
+        issue.location.to_s || missing_location
```
nobodywasishere, line 165: if `issue.location` is `nil`, won't `.to_s` convert it into an empty string and `missing_location` is never used?

Sija: yup, did fix it.

nobodywasishere (approved):

veelenga (approved):

PR #570, opened 2025-02-16
Add `Style/PercentLiteralDelimiters` rule

Sija:

PR #578, opened 2025-02-19
Improve `Style/PercentLiteralDelimiters`

veelenga:
Adding some little improvements to make it more readable and reduce the cyclomatic complexity. 

Refs #570

### DIFF
--- a/src/ameba/rule/style/percent_literal_delimiters.cr
+++ b/src/ameba/rule/style/percent_literal_delimiters.cr
@@ -1,13 +1,10 @@
 module Ameba::Rule::Style
   # A rule that enforces the consistent usage of `%`-literal delimiters.
   #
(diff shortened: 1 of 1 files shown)


Sija (commented):
TBH I find this less readable than the original version. I can imagine some form of more generalized refactor taking into account this rule and `Lint/PercentArrays`, which uses similar logic, yet that's not the case here.

veelenga:
@Sija 

Sure! I’d suggest proceeding with the refactoring you mentioned first, as the original version is quite hard to follow and support, to be honest.

Sija:
@veelenga Could you explain what's hard to follow in the original version? I'm finding it as pretty straight forward procedural code (yet of course I'm biased since it's my work in the end)

veelenga:
I will add a feedback to the PR later on.

PR #582, opened 2025-02-24
Add `Lint/DuplicateBranch` rule

Sija:


src/ameba/severity.cr
```diff
@@ -27,7 +27,7 @@ module Ameba
     def color : Colorize::Color
       case self
       in Error      then Colorize::ColorANSI::Red
-      in Warning    then Colorize::ColorANSI::Red
+      in Warning    then Colorize::ColorANSI::Red # ameba:disable Lint/DuplicateBranch
```
veelenga, line 30: can't this be reworked in a way when the disable directive is not needed?

veelenga (commented):
Review still WIP

src/ameba/severity.cr
straight-shoota: Maybe this indicates that duplicate paths in a case body should be accepted?

Sija: @veelenga Easiest way to do it is changing the color value.

straight-shoota: Then maybe that option should be enabled by default?

Sija: @straight-shoota For what reason?

Sija: @straight-shoota I'm open to see good use cases of this behavior. This case is IMO a great example where this rule uncovers dubious choices, since this branch shouldn't be repeated in the first place.

Sija: This btw uncovers the fact that we only have a single rule - `Lint/Syntax` - with an `Error` severity.

src/ameba/ast/variabling/variable.cr
```diff
@@ -137,10 +137,10 @@ module Ameba::AST
...
     def target_of?(assign)
       case assign
-      when Crystal::Assign           then eql?(assign.target)
-      when Crystal::OpAssign         then eql?(assign.target)
-      when Crystal::MultiAssign      then assign.targets.any? { |target| eql?(target) }
-      when Crystal::UninitializedVar then eql?(assign.var)
+      when Crystal::UninitializedVar          then eql?(assign.var)
+      when Crystal::Assign, Crystal::OpAssign then eql?(assign.target)
+      when Crystal::MultiAssign
+        assign.targets.any? { |target| eql?(target) }
```
straight-shoota, line 143: I'm not sure if this change (and similar above) is a good choice.

This aspect has not yet been mentioned in the discussion or anywhere in the documentation and specs: some conditions have an effect on flow typing and merging branches affects typing.

Here the different branches have technically the same body. But combining them introduce a union type `Crystal::Assign | Crystal::OpAssign` which can lead to all kinds of potentially unintended consequences.

If we keep the branches separate, the type of `assign` is restricted to the respective type.

The compiler uses this pattern intentionally in a number of places (including ones dealing with switching over AST nodes). So I would assume it's also intentional here.

src/ameba/severity.cr
straight-shoota: As mentioned before, especially `case` with `then` branches is often used for mappings and when the input set of values is bigger than the output set, there'll be multiple inputs pointing to the same output. And that's totally fine within the domain model.
However, I don't think there's a clear preference about how to express that in code. Sometimes it might be best to group input values which lead to the same value. Sometimes it's better to keep the input values separate and in their implicit order, for example.
IMO ameba shouldn't take preference here because this is a style matter specific to each individual use case.

Sure, having more strict rules may uncover dubious choices like the example here.
But there's also a high chance of too many false positives.
I think the default behaviour should be cautious to avoid noise. Nobody want's to deal with long reports of false violations.

nobodywasishere: I agree. Too many rule violations highlighting too much of the document, especially for things that are IMO more stylistic choices, will turn people away from using ameba and it's editor integrations. If this rule should be added I think it should be disabled by default.

src/ameba/ast/variabling/variable.cr
Sija: I don't see how this affects dispatch, see here: https://carc.in/#/r/hqb4

src/ameba/severity.cr
Sija: That's true, I've disabled by default as suggested 👍

PR #583, opened 2025-02-26
Add `Lint/WhitespaceAroundMacroExpression` rule

Sija:
Depends on https://github.com/crystal-lang/crystal/pull/15524

### DIFF
--- a/spec/ameba/rule/lint/whitespace_around_macro_expression_spec.cr
+++ b/spec/ameba/rule/lint/whitespace_around_macro_expression_spec.cr
@@ -1,0 +1,52 @@
+require "../../../spec_helper"
+
+module Ameba::Rule::Lint
+  describe WhitespaceAroundMacroExpression do
+    subject = WhitespaceAroundMacroExpression.new
+
+    it "passes if macro expression is wrapped with whitespace" do
+      expect_no_issues subject, <<-CRYSTAL
+        {{ foo }}
+        CRYSTAL
+    end
+
+    it "reports macro expression without whitespace around" do
+      source = expect_issue subject, <<-CRYSTAL
+        {{foo}}
+        # ^^^^^ error: Missing spaces around macro expression
+        {{ bar}}
+        # ^^^^^^ error: Missing spaces around macro expression
+        {{baz }}
+        # ^^^^^^ error: Missing spaces around macro expression
+        CRYSTAL
+
+      expect_correction source, <<-CRYSTAL
+        {{ foo }}
+        {{ bar }}
+        {{ baz }}
+        CRYSTAL
+    end
+
+    # https://github.com/crystal-lang/crystal/pull/15524
+    it "reports macro expression without whitespace around within a macro body" do
+      source = expect_issue subject, <<-CRYSTAL
+        {% begin %}
+          {{foo}}
+        # ^^^^^^^ error: Missing spaces around macro expression
+          {{ bar}}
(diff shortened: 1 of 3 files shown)


spec/ameba/rule/lint/whitespace_around_macro_expression_spec.cr
```diff
@@ -0,0 +1,52 @@
+require "../../../spec_helper"
+
+module Ameba::Rule::Lint
+  describe WhitespaceAroundMacroExpression do
+    subject = WhitespaceAroundMacroExpression.new
+
+    it "passes if macro expression is wrapped with whitespace" do
```
nobodywasishere, line 7: `{% %}` that aren't macro if / begin / verbatim / etc are also `MacroExpression`, there should be specs for those. Stuff like:

```crystal
{% puts "foo" %}
{%puts "foo" %}
{% puts "foo"%}
{%puts "foo"%}
```

Sija: These are not being touched.

Sija: Also, `{% ... %}` expressions are being formatted by the Crystal formatter anyway.

nobodywasishere:
What is the timeline for 1.7.0? If we need to wait till 1.16.0 that won't be till April, and there are a lot of improvements made so far a part of that release that I want to make use of sooner rather than later. Not rushing anything, just want to know some sort of expected timeline.

src/ameba/rule/lint/whitespace_around_macro_expression.cr
```diff
@@ -0,0 +1,45 @@
...
+  # Lint/WhitespaceAroundMacroExpression:
+  #   Enabled: true
+  # ```
+  class WhitespaceAroundMacroExpression < Base
+    include AST::Util
+
+    properties do
+      since_version "1.7.0"
+      description "Reports missing spaces around macro expressions"
+    end
```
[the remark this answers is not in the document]
Sija: Mind reader you are, I was just thinking about that! 😅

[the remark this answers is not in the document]
Sija: I'm gonna take the easy path for now...

src/ameba/rule/lint/whitespace_around_macro_expression.cr
```diff
@@ -0,0 +1,50 @@
...
+        return
+      {% end %}
+
+      return unless node.output?
+      return unless code = node_source(node, source.lines)
+      return if code.starts_with?("{{ ") && code.ends_with?(" }}")
+
+      issue_for node, MSG do |corrector|
+        corrected_code =
+          "{{ #{code.lstrip('{').rstrip('}').strip(' ')} }}"
```
nobodywasishere: (minor) from a performance perspective, this is creating a lot of intermediate strings that are immediately thrown away. May be better to use a string builder and iterate over the chars of the code string. May be worth combing thru the codebase for any other low hanging fruit like this.

nobodywasishere: Should also only remove the first and last two brackets, so it could be something like this:

```
corrected_code = String.build do |bld|
  bld << "{{ "

  code.each_char_with_index do |char, idx|
    # Skip brackets
    next if idx < 2 || idx > (code.size - 3)

    # rest of handling char
  end

  bld << " }}"
end
```

PR #586, opened 2025-03-03
Add `unless` to nodes counted by `AST::CountingVisitor`

Sija:
This PR also raises default `MaxComplexity` setting of `Metrics/CyclomaticComplexity` rule to _12_, since otherwise a lot of code is gonna trigger that warning. Aside from this fact, I believe it's more sensible value anyway.


veelenga (approved):
lgtm 👏🏻

PR #591, opened 2025-03-05
Add `Lint/SignalTrap` rule

Sija:


src/ameba/rule/lint/signal_trap.cr
```diff
@@ -0,0 +1,52 @@
...
+    end
+
+    MSG = "Use `Process.on_terminate` instead of `%s.trap`"
+
+    def test(source, node : Crystal::Call)
+      return unless (obj = node.obj).is_a?(Crystal::Path)
+      return unless path_named?(obj, "Signal::INT", "Signal::HUP", "Signal::TERM")
+      return unless node.name == "trap"
+
+      if (location = node.location) && (end_location = name_end_location(node))
```
nobodywasishere, line 43: nitpick: _technically_ you could have a comment / newline between `Signal::INT` and `trap`, which this would remove. It may be better to replace the `Signal::INT` with `Process` and `trap` with `on_terminate` separately

```
Signal::INT
  # foo bar 
  .trap do |_s|
  handler.call
end
```

https://carc.in/#/r/hr4w

Sija: Quite unlikely but possible, nice catch!

veelenga (approved):
lgtm

PR #524, opened 2024-12-03
Add `Style/MultilineCurlyBlock`

nobodywasishere:
Closes #523

This could have a corrector, but the different associativity between curly and `do`...`end` blocks could lead to semantic differences.

Issue #585, opened 2025-02-27
Add testcases for ImplicitReturnVisitor


nobodywasishere:

nobodywasishere:
Not sure how to best approach this without writing tests that are too specific to the implementation. Open to ideas if others have some.

PR #595, opened 2025-03-14
Support autocorrection in `Style/MultilineCurlyBlock`

Sija:
Refs #524

### DIFF
--- a/spec/ameba/rule/style/multiline_curly_block_spec.cr
+++ b/spec/ameba/rule/style/multiline_curly_block_spec.cr
@@ -25,11 +25,17 @@ module Ameba::Rule::Style
     end
 
     it "reports if there is a multi-line curly block" do
-      expect_issue subject, <<-CRYSTAL
+      source = expect_issue subject, <<-CRYSTAL
         foo {
           # ^ error: Use `do`...`end` instead of curly brackets for multi-line blocks
           :bar
         }
+        CRYSTAL
+
+      expect_correction source, <<-CRYSTAL
+        foo do
+          :bar
+        end
         CRYSTAL
     end
   end
--- a/src/ameba/rule/style/multiline_curly_block.cr
+++ b/src/ameba/rule/style/multiline_curly_block.cr
@@ -40,7 +40,10 @@ module Ameba::Rule::Style
       return if start_location.line_number == end_location.line_number
       return unless source.code[source.pos(start_location)]? == '{'
 
-      issue_for node, MSG
+      issue_for node, MSG do |corrector|
+        corrector.replace(start_location, start_location, "do")
+        corrector.replace(end_location, end_location, "end")
+      end
     end
   end
 end


nobodywasishere:
Technically this can change the semantics as `{}` and `do ... end` attach to blocks in calls differently, but this is a minor edge case

Sija:
Agreed, yet as you've already mentioned, I regard it too as a minor edge-case, rarely happening in practice.

veelenga (approved):
lgtm

straight-shoota:
I'd be cautious about disregarding negative effects because they only affect minor edge cases. The point of autocorrection is that it fully automatically improves code quality. A critical prerequisite for that is not to accidentally _break_ it in the process.

I'd strictly prefer having no autocorrect for a stylistic issue over the chance of breaking the code.

straight-shoota:
It really shouldn't be hard to detect when turning curly braces into do/end could introduce a semantic change. We just need to check whether the current call is part of a call chain.
So I believe ameba could handle this better. Either by refusing autocorrect on call chains, or going one step further and adding the appropriate parenthesis to maintain the intended semantics.

Sija:
> It really shouldn't be hard to detect when turning curly braces into do/end could introduce a semantic change. We just need to check whether the current call is part of a call chain.

_Shouldn't be hard_ and _just_ are not the words I'd use in terms of implementation.

From the perspective of the singular call there's no access to the call chain as-is. Also, checking for associativity would have to account for the following calls making this even harder to do.

If you see an easy way of doing it, please do post it. I'd gladly see some tangible solutions to fix this edge-case. As for now, I'm not gonna invest hours for something that rarely happens. As an alternate option I can revert this change altogether.

nobodywasishere:
Rubocop makes a distinction between normal and breaking fixes, and may be worth looking into whether we could make that distinction too for situations like this

Sija:
@straight-shoota Running `ameba --only Style/MultilineCurlyBlock --fix` against Crystal codebase didn't break anything - i.e. `make crystal spec` is 🟢 . If you have any concrete proof/examples where this edge-case might be an issue, please do not hesitate and post it here so we can evaluate its significance.

For now I see 2 immediate courses of action:

1. Keep this PR as-is, lowering number of manual corrections users will have to do, in exchange for potential code breakage / bug-reports l8r on
2. Revert this PR and increase the amount of work (i.e. punish) the majority of users on account of these edge-cases

Any thoughts?

/cc @veelenga

nobodywasishere:
> yet ATM I have no capacity to tackle it

I potentially do if you have the capacity to review it.

> If you have any concrete proof/examples where this edge-case might be an issue

https://carc.in/#/r/hs9l

Sija:
> I potentially do if you have the capacity to review it.

In some time, quite possibly, although I'd very much prefer to close already opened issues (like #585) before opening new ones, especially feature-related.

nobodywasishere:
I'll focus on that first then

Sija:
I've decided to revert this change as the most immediate and sane course of action.

PR #599, opened 2025-03-19
Misc refactors

Sija:


nobodywasishere (approved):

Sija:
@veelenga Could you please review it?

straight-shoota (commented):
That's a lot of changes, and they have varying decrees of impact. 
I'm wondering if it might be a good idea to extract some of the non-trivial changes into separate PRs to have them stand out more?

Sija:
@straight-shoota This might seem like a lot but there's not _that_ much changes, actually. If you think that some particular one deserves their own PR, please just say so.

spec/ameba/ast/visitors/node_visitor_spec.cr
```diff
@@ -1,13 +1,12 @@
...
 module Ameba::AST
-  rule = DummyRule.new
-  source = Source.new ""
-
   describe NodeVisitor do
     describe "visit" do
       it "allow to visit ASTNode" do
-        visitor = NodeVisitor.new rule, source
+        rule = DummyRule.new
+        visitor = NodeVisitor.new rule, Source.new ""
```
veelenga, line 8:
```suggestion
        visitor = NodeVisitor.new rule, Source.new("")
```

spec/ameba/rule/documentation/documentation_admonition_spec.cr
```diff
@@ -53,7 +59,7 @@ module Ameba::Rule::Documentation
           past_date = (Time.utc - 21.days).to_s(format: "%F")
           expect_issue subject, <<-CRYSTAL, admonition: admonition
             # %{admonition}(#{past_date}): sth in the past
-            # ^{admonition} error: Found a #{admonition} admonition in a comment (21 days past)
+            # ^{admonition} error: Found a %{admonition} admonition in a comment (21 days past)
```
veelenga, line 62: could you please clarify what the difference in this syntax and we change it?

spec/ameba/rule/style/parentheses_around_condition_spec.cr
```diff
@@ -73,6 +64,15 @@ module Ameba::Rule::Style
             (foo rescue 42) ? true : false
             CRYSTAL
         end
+
+        it "allows to skips ternary control expressions" do
```
veelenga, line 68:
```suggestion
        it "allows to skip ternary control expressions" do
```

src/ameba/ast/visitors/implicit_return_visitor.cr
```diff
@@ -56,13 +54,7 @@ module Ameba::AST
...
       report_implicit_return(node)
 
-      incr_stack do
-        node.obj.try &.accept(self)
-        node.args.each &.accept(self)
-        node.named_args.try &.each &.accept(self)
-        node.block_arg.try &.accept(self)
-        node.block.try &.accept(self)
-      end
+      incr_stack { node.accept_children(self) }
```
veelenga, line 57: :+1:

src/ameba/ast/visitors/implicit_return_visitor.cr
```diff
@@ -168,10 +160,7 @@ module Ameba::AST
     def visit(node : Crystal::FunDef) : Bool
       report_implicit_return(node)
 
-      incr_stack do
-        node.args.each &.accept(self)
-        node.body.try &.accept(self)
-      end
+      incr_stack { node.accept_children(self) }
```
veelenga, line 163: now it traverses `node.return_type` which was not the case before. Is this expected? 

see https://github.com/crystal-lang/crystal/blob/master/src/compiler/crystal/syntax/ast.cr#L1968C8-L1968C19

src/ameba/rule/base.cr
```diff
@@ -115,7 +115,7 @@ module Ameba::Rule
 
     # Adds an issue to the *source*
     macro issue_for(*args, **kwargs, &block)
-      source.add_issue(self, {{ args.splat }}, {{ kwargs.double_splat }}) {{ block }}
+      source.add_issue(self, {{ args.splat(", ") }}{{ kwargs.double_splat }}) {{ block }}
```
veelenga, line 118: can you elaborate here please? Does this change the behaviour? What's the reason of this change?

veelenga (commented):
Reviewed partially.

spec/ameba/rule/documentation/documentation_admonition_spec.cr
Sija: There's none. I've changed it for consistency reasons, as I've already used the `%{}` variant.

src/ameba/ast/visitors/implicit_return_visitor.cr
Sija: I thought it's fine as IIUC it doesn't change much.

src/ameba/rule/base.cr
Sija: Without this change this macro produces invalid code if called with only named arguments.

Sija:
CI nightly jobs started failing with error:

```
Error: expected argument #2 to 'Ameba::Source#add_issue' to be Crystal::Token or Tuple(Int32, Int32), not (Crystal::ASTNode | Nil)
```

As far as I can see it's related to: https://github.com/crystal-ameba/ameba/blob/26db705effc0e2c198f69d6bf2ed8fe69aef1e56/src/ameba/ast/variabling/assignment.cr#L68

Fixed in https://github.com/crystal-lang/crystal/pull/15589

src/ameba/ast/visitors/implicit_return_visitor.cr
nobodywasishere: It won't cause any issues since it's being marked as used by the visitor

src/ameba/rule/documentation/documentation_admonition.cr
```diff
@@ -62,9 +62,11 @@ module Ameba::Rule::Documentation
         matches.each do |match|
           admonition = match["admonition"]
 
+          begin_location =
+            token.location.adjust(column_number: 2) # adjust for "# "
           end_location =
-            token.location.adjust(column_number: admonition.size + 1)
-          token_location = {token.location, end_location}
+            begin_location.adjust(column_number: admonition.size - 1)
+          token_location = {begin_location, end_location}
```
veelenga, line 69: does this change the behaviour? The column number is now shifted by 2, while before it was not.

src/ameba/rule/lint/unused_self_access.cr
```diff
@@ -32,10 +32,6 @@ module Ameba::Rule::Lint
       AST::ImplicitReturnVisitor.new(self, source)
     end
 
-    def test(source, node : Crystal::Self, in_macro : Bool)
-      issue_for node, MSG
-    end
```
veelenga, line 37: why do we remove this one? This not being traversed by `AST::ImplicitReturnVisitor` ?

src/ameba/ast/visitors/implicit_return_visitor.cr
```diff
@@ -324,20 +304,11 @@ module Ameba::AST
...
 
-    def visit(
-      node : Crystal::BoolLiteral | Crystal::CharLiteral | Crystal::NumberLiteral |
-             Crystal::StringLiteral | Crystal::SymbolLiteral | Crystal::ProcLiteral,
-    ) : Bool
-      report_implicit_return(node)
-
-      true
-    end
-
```
veelenga, line 340: Hard to follow why this was removed as well.

veelenga:
I would appreciate if we could extract changes in `VerboseBlock` into separate PR. That one looks like a tough one.

src/ameba/rule/documentation/documentation_admonition.cr
Sija: Yes, it does - intentionally.

Before:

```
# TODO: xxx
^^^^^^
```

After:

```
# TODO: xxx
  ^^^^
```

src/ameba/rule/lint/unused_self_access.cr
Sija: Because it's pointless. Among others, that's the reason we have specs - to test for regressions, and as you can see there are none failing...

src/ameba/ast/visitors/implicit_return_visitor.cr
Sija: It was replaced by definition with an union instead of duplicating it for each argument type.

Sija:
> I would appreciate if we could extract changes in `VerboseBlock` into separate PR. That one looks like a tough one.

There's no point to do so, I've just shuffled the order of method definitions, nothing really changes beside that.

Sija:
@veelenga I'd advise you to look at the individual commits, as opposed to trying to grok the differences from the overall view - this might answer some of your questions.

src/ameba/rule/lint/unused_self_access.cr
nobodywasishere: I agree, this should be fine

veelenga:
> @veelenga I'd advise you to look at the individual commits, as opposed to trying to grok the differences from the overall view - this might answer some of your questions.

Thanks for the heads up, but there are 37 commits, I don't think it is easier to review the individual commits one by one.

veelenga:
> > I would appreciate if we could extract changes in `VerboseBlock` into separate PR. That one looks like a tough one.
> 
> There's no point to do so, I've just shuffled the order of method definitions (so the private methods will come after the public ones), nothing really changed beside that.

Could you please clarify the reason for this change? It seems subjective and isn't enforced by any established rules, meaning it can't be guaranteed in the future. Additionally, someone else could join the project with a different perspective and revert the changes back or enforce other rules they think look better. Do you agree?

src/ameba/ast/visitors/implicit_return_visitor.cr
nobodywasishere: They're covered by the untyped visit method now https://github.com/crystal-ameba/ameba/pull/599/files#diff-8eaca1fca4746288c3e629e6af8c7e569d6b72cfad5cf423566e14949bbef348R368

veelenga: Ok, thanks

Sija:
> Could you please clarify the reason for this change? It seems subjective and isn't enforced by any established rules, meaning it can't be guaranteed in the future. Additionally, someone else could join the project with a different perspective and revert the changes back or enforce other rules they think look better. Do you agree?

Sure. That's the rule I wrote long time ago and now revisiting it after time, I'm finding it's odd having to dig through all of the private methods in order to get to the public ones. Most of other rules are written in such fashion, and I believe it's for a reason (i.e. readability). I'd happily accept a rule requiring an ordering of `public/private/protected` methods, yet ATM there's none. Regarding your last question: no, I don't agree since, as I've already mentioned - most of the rules are written in that fashion, and this single one was kinda an outlier.

veelenga:
> > Could you please clarify the reason for this change? It seems subjective and isn't enforced by any established rules, meaning it can't be guaranteed in the future. Additionally, someone else could join the project with a different perspective and revert the changes back or enforce other rules they think look better. Do you agree?
> 
> Sure. That's the rule I wrote long time ago and now revisiting it after time, I'm finding it's odd having to dig through all of the private methods in order to get to the public ones. Most of other rules are written in such fashion, and I believe it's for a reason (i.e. readability). I'd happily accept a rule requiring an ordering of `public/private/protected` methods, yet ATM there's none. Regarding your last question: no, I don't agree since, as I've already mentioned - most of the rules are written in that fashion, and this single one was kinda an outlier.

Got it, thanks for explaining. However, I personally don't find this version more readable than the previous one. The change doesn't seem to offer any significant improvement from my perspective. 

Also, I want to re-call what I mentioned few times during previous Misc refactoring:

1. No major benefit
2. Hard to review, takes a lot of time
3. Easy to introduce a regression

This PR contains some valuable changes and improvements but mixing it with code style suggestions, while these styles are not enforced on the project, is counterproductive IMO. Code style changes should be discussed separately, maybe written somewhere, and ideally automated with linting rules before being implemented. Otherwise, we risk spending valuable review time on subjective matters rather than focusing on functional improvements.

veelenga:
@Sija don't want to block this one. Please feel free to finish / merge it if you think it is necessary.

PR #601, opened 2025-03-24
Add `Lint/ElseNil` rule

Sija:


spec/ameba/rule/lint/else_nil_spec.cr
```diff
@@ -0,0 +1,100 @@
...
+        CRYSTAL
+    end
+
+    it "reports if there is an `else` block of a `rescue` with a `nil` body" do
+      source = expect_issue subject, <<-CRYSTAL
+        begin
+          do_something
+        rescue
+          rescue_error
+        else
```
nobodywasishere, line 85: I wouldn't for exception handlers as it replaces the return of the block when there's no exception.

Sija: Good point, fixed!

nobodywasishere:
Will the correction remove any comments inside the else block?

veelenga (approved):
lgtm

PR #606, opened 2025-04-04
Do not report `focus` values which are calls or vars in `Lint/SpecFocus`

### LINKED ISSUE #605: `Lint/SpecFocus` should be more focused
Say you have a method that wraps an `it` block to make testing common assertions easier E.g.

```cr
private def expect_to_s(original, expected = original, file = __FILE__, line = __LINE__, focus = false)
  it "does to_s of #{original.inspect}", file, line, focus: focus do
    # ...
  end
end
```
This'll get flagged as having a focused spec item. It's not technically _wrong_ per se, but maybe there's something we could do to improve this context? Maybe only if its a bool literal? This rule also isn't as impactful as it was in the past now that the spec runner itself will return a non-zero exit code.

Sija:
Resolves #605

### DIFF
--- a/src/ameba/rule/lint/spec_focus.cr
+++ b/src/ameba/rule/lint/spec_focus.cr
@@ -65,7 +65,9 @@ module Ameba::Rule::Lint
       return unless node.block
 
       arg = node.named_args.try &.find(&.name.== "focus")
-      return unless arg
+      return if arg.nil? ||
+                arg.value.is_a?(Crystal::Call) ||
+                arg.value.is_a?(Crystal::Var)
 
       issue_for arg, MSG
     end
--- a/spec/ameba/rule/lint/spec_focus_spec.cr
+++ b/spec/ameba/rule/lint/spec_focus_spec.cr
@@ -60,9 +60,29 @@ module Ameba::Rule::Lint
       subject.catch(s).should_not be_valid
     end
 
+    it "reports if there is a spec item with `focus: !true`" do
+      s = Source.new %(
+        it "it", focus: !true do
+        end
+      ), path: "source_spec.cr"
+
+      subject.catch(s).should_not be_valid
+    end
+
     it "does not report if there is non spec block with :focus" do
       s = Source.new %(
         some_method "foo", focus: true do
+        end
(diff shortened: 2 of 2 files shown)


straight-shoota:
I know I often write `focus: !true`. That wouldn't be caught with this change. And it wouldn't fail the spec suite either.

Sija:
@straight-shoota That's intentional.

straight-shoota:
Huh, why?

Sija:
I reckon because any `skip` option leftover is treated as a code smell and I'd agree as this is akin to leaving `if false` debug constructs. Even if used as you said, such code IMO shouldn't be committed.

Sija:
Actually I think it wouldn't be reported since it's not a `BoolLiteral` anymore.

Sija:
> Yes, expressions such as `!true` should be a violation. But with this change, they wouldn't.

Agreed, I misread your previous comment.

src/ameba/rule/lint/spec_focus.cr
```diff
@@ -65,7 +65,7 @@ module Ameba::Rule::Lint
       return unless node.block
 
       arg = node.named_args.try &.find(&.name.== "focus")
-      return unless arg
+      return if arg.nil? || arg.value.is_a?(Crystal::Call)
```
veelenga, line 68: Will this resolve the OP's example, when the value is an argument?

```crystal
def foo(focus = false)
  it "foo", focus: focus do
  end
end
```

Sija: It doesn't matter, this rule operate on calls only.

veelenga: But it means it does not resolve #605, right? Sorry, I may have missed something.

Sija: It does, you have the test added to cover it.

Sija: I've fixed the code and modified the test case to cover this exact code path.

veelenga (approved):
LGTM, thanks

Issue #566, opened 2025-02-15
`ameba --gen-config` produces a mess


straight-shoota:
When adding `ameba` to a project, it seems like a good idea to run `ameba --gen-config` to get a neat starting point.
The idea is that this command generates a `.ameba.yml` configuration that records all rule violations that currently in the code base. That gives you time spread the adoption process and fix existing issues over time while ideally avoiding to add additional ones.

However I've found this process is not very straightforward and the autogenerated configuration has some major issues.
This means a great deal for user experience, especially when trying to adopt ameba into an existing codebase of non-trivial size.

As a practical example, I have run `ameba --gen-config` on the shards (using ameba 1.6.4). The codebase is about 8K sloc big and ameba found 383 violations. It produces a configuration file that's 357 lines long.
https://github.com/straight-shoota/shards/blob/011ddeef104e996d01ffd774f437ae0467688c7e/.ameba.yml

1. The entries in the config file are in complete disorder. It seems they are sorted by filename of the violations or something like that. When removing the violation of `Documentation/DocumentationAdmonition` in the first reported file (`src/helpers.cr`), a newly generated config removes that exclusion, but now the entries are in a _different_ order. That's very inconvenient for making iterative changes.
Configuration entries should be ordered by rule name. That would also group rules based on category (lint, naming, style etc.).
2. The level of detail in the generated default configuration is surprising. 
   	What am I to make of a list of 17 `AllowedNames` for the `Naming/BlockParameterName` rule? Are those representing the default configuration? If so, why would I want it to be printed out? I would usually trust the default configuration and don't want to mess with it unless I have some specific need for it.
     Also, every config entry has a `Severity` property. It's nice that I _could_ change the severity of individual rules, but I figure there would rarely be a need to. Nor do I see a need to have this available for informational purposes. Most users will be happy with whatever the default is. No need for this noise.
      Another issue is that writing out the _current_ default configuration values freezes them. Should the default change for a good reason in future versions, this config setting would still explicitly reference the old default. This might have some value in terms of consistency. But written out default value is indistinguishable from an explicit override. Without proper change management, you might miss useful updates in future versions.
      The current development version of `ameba` adds another curious configuration property for all entries: `SinceVersion`. This could be useful information, I guess. Not sure what exactly it would help with, though. But the more pressing question: Why is this a configuration option that I could presumably change? What would happen if I changed this setting?

2. The configuration entries seem to mix up different concerns, which I think should better be handled separately.
[...]

PR #610, opened 2025-05-27
Sort TODO issues by rule name

straight-shoota:
Addresses issue 1 from #566


straight-shoota:
The spec failure on nightly seems unrelated to this change.

veelenga (approved):
LGTM

nobodywasishere (approved):

spec/ameba/formatter/todo_formatter_spec.cr
```diff
@@ -118,6 +118,47 @@ module Ameba
...
+            # Problems found: 1
+            # Run `ameba --only BreakingRule` for details
+            BreakingRule:
+              Description: A rule with a custom name.
+              Excluded:
+              - source1.cr
+              Enabled: true
+              Severity: Convention
+
+            YAML
```
Sija, line 158: Let's keep the test in line with others in the spec file. This will also ensure it won't break when the version changes.

```suggestion
          content.should contain <<-YAML
            # Problems found: 1
            # Run `ameba --only Ameba/DummyRule` for details
            Ameba/DummyRule:
              Description: Dummy rule that does nothing.
              Dummy: true
              Excluded:
              - source2.cr
              Enabled: true
              Severity: Convention

            # Problems found: 1
            # Run `ameba --only BreakingRule` for details
            BreakingRule:
              Description: A rule with a custom name.
              Excluded:
              - source1.cr
              Enabled: true
              Severity: Convention
            YAML
```

Sija (changes requested):

src/ameba/formatter/todo_formatter.cr
```diff
@@ -22,6 +22,7 @@ module Ameba::Formatter
         return
       end
 
+      issues.sort_by!(&.rule.name)
```
Sija, line 25: I'd also update the `rule_todo` method to sort the issues by filename.

Sija: I'd add another issue to the test case with different filename to ensure the ordering is kept though.

Sija (approved):

PR #616, opened 2025-05-28
Remove noise from `--gen-config`

straight-shoota:
This addresses a large portion of #566.
It removes all non-essential information from the generated config file.
The only remaining contents are the header, `Version` property and for each violated rule the list of `Excludes`.

Depends on #610


veelenga (approved):
lgtm

PR #625, opened 2025-06-05
`Lint/UselessAssign`: Fix remove variable scope inheritance in `def`

### LINKED ISSUE #623: `Lint/UselessAssign` unaware of `def` and `fun` variables scopes
Apparently, `Lint/UselessAssign` is unaware of the fact that a `def` constitutes an entirely different context for variables which has zero connection to local variables declared outside of it.
Yet it seems to merge variables of the same name in top-level code and def body.

```cr
x = 1 # Useless assignment to variable `x`

def bar
  x = 2
end

puts x
```

Expected behaviour is that `x = 2` should be considered a useless assign.

straight-shoota:
Resolves #623


src/ameba/ast/scope.cr
```diff
@@ -195,6 +195,16 @@ module Ameba::AST
       outer_scope.nil?
     end
 
+    def inherits?
```
Sija, line 198: This reads better, as adjective demands a following noun, which there's none in this case.

```suggestion
    def inherited?
```

src/ameba/ast/scope.cr
```diff
@@ -195,6 +195,16 @@ module Ameba::AST
...
+      !(node.is_a?(Crystal::Def) || node.is_a?(Crystal::FunDef))
```
Sija, line 199: For completeness sake I'd add enums as well.

```suggestion
      !(node.is_a?(Crystal::Def) || node.is_a?(Crystal::FunDef) || node.is_a?(Crystal::EnumDef))
```

src/ameba/ast/scope.cr
```diff
@@ -195,6 +195,16 @@ module Ameba::AST
...
+    end
+
+    def inherits?(&)
+      if inherits?
+        yield
+      end
```
Sija, line 205:
```suggestion
      yield if inherits?
```

Sija (changes requested):

src/ameba/ast/scope.cr
Sija: They can't indeed. Disregard then.

Sija (approved):
LGTM

src/ameba/ast/scope.cr
```diff
@@ -145,7 +145,7 @@ module Ameba::AST
     # Returns `true` if type declaration variable is assigned in this scope.
     def assigns_type_dec?(name)
       type_dec_variables.any?(&.name.== name) ||
-        !!outer_scope.try(&.assigns_type_dec?(name))
+        inherited? { !!outer_scope.try(&.assigns_type_dec?(name)) }
```
Sija:
```suggestion
        !!inherited? { outer_scope.try(&.assigns_type_dec?(name)) }
```

PR #626, opened 2025-06-06
`Lint/UselessAssign`: Skip assigns in call args

### LINKED ISSUE #447: `Lint/UselessAssign` reports macros accepting `Crystal::Macros::TypeDeclaration`

straight-shoota:
Assignments in method call args are very unusual. They should likely be considered a style violation, regardless of whether they are useless or not.
However, assignments in arguments to macro calls are a quite common use case (for example, `record` and `property` macros). Currently, there are some hard coded special case rules to ignore calls to these well-known stdlib macros. But this only trying to paint over the most prominent failures. It does not solve the problem. Calls to custom macros are still affected. Same for the `ExcludeTypeDeclarations` config setting, which serves as a similar crutch to reduce a portion of false positives.

We cannot tell whether a call is to a macro or a method without semantic analysis. So currently, such macro calls trigger false positives.

This change ignores assignments inside call args and removes the special-case code that dealt with some well-known macros.

As a result, it may fail to recognize some actual violations when an assignment in a method call argument is useless.
I believe this is a good deal, though. Useless assigns are not a critical issue.
If we cannot achieve the necessary accuracy, it's better to miss out on some violations than spam users with many non-violations.

Resolves #447

### DIFF
--- a/spec/ameba/rule/lint/useless_assign_spec.cr
+++ b/spec/ameba/rule/lint/useless_assign_spec.cr
@@ -402,12 +402,13 @@ module Ameba::Rule::Lint
           CRYSTAL
       end
 
-      pending "doesn't report type declaration as a call argument" do
+      it "doesn't report type declaration as a call argument" do
         expect_no_issues subject, <<-CRYSTAL
-          foo Foo(T), foo : T
-          foo Foo, foo : Nil
-          foo foo : String
-          foo foo : String, bar : Int32?, baz : Bool
+          foo bar : String
+          foo bar : String = ""
+          foo bar : String = baz
+          foo bar = baz
+          foo bar = ""
           CRYSTAL
       end
 
--- a/src/ameba/ast/scope.cr
+++ b/src/ameba/ast/scope.cr
@@ -34,6 +34,8 @@ module Ameba::AST
     # The actual AST node that represents a current scope.
     getter node : Crystal::ASTNode
 
+    getter? in_call_args : Crystal::ASTNode? = nil
+
     delegate location, end_location, to_s,
       to: @node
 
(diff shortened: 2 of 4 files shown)


Sija:
> Assignments in method call args are very unusual.

They're unusual perhaps, but they do happen, especially when someone comes from Python-esque background, where passing keyword arguments is done with `=` instead of `:`, creating an assignment in Crystal world.



> However, assignments in arguments to macro calls are a quite common use case (for example, `record` and `property` macros). Currently, there are some hard coded special case rules to ignore calls to these well-known stdlib macros. But this only trying to paint over the most prominent failures. It does not solve the problem. Calls to custom macros are still affected. Same for the `ExcludeTypeDeclarations` config setting, which serves as a similar crutch to reduce a portion of false positives.
> 
> We cannot tell whether a call is to a macro or a method without semantic analysis. So currently, such macro calls trigger false positives.

Agreed.

> This change ignores assignments inside call args and removes the special-case code that dealt with some well-known macros.

I don't like the fact that the implementation changes a general visitor class for just this particular example. Feels to me like another hack, this time touching upon more generalized structures.

> As a result, it may fail to recognize some actual violations when an assignment in a method call argument is useless.
> I believe this is a good deal, though. Useless assigns are not a critical issue.
> If we cannot achieve the necessary accuracy, it's better to miss out on some violations than spam users with many non-violations.

I'd agree if that was a common situation, yet to the best of my knowledge only particular (heavy macro users in general) apps/shards are affected.

straight-shoota:
Could you please clarify what changes you'd like to see to the implementation? I'm not quite sure what's expected.

IMO it's totally okay for a scope visitor to have a configuration option that it should not descent into some specific locations (such as call args). And I don't see any alternative with a diff nearly as simple as this one.

PR #628, opened 2025-06-06
Fix specs to support Crystal 1.17

Sija:


straight-shoota:
IMO this is a regression. We should try to fix the behaviour.

Sija:
@straight-shoota It is, yet in the same time it seems like a correct behavior - given the inner workings of ECR and `loc` pragmas.

straight-shoota:
I don't follow. Why would it be correct to introduce a line break between the target and call. Is there a good reason?

Sija:
That's what the `loc` pragmas inserted by the ECR are explicitly stating 🤷‍♂️ 
See [my previous message](https://discord.com/channels/591460182777790474/591597160492171264/1377809949740105840) on Discord.

PR #630, opened 2025-06-07
Fix regression in `Lint/LiteralsComparison`

### LINKED ISSUE #619: `Lint/LiteralsComparison` only considers comparisons between equal values
The `Lint/LiteralsComparison` complains about comparisons between literals, but only if both operands are equal (`1 == 1`, `1 != 1`).
It does not complain if the operands are unequal (`1 == 2`, `1 != 2`).
This seems incomplete. I don't see any reason why it would make a difference whether the operands are equal or not.

It appears this was changed in https://github.com/crystal-ameba/ameba/pull/417, probably unintentionally.

Sija:
Fixes #619


nobodywasishere (approved):

veelenga (approved):
LGTM

PR #636, opened 2025-06-11
Add `Lint/SelfInitializeDefinition` rule

Sija:
Inspired by https://github.com/crystal-lang/crystal/pull/15824

### DIFF
--- a/spec/ameba/rule/lint/self_initialize_definition_spec.cr
+++ b/spec/ameba/rule/lint/self_initialize_definition_spec.cr
@@ -1,0 +1,60 @@
+require "../../../spec_helper"
+
+module Ameba::Rule::Lint
+  describe SelfInitializeDefinition do
+    subject = SelfInitializeDefinition.new
+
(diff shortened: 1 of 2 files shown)


spec/ameba/rule/lint/initialize_definition_with_receiver_spec.cr
```diff
@@ -0,0 +1,39 @@
...
+          end
+          CRYSTAL
+      end
+
+      it "fails for `initialize` method definition with a receiver" do
+        expect_issue subject, <<-CRYSTAL
+          {{ keyword }} Foo
+          end
+
+          def Foo.initialize
```
Sija, line 32: I'm not sure if this case should be flagged since it's more likely intentional.

veelenga: is there a difference between `def Foo.initialize` and just `def initialize` ?

veelenga (approved):
lgtm

PR #639, opened 2025-06-14
Add `Lint/UselessVisibilityModifier` rule

### LINKED ISSUE #559: Add `Lint/UselessDef` and `Lint/UselessVisibilityModifier`
This may be something that should be a compiler warning instead, so feel free to close if it should be. It's possible to define operator methods top-level that are impossible to call:

```crystal
def +(other)
  p! other
end

+1 # => 1
```

As well, it's possible to define a method with `protected` top-level, when the visibility modifier has no effect

```crystal
protected def hello(a)
  p! a
end

hello(1)
# a # => 1
```

See: https://github.com/crystal-lang/crystal/issues/12360

Sija:
Resolves #559


veelenga (approved):
lgtm

PR #641, opened 2025-06-18
Add git sha to the printed version

### LINKED ISSUE #640: Version information of an `ameba` build is lacking
`ameba --version` prints the version of the program. It prints the version number, which is sufficient to identify the build version of a release.
But this is rather insufficient when working with non-release builds, such as development builds from master. A build from any commit in master since 1.6.4 identifies as 1.7.0-dev without any means to differentiate.

It would be helpful to add more context, for example the git commit hash, and maybe also the build date for additional context.

For example, the output of `crystal --version`:

```console
$ crystal --version | head -n1
Crystal 1.16.3 (2025-05-12
$ crystal-nightly --version | head -n1
Crystal 1.17.0-dev [c5af98371] (2025-06-16)
```

(Note I'm piping through `head -n1` because `crystal --version` shows additional information which is not relevant for this discussion).

A drawback is that this makes the version harder to parse. It should still be reasonably easy to extract the second output column (e.g. `cut -d' ' -f2`). But if some tools expect the output of `ameba --version` to be _just_ a version number, they'll be broken.

Sija:
Resolves #640

### DIFF
--- a/src/ameba.cr
+++ b/src/ameba.cr
@@ -18,8 +18,6 @@
 module Ameba
   extend self
 
-  VERSION = {{ `shards version "#{__DIR__}"`.chomp.stringify }}
-
   macro ecr_supported?(&)
     {% if compare_versions(Crystal::VERSION, "1.15.0") >= 0 %}
       {{ yield }}
--- a/src/ameba/cli/cmd.cr
+++ b/src/ameba/cli/cmd.cr
@@ -226,7 +226,11 @@ module Ameba::Cli
   end
 
   private def print_version
-    puts VERSION
+    if GIT_SHA
+      puts "%s [%s]" % {VERSION, GIT_SHA}
+    else
+      puts VERSION
+    end
     exit 0
   end
 
--- a/src/ameba/version.cr
+++ b/src/ameba/version.cr
@@ -1,0 +1,6 @@
+module Ameba
+  extend self
+
+  VERSION = {{ `shards version "#{__DIR__}"`.chomp.stringify }}
+  GIT_SHA = {{ `(git rev-parse --short HEAD || true) 2>/dev/null`.chomp.stringify }}.presence
+end


veelenga (approved):
lgtm

straight-shoota:
I think it would be useful to pass in a commit hash per environment variable at compile time. This allow building from a non-git source (e.g. a tarball download) while still being able to reference the respective version.

For the crystal compiler we use `CRYSTAL_CONFIG_BUILD_COMMIT` for this, for example. It's configured in the make recipe.

veelenga:
> I think it would be useful to pass in a commit hash per environment variable at compile time. This allow building from a non-git source (e.g. a tarball download) while still being able to reference the respective version.
> 
> For the crystal compiler we use `CRYSTAL_CONFIG_BUILD_COMMIT` for this, for example. It's configured in the make recipe.

https://github.com/crystal-lang/crystal/blob/63650d76a4568bcae9e17521f75be7891b5c92cf/Makefile#L48

@straight-shoota could you please elaborate how the `CRYSTAL_CONFIG_BUILD_COMMIT` is passed for tarball download?

straight-shoota:
> @straight-shoota could you please elaborate how the `CRYSTAL_CONFIG_BUILD_COMMIT` is passed for tarball download? Is it captured at the moment the tarball is created?

This does not happen automatically. Unfortunately, tarballs from GitHub contain only the raw code and no indication of the source version.

But having the environment variable exposed allows doing something like this:
```sh
export SHA=63650d7
download_tarball https://github.com/crystal-lang/crystal/archive/refs/$SHA.tar.gz
make CRYSTAL_CONFIG_BUILD_COMMIT=$SHA
```

Sija:
Theoretically yes, but at the current stage we don't have any release procedures outside of directly building from git tree, whether it's done directly or via Homebrew, CI or GitHub Action. Linux packages use version instead of git commit sha. Also, mind you that's not what you've requested in #640.

> I'd like to be able to explicitly specify the commit hash when git information is unavailable.

Where and why you want this functionality?

PR #556, opened 2025-01-29
Add `Lint/UnusedSelfAccess` and `Lint/UnusedPseudoMethodCall`

nobodywasishere:
Closes #543

PR #646, opened 2025-06-24
Consolidate unused expression rules into one `Lint/UnusedExpression` rule

### LINKED ISSUE #645: Many separate rules for the core problem of unused expressions
There are currently eight different rules about essentially the same core problem: unused expressions with no expected side effects.

I'm wondering why these are all individual rules. They are all about the same concept, just for different expression types.
The implementations are generally very similar except for some small special-case modifications (all are just different filters for `ImplicitReturnVisitor`).

These following rules are conceptually very similar:

* `Lint/UnusedClassVariableAccess`
* `Lint/UnusedComparison`
* `Lint/UnusedGenericOrUnion`
* `Lint/UnusedInstanceVariableAccess`
* `Lint/UnusedLiteral`
* `Lint/UnusedLocalVariableAccess`
* `Lint/UnusedPseudoMethod`
* `Lint/UnusedSelfAccess`

They were all added recently in #552, #556, #537, #507 and are not released yet.

What's the point of having all these as individual rules instead of a combined one which handles all cases of unused expressions?

It would be easier to have only a single rule which applies to a couple different kinds of expressions. Fewer things that users need to be aware of.
Name suggestion for a combined rule: `Lint/UnusedExpression`.

If there is any need to disable this rule for some specific expression types, it should be possible to introduce a configuration option that allows opting out at the same granularity as the separate rules currently do.

nobodywasishere:
Closes #645
Refs #507, #537, #552, #556


src/ameba/ast/visitors/implicit_return_visitor.cr
```diff
@@ -181,6 +181,14 @@ module Ameba::AST
...
     end
 
+    def visit(node : Crystal::TypeOf) : Bool
+      report_implicit_return(node)
+
+      incr_stack { node.accept_children(self) }
+
+      false
+    end
+
```
Sija, line 191: IIRC support for this was ruled out since `typeof` is used in specs because of its side-effects.

src/ameba/rule/lint/unused_expression.cr
```diff
@@ -0,0 +1,157 @@
...
+  #
+  # bar = pointerof(foo)
+  # ```
+  #
+  # This rule currently supports checking for unused:
+  # - comparison operators: `<`, `>=`, etc.
+  # - generics and unions: `String?`, `Int32 | Float64`, etc.
+  # - literals: strings, bools, chars, hashes, arrays, range, etc.
+  # - pseudo-method calls: `sizeof`, `is_a?` etc.
+  # - variable access: `local`, `@ivar`, `@@cvar`, `self`
```
Sija, line 34:
```suggestion
  # - variable access: local, `@ivar`, `@@cvar` and `self`
```

src/ameba/rule/lint/unused_expression.cr
```diff
@@ -0,0 +1,157 @@
...
+    CLASS_VAR_MSG = "Value from class variable access is unused"
+
+    def test(source, node : Crystal::ClassVar, in_macro : Bool)
+      # Class variables aren't supported in macros
+      return if in_macro
+
+      issue_for node, CLASS_VAR_MSG
+    end
+
+    COMPARISON_MSG       = "Comparison operation is unused"
```
Sija, line 61: Please group the constants together, splitting them to ones containing issue messages and others.
Also, please stick to `MSG_XXX` format.

Sija (changes requested):

src/ameba/ast/visitors/implicit_return_visitor.cr
Sija: I see, thanks for clarification ❤️

src/ameba/rule/lint/unused_expression.cr
```diff
@@ -0,0 +1,151 @@
...
+  #
+  # bar = pointerof(foo)
+  # ```
+  #
+  # This rule currently supports checking for unused:
+  # - comparison operators: `<`, `>=`, etc.
+  # - generics and unions: `String?`, `Int32 | Float64`, etc.
+  # - literals: strings, bools, chars, hashes, arrays, range, etc.
+  # - pseudo-method calls: `sizeof`, `is_a?` etc.
+  # - variable access: local, `@ivar`, `@@cvar` and `self`
```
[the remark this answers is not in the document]
straight-shoota: IMO this is not an important feature. As far as I can tell, this rule with all its variants should be pretty solid. The chance for false positives is very low.
And in case there is an issue, there's not much harm in just disabling the entire rule instead of just individual aspects of it.

So unless there's a clear use case why one would want to disable a subset of features, it's probably not worth it.

PR #647, opened 2025-06-29
Rename `Documentation/DocumentationAdmonition` rule to just `Documentation/Admonition`

Sija:


Sija:
@nobodywasishere True, that was my concern as well and the next PR gonna disable this rule by default.

nobodywasishere (approved):

PR #649, opened 2025-06-29
Add `exclude_multiline` setting to `Style/ParenthesesAroundCondition` rule

### LINKED ISSUE #615: `Style/ParenthesesAroundCondition` with multi-line conditions
When a condition spans multiple lines, I believe parenthesis might serve a function to define a delimiter for the condition.
Compare the following examples, IMO the first one is better readable.
But `Style/ParenthesesAroundCondition` complains about a violation.

```cr
if (
     1 &&
     2
   )
  foo
end

if
    1 &&
    2
  foo
end
```

It might make sense to exclude parentheses around complex conditions spanning multiple lines.

However, in case this stays a violation, the autofix needs improvement. It only removes the parenthesis characters, thus leaving trailing whitespace on both lines.

Sija:
Resolves #615

### DIFF
--- a/src/ameba/rule/style/parentheses_around_condition.cr
+++ b/src/ameba/rule/style/parentheses_around_condition.cr
@@ -24,6 +24,7 @@ module Ameba::Rule::Style
   # Style/ParenthesesAroundCondition:
   #   Enabled: true
   #   ExcludeTernary: false
+  #   ExcludeMultiline: false
   #   AllowSafeAssignment: false
   # ```
   class ParenthesesAroundCondition < Base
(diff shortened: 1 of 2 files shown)


src/ameba/rule/style/parentheses_around_condition.cr
```diff
@@ -32,12 +33,14 @@ module Ameba::Rule::Style
...
 
       exclude_ternary false
+      exclude_multiline false
       allow_safe_assignment false
     end
 
     MSG_REDUNDANT = "Redundant parentheses"
     MSG_MISSING   = "Missing parentheses"
 
+    # ameba:disable Metrics/CyclomaticComplexity
```
veelenga, line 43: do we want to break down the method to reduce the complexity? If we don't follow our own suggetions, who would follow? 😄

veelenga (approved):
lgtm

PR #650, opened 2025-06-30
Spec tweaks

Sija:


spec/ameba/runner_spec.cr
```diff
@@ -138,41 +138,42 @@ module Ameba
...
+        output.clear
+      end
 
       it "writes nothing if sources are valid" do
-        io.clear
         runner = runner(formatter: formatter).run
-        runner.explain({file: "source.cr", line: 1, column: 2}, io)
-        io.to_s.should be_empty
+        runner.explain({file: "source.cr", line: 1, column: 2}, output)
+        output.to_s.should be_empty
```
straight-shoota, line 158: **suggestion:** Is there a specific reason to reuse a shared `IO::Memory` instance between all these examples? Wouldn't a simple `String.build` be sufficient?
It's usually better to encapsulate the individual examples. Makes the code easier to reason about as well.

```suggestion
        String.build do |output|
          runner.explain({file: "source.cr", line: 1, column: 2}, output)
        end.should be_empty
```

Sija: That's convention used in several places within the specs and personally I don't mind the either option, although I'd prefer to refactor it once in every place, not just in here - preferably in a separate PR.

veelenga (approved):
lgtm

PR #651, opened 2025-06-30
Add `Ext::Location#same_line?` helper method

Sija:


src/ameba/rule/style/verbose_block.cr
```diff
@@ -186,10 +186,7 @@ module Ameba::Rule::Style
     end
 
     private def same_location_lines?(a, b)
-      return unless a_location = name_location(a)
-      return unless b_location = b.location
-
-      a_location.line_number == b_location.line_number
+      name_location(a).try &.same_line?(b.location)
```
veelenga, line 189: is this one-liner method needed anymore?

PR #655, opened 2025-08-11
Fix issue reporting single-line constant assigns inside namespaces

nobodywasishere:


spec/ameba/rule/naming/constant_names_spec.cr
```diff
@@ -34,6 +34,12 @@ module Ameba::Rule::Naming
         CRYSTAL
     end
 
+    it "Doesn't report single-line constant assigns inside namespaces" do
+      expect_no_issues subject, <<-CRYSTAL
+        Example::FOO_BAR = "bar"
+        CRYSTAL
+    end
```
Sija, line 41:
```suggestion
    it "doesn't report single-line constant assigns inside namespaces" do
      expect_no_issues subject, <<-CRYSTAL
        Example::FOO_BAR = "bar"
        CRYSTAL
    end
```

Sija (commented):
Change is correct and logically sound, the test seem misleading though since you cannot declare constants dynamically.

Sija (approved):

straight-shoota:
> it didn't work with `pp Foo::BAR = :foobar`

Because `pp` is a method, so its arguments are a dynamic context. Would be different if it was a macro.
(`pp!` is a macro, but it delegates to `pp` 🤷)

PR #665, opened 2025-10-06
Simplify adding issues for node names

Sija:


veelenga (approved):
Was thinking about something similar while reviewing previous PR.
LGTM 👍🏻

PR #670, opened 2025-10-20
Fix missing logic for block arguments

Sija:
Before this change, block arguments weren't treated as blocks themselves, leading to buggy behavior in many cases.


spec/ameba/ast/util_spec.cr
```diff
@@ -308,6 +308,28 @@ module Ameba::AST
...
+        subject.has_block?(node).should be_true
+      end
+
+      it "returns true if the node has a block (shorthand)" do
+        node = as_node("%w[foo bar].find(&.starts_with?('f'))")
+        subject.has_block?(node).should be_true
+      end
+
+      it "returns true if the node has a block (block argument)" do
+        node = as_node("%w[foo bar].find(&block)")
```
nobodywasishere: (minor) could add a test case for `&->block`

nobodywasishere (approved):

PR #672, opened 2025-10-29
Add `Performance/TimesMap` rule

### LINKED ISSUE #654: Feature: Lint rule to transform `times.map.to_a` to `Array.new`
It seems to be a somewhat common pattern to write code like `3.times.map { |i| foo(i) }.to_a` to create an array of 3 items, maybe with each item referring to its index.

This seems a bit convoluted though, and we could instantiate an array instance directly instead: `Array.new(4){ |i| foo(i) }`. That's shorter, more straightforward and simpler overall.

I'm proposing a rule that suggests transforming `%number.times.map(&%block).to_a` to `Array.new(%number, &%block)`.

Sija:
Resolves #654

---

Brings improvement of ~5%.

```cr
require "benchmark"

Benchmark.ips do |x|
  x.report("X1") { Array.new(10_000) { |i| "foo-#{i}" } }
  x.report("X2") { 10_000.times.map { |i| "foo-#{i}" }.to_a }
end
```

```
X1   1.97k (507.79µs) (± 2.01%)  1.75MB/op        fastest
X2   1.87k (535.46µs) (± 1.16%)  2.08MB/op   1.05× slower
```

`crystal --version`:

```
Crystal 1.18.2 (2025-10-21)

LLVM: 18.1.8
Default target: aarch64-apple-darwin21.6.0
```

### DIFF
--- a/src/ameba/rule/performance/times_map.cr
+++ b/src/ameba/rule/performance/times_map.cr
@@ -1,0 +1,66 @@
+require "./base"
+
+module Ameba::Rule::Performance
+  # This rule is used to identify usage of `times.map { ... }.to_a` calls.
+  #
+  # For example, this is considered invalid:
+  #
+  # ```
+  # 5.times.map { |i| i * 2 }.to_a
+  # ```
+  #
+  # And it should be written as this:
+  #
+  # ```
+  # Array.new(5) { |i| i * 2 }
+  # ```
+  #
+  # YAML configuration example:
+  #
+  # ```
+  # Performance/TimesMap:
+  #   Enabled: true
+  # ```
+  class TimesMap < Base
+    include AST::Util
+
+    properties do
+      since_version "1.7.0"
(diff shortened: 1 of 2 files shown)


src/ameba/rule/performance/times_map.cr
```diff
@@ -0,0 +1,66 @@
...
+  # ```
+  class TimesMap < Base
+    include AST::Util
+
+    properties do
+      since_version "1.7.0"
+      description "Identifies usage of `times.map { ... }.to_a` calls"
+    end
+
+    MSG = "Use `Array.new(%s) {...}` instead of `times.map {...}.to_a`"
```
nobodywasishere: (Minor)

```suggestion
    MSG = "Use `Array.new(%s) {...}` instead of `%s.times.map {...}.to_a`"
```

nobodywasishere (approved):

PR #680, opened 2025-11-08
Add `Lint/DuplicateEnumValue` rule

### LINKED ISSUE #660: Add `Lint/DuplicateEnumValue`
```crystal
enum Foo
  BAR = 0
  BAZ = 111
  QUX = 111
      # ^^^ error: Enum member value already used
end
```

Sija:
Resolves #660

### DIFF
--- a/spec/ameba/rule/lint/duplicate_enum_value_spec.cr
+++ b/spec/ameba/rule/lint/duplicate_enum_value_spec.cr
@@ -1,0 +1,50 @@
+require "../../../spec_helper"
+
+module Ameba::Rule::Lint
+  describe DuplicateEnumValue do
+    subject = DuplicateEnumValue.new
+
+    it "passes if there are no enum members with values" do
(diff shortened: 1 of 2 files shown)


straight-shoota (commented):
Would be nice to have a test case covering assignment of non-literals, for example aliasing a member.
```cr
enum Foo
  BAR = 1
  BAZ = BAR
end
```
I presume this should pass.

nobodywasishere:
There are other edge cases this should probably handle:

```
enum MyEnum
  Foo = 1
  Bar
  Baz = 2 # duplicate of Bar
end
```

```
enum MyEnum
  Foo
  Bar = 0 # duplicate of Foo
end

enum MyEnum2
  Foo
  Bar
  Baz
  Qux = 1 # duplicate of Bar
end
```

Sija:
Mixing members having values with the ones without is, in the first place, a code smell to me, so I'd rather make another rule than add more logic to this one.

PR #698, opened 2025-11-15
Remove documentation link in JSON Schema rule description

Sija:
It looks bad and is not clickable :(

Refs #497


nobodywasishere:
Will need to backport this as well

Issue #532, opened 2024-12-07
Versioned documentation


Sija:
It would be great to have online documentation versioned, in order to be able to link to it in our tooling, since rules and rule options are subject to change between versions.

References:

- https://github.com/crystal-ameba/ameba/pull/531#discussion_r1874552324
- https://github.com/crystal-ameba/vscode-crystal-ameba/pull/146

Sija:
Gameplan draft:

1. Remove everything from `gh-pages` branch
2. Commit `versions.json` file (see for example [Crystal's one](https://crystal-lang.org/api/versions.json)), containing a single `nightly` entry
3. Tweak the [docs](https://github.com/crystal-ameba/ameba/blob/5b3c2cfcf2b9c720fb742913d50bdab49f9933ea/.github/workflows/docs.yml) workflow by:
  - Adding `--json-config-url=/ameba/versions.json` argument to `crystal docs` invocation
  - Building the docs on every push to the `master` branch with `target-folder: master`
  - Building the docs on every release with `target-folder: <version>` & appending the version into the `versions.json`
4. Run the workflow for `master` & every published version so far
5. Create `index.html` in the root directory with redirect to `/master`

/cc @veelenga

PR #704, opened 2025-11-20
Implementing version list auto-update

Sija:
Refs #532

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -14,6 +14,7 @@ jobs:
   setup:
     runs-on: ubuntu-latest
     outputs:
+      is-released: ${{ steps.variables.outputs.is-released }}
       version: ${{ steps.variables.outputs.version }}
       ref: ${{ steps.variables.outputs.ref }}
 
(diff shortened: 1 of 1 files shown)


.github/workflows/docs.yml
```diff
@@ -38,7 +45,62 @@ jobs:
...
+          ref: gh-pages
+
+      - name: Update version list
+        run: |
+          VERSION="${{ needs.setup.outputs.version }}"
+          IS_RELEASED="${{ needs.setup.outputs.is-released }}"
+
+          JQ_EXPRESSION="
+            .versions |= (
+              if ( . | map(.name) | index(\"${VERSION}\") | not ) then
```
nobodywasishere: Will this remove or replace a version if it already exists?

Sija: Conceptually it could be translated to Crystal as:

```cr
versions = 
  if !versions.map(&.name).includes?(VERSION)
    versions << Version.new(name: VERSION, url: "/ameba/#{VERSION}/", released: IS_RELEASED)
  else
    versions
  end
```

PR #718, opened 2025-12-12
Expand CLI glob args against project root directory

Sija:
```shell
ameba /path/to/project '!**/foo.cr'
```

**Before**: glob matches files in current directory (`pwd`)
**After**: glob matches files in project directory (`/path/to/project`)

Extracted from #713
Refs https://github.com/crystal-ameba/ameba/pull/713#discussion_r2598047156


Sija:
@straight-shoota I agree, it's certainly unexpected to see different results based on such subtle markers.

PR #736, opened 2025-12-20
Add `Style/VerboseNilType` rule

Sija:


spec/ameba/rule/style/verbose_nil_type_spec.cr
```diff
@@ -0,0 +1,76 @@
...
+
+    it "passes if the union includes metaclass (Foo.class)" do
+      expect_no_issues subject, <<-CRYSTAL
+        foo : Foo.class | Nil = nil
+        CRYSTAL
+    end
+
+    it "reports if there is a verbose nil type" do
+      source = expect_issue subject, <<-CRYSTAL
+        foo : String | Nil = nil
```
nobodywasishere: Could add a test case like:

```
foo : (Array(String | Nil) | Nil) | Foo
```

And would be curious to see what happens for:

```
foo : String? | Nil
bar : String | Nil?
```

nobodywasishere (approved):

spec/ameba/rule/style/verbose_nil_type_spec.cr
nobodywasishere: Some more to test:

```
foo : (String | Nil)?
bar : Nil | String
```

Sija: No, I'm sorry, you were actually right! 🙏 Gonna post the fix soon.

PR #749, opened 2026-01-08
Fix specs for Crystal 1.19.0

Sija:


nobodywasishere:
Are we going to increase the min required version to 1.19?

PR #754, opened 2026-01-10
Add `body_auto_dedent` property to `Style/HeredocIndent` rule

Sija:


nobodywasishere:
It could be frustrated if the indent is intended though and the linter keeps "fixing" it, though I understand the reason it exists

Sija:
As I said, 100% cases I've stumbled upon so far were exhibiting the signs of erroneous assumption more than the intended behavior. I was doin' the same thing some time ago, actually. And yes, I understand it could be frustrating, hence the option.

straight-shoota:
I don't think it's a good idea for a linter to change the contents of string literals. Even if it often corrects errors.
While the probability of being a good change is high, this autocorrect could still be incorrect and remove intentional indentation.

Fixing the indentation requires an explicit review to make sure it aligns with the intent of the code.
So IMO this should not be applied automatically. Certainly not enabled by default.

straight-shoota:
The review argument is difficult: Yes, all changes should be reviewed and tested. But humans get easily bored when reviewing automated repetitive changes. Especially ones that are usually expected to be quite safe.
But this change is very explicitly not safe. It's intentionally breaking, changing the contents of string literals.
In my opinion, such a dangerous behaviour should not be a default feature.

Especially since I don't see much gain. This is a niche issue. Not much code will be affected either way. So why take the risk?

That also makes it particularly hard to gauge the impact. So the fact that an ecosystem validation didn't show any issues is not very reassuring by itself.

Sija:
I get you see this differently and I'm fine with disagreeing on this with you.

PR #100, opened 2019-04-13
Feat/severity

veelenga:
Should close #39.

- [x] Each rule should have it's own severity
- [x] Severity could be one of `error`, `warning`, `refactoring`
- [x] Severity should be inferred by the name and category by default. For example:
  
  * [Layout/*](https://github.com/veelenga/ameba/tree/master/src/ameba/rule/layout) - refactoring
  * [Metrics/*](https://github.com/veelenga/ameba/tree/master/src/ameba/rule/metrics) - refactoring
  * [Performance/*](https://github.com/veelenga/ameba/tree/master/src/ameba/rule/performance) - refactoring
  * [Style/*](https://github.com/veelenga/ameba/tree/master/src/ameba/rule/style) - refactoring
  * [Lint/*](https://github.com/veelenga/ameba/tree/master/src/ameba/rule/lint) - warning
  * [Lint/Syntax](https://github.com/veelenga/ameba/blob/master/src/ameba/rule/lint/syntax.cr) - error

for now we will have only one rule with `error` severity but it can be changed later.
  
- [x] Severity should be customizable via config
- [x] User should be able to run only rules which have some severity: `ameba --fail-level SEVERITY`

For example:

* `ameba --fail-level refactoring` - default and currently existed behaviour. Fails if at least one rule of severity `refactoring`, `warning` or `error` fails
* `ameba --fail-level warning` - fails if at least one rule of severity `warning` or `error` fails
* `ameba --fail-level error` - fails if at least one rule of severity `error` fails

- [x] User should see an error if he provides inappropriate severity

### Usage example:

```sh
$ ameba                                                                                                                                                                                                          
Inspecting 143 files.

...........................................F...................................................................................................

spec/ameba/formatter/flycheck_formatter_spec.cr:12:9
[W] Lint/UselessAssign: Useless assignment to variable `test`
> test = 1
  ^
Finished in 409.4 milliseconds

143 inspected, 1 failure.
$ echo $?                                                                                                                                                                                                       
1

$ ameba --fail-level Error                                                                                                                                                                                        
Inspecting 143 files.

...........................................F...................................................................................................

spec/ameba/formatter/flycheck_formatter_spec.cr:12:9
[W] Lint/UselessAssign: Useless assignment to variable `test`
> test = 1
  ^
Finished in 408.64 milliseconds

143 inspected, 1 failure.
$ echo $?                                                                                                                                                                                                         
0
```

### Formatters

#### Dot Formatter
[...]

PR #759, opened 2026-01-12
Fix `--fail-level` CLI switch not having any effect

Sija:
Found this while working on https://github.com/crystal-ameba/github-action/issues/24. This slipped through the cracks for a looong time (I guess since the beginning, i.e., 575fe07879a3c309dfca2fa26cea78ac1e1b0734 - released as a part of `v0.9.2`), not a popular switch it seems 😅

Refs #100


veelenga (approved):
lgtm

PR #761, opened 2026-01-12
Replace `--fail-level` CLI switch with `--min-severity`

Sija:
### Before

Reporting all issues, regardless of their severity, and exiting with `0` or `1` error code based on the `--fail-level` flag — as in RuboCop.

### After

Limiting the issues reported based on the `--min-severity` — as in Credo.
_Exit behaviour is unaffected._

References:

- RuboCop: [`--fail-level`](https://docs.rubocop.org/rubocop/1.82/usage/basic_usage.html#command-line-flags)
- Credo: [`--min-priority`](https://hexdocs.pm/credo/cli_switches.html)

Rework of #759
Refs #100

### DIFF
--- a/spec/ameba/cli/cmd_spec.cr
+++ b/spec/ameba/cli/cmd_spec.cr
@@ -142,25 +142,25 @@ module Ameba::CLI
         end
       end
 
-      context "--fail-level" do
+      context "--min-severity" do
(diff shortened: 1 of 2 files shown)


veelenga (approved):
LGTM. Would be nice to cover it by unit tests

PR #807, opened 2026-04-07
Don’t install `bin/ameba.cr` executable

Sija:
This file gets overridden on every successful `shards install/update` call, which pretty much defeats its whole purpose - i.e., being able to customize your local ameba installation.

Instead, it's better to use the `lib/ameba/bin/ameba.cr` which already comes with the Ameba dependency, or just to copy it over to the `bin` directory for local modification—if necessary.

Refs #470

### DIFF
--- a/README.md
+++ b/README.md
@@ -126,7 +126,7 @@ target to the `shard.yml` file:
 ```yaml
 targets:
   ameba:
-    main: bin/ameba.cr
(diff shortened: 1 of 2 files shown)


veelenga (approved):
This looks reasonable. LGTM

Sija:
Heads up @hahwul @ellmetha

PR #809, opened 2026-04-08
Extract `Lint/NonExistentRule` from `Lint/BadDirective`

Sija:


src/ameba/rule/lint/non_existent_rule.cr
```diff
@@ -0,0 +1,50 @@
...
+  # ```
+  # Lint/NonExistentRule:
+  #   Enabled: true
+  # ```
+  class NonExistentRule < Base
+    include AST::Util
+
+    properties do
+      since_version "1.7.0"
+      description "Reports non-existent rules in comment directives"
```
veelenga, line 25: My first initial thought was to merge the BadDirective into UnneededDisableDirective to simplify it for end users. But I see you have a different vision and follow a more granular approach. What's the purpose?

Sija: I believe that error granularity improves insight into what's exactly wrong with the topic at hand, in this case, the ameba pragma. One might want to forgo non-existent rules (for example, ignoring non-bundled internal rules), while still flagging invalid pragma syntax.

veelenga (approved):
lgtm

PR #812, opened 2026-04-10
Move ameba-related rules into their own `Internal` group

Sija:
- `Lint/BadDirective` -> `Internal/BadDirective`
- `Lint/DeprecatedRule` -> `Internal/DeprecatedRule`
- `Lint/UnneededDisableDirective` -> `Internal/UnneededDisableDirective`
- `Lint/NonExistentRule` -> `Internal/NonExistentRule`

### DIFF
--- a/spec/ameba/base_spec.cr
+++ b/spec/ameba/base_spec.cr
@@ -15,6 +15,7 @@ module Ameba::Rule
         Rule.rules.map(&.group_name).uniq!.reject!(&.empty?).sort.should eq %w[
           Ameba
           Documentation
+          Internal
           Layout
           Lint
           Metrics
--- a/spec/ameba/rule/internal/bad_directive_spec.cr
+++ b/spec/ameba/rule/internal/bad_directive_spec.cr
@@ -1,18 +1,18 @@
 require "../../../spec_helper"
 
-module Ameba::Rule::Lint
+module Ameba::Rule::Internal
   describe BadDirective do
     subject = BadDirective.new
 
     it "does not report if action is correct" do
       expect_no_issues subject, <<-CRYSTAL
-        # ameba:disable Lint/BadDirective
+        # ameba:disable Internal/BadDirective
         CRYSTAL
     end
 
     it "reports if there is incorrect action" do
       expect_issue subject, <<-CRYSTAL
-        # ameba:foo Lint/BadDirective
+        # ameba:foo Internal/BadDirective
               # ^^^ error: Bad action in comment directive: `foo`. Possible values: `disable`, `enable`
         CRYSTAL
     end
--- a/spec/ameba/rule/internal/deprecated_rule_spec.cr
+++ b/spec/ameba/rule/internal/deprecated_rule_spec.cr
@@ -1,8 +1,8 @@
 require "../../../spec_helper"
 
-module Ameba::Rule::Lint
-  describe NonExistentRule do
-    subject = NonExistentRule.new
+module Ameba::Rule::Internal
+  describe DeprecatedRule do
+    subject = DeprecatedRule.new
 
     it "does not report if the rule name is correct" do
       expect_no_issues subject, <<-CRYSTAL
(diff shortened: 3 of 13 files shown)


Sija:
Any suggestions regarding the group name are welcome.
FTR, we can't use `Ameba` due to the Crystal namespace conflicts (avoiding it is technically possible but cumbersome)

straight-shoota:
Why do these rules need their own namespace?

Sija:
@straight-shoota They don't.

I happen to think it's a better organization to have them separated, as they're not related to the Crystal code in any way—unlike the rest of the rules in the `Lint` group, and elsewhere.

`Lint/BadDirective`, for example, is concerned only with Ameba directives; there are no _Crystal directives_, so having it within the `Lint` group, where the naming directly refers to Crystal language constructs, doesn't make much sense to me.

straight-shoota:
Ameba directives are part of the Crystal code. So I do see a strong relation there.

That doesn't mean they can't be in their own group, but the motivation is not as strong. Either way is fine probably. But the name `Internal` is not very intuitive.

FWIW Rubocop has directive-related rules in the Lint department.

Sija:
> Ameba directives are part of the Crystal code. So I do see a strong relation there.

Not really, no. Even though they use code comments as the primary medium, their format and terminology are completely domain-specific. Hence, the connection between the Crystal and Ameba domains is pretty weak in my mind.

> [...] But the name `Internal` is not very intuitive.

Agreed.

> FWIW Rubocop has directive-related rules in the Lint department.

While Rubocop is one source of inspiration, I don't agree with all of its choices.

veelenga:
@Sija do we really want to introduce a breaking change just to have 3 rules in a separate Internal group? I don't see many cases where this namespace would grow significantly.

Sija:
@veelenga Breaking change here is purely virtual. I still have to see cases where these rules are mentioned anywhere (whether inline or in the config file).

As to your 2nd point, what matters to me more than having 3 rules in a group (`Metrics` has just one, for example, `Documentation`—2 and `Layout`—3) is a clear distinction between Crystal-code related rules and Ameba ones.

nobodywasishere:
I agree that I'm not sure if this is the best idea, mostly from the perspective that we should avoid changing rule names as much as possible (alongside remove / changing config option names).

Sija:
> [...] mostly from the perspective that we should avoid changing rule names as much as possible (alongside remove / changing config option names).

This perspective doesn't hold water in this case—I've described the reasons I think so in my previous comment.

veelenga:
> As to your 2nd point, what matters to me more than having 3 rules in a group (`Metrics` has just one, for example, `Documentation`—2 and `Layout`—3) is a clear distinction between Crystal-code related rules and Ameba ones.

Yeah, what I meant is not the number of rules, but rather having groups that are future-proof. `Metrics`, `Documentation`, and `Layout` have a clear theme and can naturally accommodate additional rules in the future, whereas `Internal` feels forced here since it doesn't seem to represent a well-defined category in the same way.

Sija:
I'm not sure why you see the `Internal` group as forced and not future-proof? It's meant to consist of Ameba-specific rules, which might increase in number over time—see the newly added `Lint/DeprecatedRule` as an example.

veelenga:
> I'm not sure why you see the `Internal` group as forced and not future-proof? It's meant to consist of Ameba-specific rules, which might increase in number over time—see the newly added `Lint/DeprecatedRule` as an example.

Right. I agree there might be few more examples, but I don't expect it to increase because these rules do not add much value to the users' codebase, they just help to properly use Ameba, do smoother migrations etc. And I don't think that Ameba must focus on such. 

Well, I don't have strong "no". If you have a good vision here I am ok with that.

veelenga (approved):
LGTM

Sija:
> Right. I agree there might be few more examples, but I don't expect it to increase because these rules do not add much value to the users' codebase, they just help to properly use Ameba, do smoother migrations etc. And I don't think that Ameba must focus on such.

Quite possibly you might be right. I was just coming from the side of keeping the rules categorized, in which scenario the Ameba-related ones IMO didn't fit into the `Lint` category. I dunno, maybe that's my AuDHD kicking in xD 

Anyhow, I'm gonna leave this lie for some time until we've come to some mutually satisfactory solution.

PR #818, opened 2026-04-20
Change `Lint::NonExistentRule#since_version` to `0.13.0`

Sija:
I believe that's more sensible, since this rule was originally introduced in version `0.13.0` as a part of the `Lint/BadDirective` rule.

Followup to #809


veelenga (approved):
lgtm

PR #820, opened 2026-05-04
Make `AST::RedundantControlExpressionVisitor` report `if`/`unless` branches with no `else`

Sija:


spec/ameba/rule/style/redundant_return_spec.cr
```diff
@@ -39,7 +39,7 @@ module Ameba::Rule::Style
       expect_no_issues subject, <<-CRYSTAL
         def method(a)
           case a
-          when true then return true
+          when true  then return true
```
veelenga, line 42:
```suggestion
          when true then return true
```

Sija: That was an intentional change to align it with the condition beneath it.

veelenga: Don't we have a crystal formatter for this purpose? What's the reason of doing manual alignment?

Sija: No, as that's a string within the specs file. And the reason for this change is my OCD 😅

veelenga (approved):
lgtm

PR #823, opened 2026-05-04
Support Crystal v1.21

Sija:


src/ameba/tokenizer.cr
```diff
@@ -85,7 +85,11 @@ module Ameba
         # NOTE: Crystal::Token is a class and the lexer modifies @token in place,
         # so the assignment here is for clarity/consistency with run_delimiter_state,
         # not for correctness (the behavior is identical without the assignment).
-        token = lexer.next_string_array_token
+        token = {% if compare_versions(Crystal::VERSION, "1.21.0") < 0 %}
+                  lexer.next_string_array_token
+                {% else %}
+                  lexer.next_string_token(token.delimiter_state)
+                {% end %}
```
straight-shoota: **suggestion:** Better use a functional test then depend on the version number. Development builds before crystal-lang/crystal@a9cca955a835f6641c48e77bd852ca81b6fd4558 report as version 1.21.0 but `next_string_token` doesn't work correctly.
```suggestion
        token = if lexer.responds_to?(:next_string_array_token)
                  # Crystal < 1.21.0
                  lexer.next_string_array_token
                else
                  lexer.next_string_token(token.delimiter_state)
                end
```

Sija: I had it done exactly like this, but decided on a compile-time implementation since it's:

- more performant by design
- there's a `compare_versions` call in the code, which makes it easier to grep and remove it in the future, when the support for older Crystal versions is going to be dropped

> Development builds before [crystal-lang/crystal@a9cca95](https://github.com/crystal-lang/crystal/commit/a9cca955a835f6641c48e77bd852ca81b6fd4558) report as version 1.21.0 but `next_string_token` doesn't work correctly.

No, they don't. They report `1.21.0-dev`, which is lower - thus the CI (just Windows, for now) nightly is still failing. I'm planning to merge it once the Crystal 1.21.0 is released.

ysbaddaden: - `#responds_to?` is a comptime check: only the branch that works will be compiled (no performance impact).
- Check `compare_versions(Crystal::VERSION, "1.21.0-dev") >= 0` and that would fix nightlies now—why wait for 3 months?

Sija: > * `#responds_to?` is a comptime check: only the branch that works will be compiled (no performance impact).

Good to know 👍 

> * Check `compare_versions(Crystal::VERSION, "1.21.0-dev") >= 0` and that would fix nightlies now

Yeah, that's what I was planning to do.

> * why wait for 3 months?

Ouch, that's a tad long indeed. Somehow, I thought they were released more frequently.

straight-shoota: With a version comparison, _some_ development builds are always broken. Either the onces before [crystal-lang/crystal@a9cca95](https://github.com/crystal-lang/crystal/commit/a9cca955a835f6641c48e77bd852ca81b6fd4558) (or 1.21.0), or the ones after.

**Feature detection is the correct choice here.**

In other cases, when a new feature is introduced, we can safely continue using the old API until the new is available. But here we need to switch to the new API immediately because the old doesn't work anymore.

There is no difference in performance.
Feel free to add whatever comment to keep track of removing this compatibility code (e.g. `# FIXME: Remove this after XY` or `# this is similar to compare_versions(Crystal::VERSION, "1.21.0") < 0` for grepability.

Sija: Might be, even so, atm I still prefer the way it's being done right now, and I'm not convinced that the breaking changes in Crystal v1.21.0 are over yet.

straight-shoota: > I'm not convinced that the breaking changes in Crystal v1.21.0 are over yet.

I'm curious what does that have to do with this?

Sija: I meant that even with feature detection, there might be cases where using `compare_versions` would still be preferred.

straight-shoota: Sure. But for this particular change, `compare_versions` is insufficient while `responds_to?` is correct.

veelenga (approved):
lgtm

Issue #144, opened 2020-04-08
Incorrectly reporting shadowingOuterLocalVar within macro


Blacksmoke16:
```cr
annotation MyAnn; end

class Controller; end

class Foo
  def initialize
    {% for klass, c_idx in Controller.all_subclasses.reject &.abstract? %}
      {% methods = klass.methods.select { |m| m.annotation(MyAnn) } %}

      {% for m, m_idx in methods %}

        {% if d = m.annotation(MyAnn) %}
          {% d %}
        {% end %}
      {% end %}
    {% end %}
  end
end
```

On `0.12.0`
```
[W] Lint/ShadowingOuterLocalVar: Shadowing outer local variable `m`
> {% methods = klass.methods.select { |m| m.annotation(MyAnn) } %}
                                       ^
Finished in 5.07 milliseconds 

1 inspected, 1 failure.
```

On `0.11.0`:
```
Inspecting 1 file.

.

Finished in 805 microseconds 

1 inspected, 0 failures.
```

Looks like this specific case wasn't handled by aff723b682ce0f80ca9490c21901b851ff9b4faa

veelenga:
This is different and introduced in #135 (because it changes the order of how the scopes are hooked)

Reduced:

```crystal
a = [1,2,3].select { |m| m.even? }
                      ^
m = a.last
```

PS: might be different on `0.11.0` but both issues will be covered by the coming fix. 

Thanks for reporting.

PR #826, opened 2026-05-12
fix: do not flag block argument shadowing inside the initial assignment in `Lint/ShadowingOuterLocalVar`

### LINKED ISSUE #825: `Lint/ShadowingOuterLocalVar` complains about shadowing local variable in assignment to that variable
`Lint/ShadowingOuterLocalVar` complains about shadowing a local variable in an assignment to that variable.

```cr
x = foo { |x| }
         # ^ Lint/ShadowingOuterLocalVar: Shadowing outer local variable `x`
```

The method call (and thus, block) come first semantically. They go before the outer `x` is even declared.
So it's debatable whether this should be considered a violation.

The above code can be separated into a method call and an assignment. This is semantically equivalent to the above, but does _not_ violate `Lint/ShadowingOuterLocalVar` (because `x` not declared after the block).

```cr
%y = foo { |x| }
x = %y
```

Similar to #819 this depends on control flow. But it's more controversial, so happy to discuss this.

veelenga:
Closes #825
Relates to #144, https://github.com/crystal-ameba/ameba/commit/458c49273000a99798ac04acee50ef7755595935

### DIFF
--- a/spec/ameba/rule/lint/shadowing_outer_local_var_spec.cr
+++ b/spec/ameba/rule/lint/shadowing_outer_local_var_spec.cr
@@ -194,6 +194,41 @@ module Ameba::Rule::Lint
         CRYSTAL
     end
 
+    context "block argument inside the introducing assignment" do
+      it "does not report when block argument shares a name with the outer assignment target" do
+        expect_no_issues subject, <<-CRYSTAL
+          x = foo { |x| x + 1 }
+          unit = units.find! { |unit| unit.name == name }
+          CRYSTAL
+      end
+
+      it "does not report when the block is nested inside the assignment value" do
+        expect_no_issues subject, <<-CRYSTAL
+          x = if cond
+                1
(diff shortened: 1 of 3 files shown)


spec/ameba/rule/lint/shadowing_outer_local_var_spec.cr
```diff
@@ -194,6 +194,41 @@ module Ameba::Rule::Lint
         CRYSTAL
     end
 
+    context "block argument inside the introducing assignment" do
+      it "does not report when block argument shares a name with the outer assignment target" do
+        expect_no_issues subject, <<-CRYSTAL
+          x = foo { |x| x + 1 }
+          unit = units.find! { |unit| unit.name == name }
```
Sija: Suggestion: remove redundant call.

```suggestion
```

Sija (approved):

PR #827, opened 2026-05-16
Raise on invalid file path

### LINKED ISSUE #388: Providing an invalid file path should raise instead of silently ignoring it

Sija:
Resolves #787
Closes #388

Adds `--ignore-unmatched-paths` CLI flag to opt out of the new behaviour.

### Original description

> The intent of this PR is to raise if any of the provided inputs resolve to no files.
> 
> #### Before (silent success — misleading):
> ```
> # Typo in filename — ameba says "all good!" but checked nothing
> $ ameba typo.cr # => exits 0, no errors reported
> 
> # Non-existent directory — same problem
> $ ameba "src/nonexistent/**/*.cr" # => exits 0, no errors reported
> ```
> 
> #### After (clear error):
> ```
> $ ameba typo.cr
> # => raises: "No files found matching `typo.cr`"
> 
> $ ameba "src/nonexistent/**/*.cr"
> # => raises: "No files found matching `src/nonexistent/**/*.cr`"
> ```

### DIFF
--- a/spec/ameba/cli/cmd_spec.cr
+++ b/spec/ameba/cli/cmd_spec.cr
@@ -7,8 +7,26 @@ module Ameba::CLI
   describe "Cmd" do
     describe ".run" do
       it "runs ameba" do
-        r = CLI.run %w[-f silent file.cr]
+        r = CLI.run ["-f", "silent", "--only", "Lint/ComparisonToBoolean", __FILE__]
         r.should be_true
+      end
+
+      it "raises when a non-existent file is provided" do
+        expect_raises(Exception, "No files found matching `not_there.cr`") do
+          CLI.run %w[-f silent not_there.cr]
+        end
+      end
+
+      it "raises when a glob matches no files" do
+        expect_raises(Exception, "No files found matching") do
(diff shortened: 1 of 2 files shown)


straight-shoota:
FWIW: When running the command `ameba src/nonexistent/**/*.cr` in a shell, the shell typically already expands the glob pattern and errors if it doesn't match.

This patch only has an effect when ameba resolves the glob pattern instead of the shell.
The command `ameba 'src/nonexistent/**/*.cr'` would be better for the change description. The  quoting makes sure ameba receives the original pattern instead of its expansion.

veelenga (approved):
lgtm 👍🏻

Issue #765, opened 2026-01-19
Renamed rules report `Lint/BadDirective` and `Lint/UnneededDisableDirective`


Sija:
There were some rule renames between Ameba v1.6.4 and v1.7.0 (like `Performance/AnyInsteadOfEmpty` -> `Performance/AnyInsteadOfPresent`)

These trigger `Lint/BadDirective` and `Lint/UnneededDisableDirective` issues when linted with newer Ameba version. To avoid such a situation, we most likely need to implement some kind of deprecation support.

PR #829, opened 2026-05-21
Implement rule deprecation mechanism

Sija:
Refs #765

### DIFF
--- a/spec/ameba/rule/base_spec.cr
+++ b/spec/ameba/rule/base_spec.cr
@@ -8,6 +8,26 @@ module Ameba
       end
     end
 
+    describe ".deprecated?" do
+      it "returns false if the rule is not deprecated" do
+        DummyRule.deprecated?.should be_false
+      end
+
+      it "returns true if the rule is deprecated" do
+        DeprecatedRule.deprecated?.should be_true
+      end
+    end
+
+    describe ".deprecation_reason" do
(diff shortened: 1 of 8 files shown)


Sija:
Using `Deprecated` annotation seems neat, yet while this works, I'm not happy about the fact that deprecated class warnings show up when running specs or building the binary, and I'm not sure yet how to resolve this issue.

@veelenga @nobodywasishere @straight-shoota WDYT?

straight-shoota:
Yeah, it feels like this might be at the wrong layer. That's similar to the API docs for the rules, which don't serve as developer documentation, but user documentation.

The distinction here is that using these classes in the Ameba API is not (necessarily) deprecated. 
But using the rules in your Ameba configuration is.

Excluding deprecation warnings with `--exclude-warnings src/` would be too broad, and there's no fine-grained warning management.

I believe it would easier overall to not use `@[Deprecated]` annotation. What's the benefit of that anyway?
We could use a custom annotation (`@[Ameba::DeprecatedRule]`) which would achieve basically the same thing.

Sija:
> We could use a custom annotation (`@[Ameba::DeprecatedRule]`) which would achieve basically the same thing.

Good idea! I've updated the code with your suggestion.

Sija:
@veelenga @nobodywasishere 🏓

nobodywasishere (approved):

PR #832, opened 2026-05-25
`Lint/Formatting`: avoid `IndexError` when autocorrecting source with no trailing newline

### LINKED ISSUE #514: The range 0...N is outside the bounds of the source (`IndexError`)
Running below command (piping is just a convenience here, same goes for regular files):

```sh
echo -n "# foo" | bin/ameba --fix -
```

Results in `IndexError`:

```
The range 0...6 is outside the bounds of the source (IndexError)
  from src/ameba/source/rewriter.cr:131:7 in 'check_range_validity'
  from src/ameba/source/rewriter.cr:124:7 in 'combine:replacement'
  from src/ameba/source/rewriter.cr:75:7 in 'replace'
  from src/ameba/source/corrector.cr:18:7 in 'replace'
  from src/ameba/rule/lint/formatting.cr:53:9 in '->'
  from src/ameba/issue.cr:43:20 in 'correct'
  from src/ameba/source.cr:29:21 in 'correct?'
  from src/ameba/runner.cr:133:38 in 'run_source'
  from src/ameba/runner.cr:93:11 in '->'
  from /opt/homebrew/Cellar/crystal/1.14.0/share/crystal/src/fiber.cr:143:11 in 'run'
  from /opt/homebrew/Cellar/crystal/1.14.0/share/crystal/src/fiber.cr:95:34 in '->'
```

Tested with 51cb88e4fb9016981beb78726dff881c2c1dac4a

---

Happens only when `Lint/Formatting` and `Layout/TrailingBlankLines` issues are present at the same time.
Possibly related to #489

/cc @FnControlOption

SAY-5:
# Description

Fixes `Lint/Formatting` autocorrection crashing with `IndexError` on any source file that lacks a trailing newline.

Fixes #514

## Root cause

`Lint/Formatting` computed the replacement range end column as `source_lines.last.size + 1`. Crystal's `String#lines` strips trailing newlines, so for a source like `"# foo"` (5 chars, no trailing newline):

- `source_lines.last.size` = 5
- end column = 6
- `Corrector#replace` converts this to byte offset 5 then adds 1 more (for inclusive-to-exclusive conversion), yielding position 6
- `Rewriter#check_range_validity` raises `IndexError: The range 0...6 is outside the bounds of the source (0...5)`

For sources *with* a trailing newline the +1 happened to be harmless because `code.size` was one larger, but for sources without one it always overshot.

## Fix

Use the range-based `Corrector#replace(0...source_code.size, result)` overload, which directly expresses "replace the entire source" without any line/column arithmetic. This works correctly regardless of whether the source ends with a newline.

## Type of change

- [x] Bug fix (non-breaking change which fixes an issue)
- [x] Unit tests (added or updated unit-tests)

### DIFF
--- a/spec/ameba/rule/lint/formatting_spec.cr
+++ b/spec/ameba/rule/lint/formatting_spec.cr
@@ -30,6 +30,14 @@ module Ameba::Rule::Lint
         CRYSTAL
     end
 
+    it "does not raise IndexError when source has no trailing newline" do
+      source = Ameba::Source.new("# foo", "test.cr")
+      subject.test(source)
+      source.issues.empty?.should be_false
+      source.correct!
+      source.code.should eq "# foo\n"
+    end
+
     context "properties" do
       context "#fail_on_error" do
         it "passes on formatter errors by default" do
--- a/spec/ameba/rule/lint/formatting_spec.cr
+++ b/spec/ameba/rule/lint/formatting_spec.cr
@@ -30,12 +30,15 @@ module Ameba::Rule::Lint
         CRYSTAL
     end
 
-    it "does not raise IndexError when source has no trailing newline" do
-      source = Ameba::Source.new("# foo", "test.cr")
-      subject.test(source)
-      source.issues.empty?.should be_false
-      source.correct!
-      source.code.should eq "# foo\n"
+    it "successfully formats source with no trailing newline" do
+      source = expect_issue subject, <<-CRYSTAL
+        # foo
+        # ^{} error: Use built-in formatter to format this source
+        CRYSTAL
+
+      expect_correction source, <<-CRYSTAL
+        # foo\n
+        CRYSTAL
     end
 
     context "properties" do
--- a/src/ameba/rule/lint/formatting.cr
+++ b/src/ameba/rule/lint/formatting.cr
@@ -46,13 +46,8 @@ module Ameba::Rule::Lint
       result = Crystal.format(source_code, source.path)
       return if result == source_code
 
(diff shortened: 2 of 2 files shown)


spec/ameba/rule/lint/formatting_spec.cr
```diff
@@ -30,6 +30,14 @@ module Ameba::Rule::Lint
         CRYSTAL
     end
 
+    it "does not raise IndexError when source has no trailing newline" do
+      source = Ameba::Source.new("# foo", "test.cr")
+      subject.test(source)
+      source.issues.empty?.should be_false
+      source.correct!
+      source.code.should eq "# foo\n"
+    end
```
Sija:
```suggestion
    it "successfully formats source with no trailing newline" do
      source = expect_issue subject, <<-CRYSTAL
        # foo
        # ^{} error: Use built-in formatter to format this source
        CRYSTAL

      expect_correction source, <<-CRYSTAL
        # foo\n
        CRYSTAL
    end
```

Sija (changes requested):

src/ameba/rule/lint/formatting.cr
Sija:
```suggestion
source_code = source.code
return if source_code.empty?
```

Sija (approved):

SAY-5:
@Sija Oh! I thought you were talking about the work and not the description, then yes I have been taking help of AI to make PR descriptions, both thorough and elaborate so that it sticks to the template and I don't miss out on anything.

PR #833, opened 2026-05-27
Add `Style/TrailingCommaInMultilineArgument` rule

Sija:


Sija:
@straight-shoota WDYT about including this change/logic in the Crystal formatter?

straight-shoota:
The formatter handles this very inconsistently for different list types.
It already adds trailing commas for defs and (most) collection literals (crystal-lang/crystal#13169, crystal-lang/crystal#10916).

We should extend that to calls and other places where multi-line lists are separated by commas.

So perhaps it's not worth it do add this to ameba and instead let the formatter handle it.

Sija:
> So perhaps it's not worth it do add this to ameba and instead let the formatter handle it.

That's what I was thinking.

src/ameba/rule/lint/trailing_comma_in_multiline_argument.cr
```diff
@@ -0,0 +1,62 @@
+module Ameba::Rule::Lint
```
veelenga: Could you please elaborate why is this not in Style group? 

Lint for me looks like a the group catching the code smells, while the trailing command in multiline argument is just a pure style preference

Sija: You're right, that's an oversight on my part.

veelenga (approved):
lgtm

PR #837, opened 2026-06-01
Fixup `Style/CallParentheses` rule

Sija:


src/ameba/rule/style/call_parentheses.cr
```diff
@@ -114,41 +119,51 @@ module Ameba::Rule::Style
...
+      end_location ||= node
+
+      unless end_location.is_a?(Crystal::Location)
+        end_location = end_location.end_location
+      end
+
+      # TODO: verify whether it's rly needed
+      return unless end_location
+      return unless end_location.line_number.positive? &&
+                    end_location.column_number.positive?
```
straight-shoota: **thought:** This really seems unnecessary here.
If any location is invalid, we should fix that.

Line numbers can potentially be zero for a virtual location. But those shouldn't appear after just the parser stage.

veelenga (approved):
lgtm

PR #838, opened 2026-06-01
fix: avoid false positives across unconnected branches in `Lint/ShadowingOuterLocalVar`

### LINKED ISSUE #835: `Lint/ShadowingOuterLocalVar` consideres local variables from unconnected code branch
This is similar to #819: The shadowed variable is only declared in a code branch that's mutually exclusive with the branch where the block parameter appears.

```cr
if 1
  x = 1
  return
end

foo { |x| } # Lint/ShadowingOuterLocalVar: Shadowing outer local variable `x`
```

The same applies to `next`, `break`, `raise` and any other `NoReturn` call. The latter would require semantic analysis though. So keywords should be good enough for now. Maybe a special case for standalone `raise`.

veelenga:
Reworks `Lint/ShadowingOuterLocalVar` to decide shadowing through a forward reaching-definition analysis: a block/proc argument is flagged only when an outer assignment can actually reach the point where it is introduced. Branches that terminate (`return`/`next`/`break`/`raise`) no longer leak assignments past them.

closes https://github.com/crystal-ameba/ameba/issues/835

### DIFF
--- a/spec/ameba/ast/reaching_definition_analyzer_spec.cr
+++ b/spec/ameba/ast/reaching_definition_analyzer_spec.cr
@@ -140,6 +140,55 @@ module Ameba::AST
       end
     end
 
+    context "loops" do
+      it "reaches a block assigned earlier in the same iteration" do
+        reaches?(<<-CRYSTAL, "x").should be_true
+          while cond
+            x = 1
+            foo { |x| }
+          end
+          CRYSTAL
+      end
+
+      it "reaches a block when assigned before the loop" do
+        reaches?(<<-CRYSTAL, "x").should be_true
+          x = 1
+          while cond
+            foo { |x| }
+          end
+          CRYSTAL
+      end
+
+      # Crystal scoping is lexical: a variable assigned later in the body is not
+      # in scope at an earlier point, even though the loop's back edge would
+      # reach it at runtime.
+      it "does not reach a block that precedes the assignment in the body" do
+        reaches?(<<-CRYSTAL, "x").should be_false
+          while cond
+            foo { |x| }
+            x = 1
+          end
+          CRYSTAL
+      end
+
+      it "does not reach a block that precedes the assignment in an until body" do
+        reaches?(<<-CRYSTAL, "x").should be_false
+          until cond
+            foo { |x| }
+            x = 1
+          end
+          CRYSTAL
+      end
+
+      it "does not reach a block in a loop that never assigns the variable" do
+        reaches?(<<-CRYSTAL, "x").should be_false
+          while cond
+            foo { |x| }
+          end
+          CRYSTAL
+      end
+    end
+
     context "captured definitions" do
       it "reaches through a nested block from an outer assignment" do
         reaches?(<<-CRYSTAL, "x").should be_true
--- a/spec/ameba/ast/reaching_definition_analyzer_spec.cr
+++ b/spec/ameba/ast/reaching_definition_analyzer_spec.cr
@@ -2,8 +2,7 @@ require "../../spec_helper"
 
 private def scopes_for(code)
   rule = Ameba::ScopeRule.new
-  source = Ameba::Source.new(code)
-  Ameba::AST::ScopeVisitor.new(rule, source)
+  Ameba::AST::ScopeVisitor.new(rule, Ameba::Source.new(code))
   rule.scopes
 end
 
--- a/spec/ameba/ast/reaching_definition_analyzer_spec.cr
+++ b/spec/ameba/ast/reaching_definition_analyzer_spec.cr
@@ -1,0 +1,169 @@
+require "../../spec_helper"
+
+private def scopes_for(code)
+  rule = Ameba::ScopeRule.new
+  source = Ameba::Source.new(code)
+  Ameba::AST::ScopeVisitor.new(rule, source)
+  rule.scopes
+end
+
+private def reaches?(code, name)
+  inner = scopes_for(code).find! do |scope|
+    (scope.node.is_a?(Crystal::Block) || scope.node.is_a?(Crystal::ProcLiteral)) &&
+      scope.arguments.any?(&.name.== name)
+  end
+  outer = inner.outer_scope || raise "expected an outer scope"
+  outer.declared_at?(name, inner.node)
+end
+
+module Ameba::AST
+  describe ReachingDefinitionAnalyzer do
+    context "sequential flow" do
+      it "reaches a block after the assignment" do
+        reaches?(<<-CRYSTAL, "x").should be_true
+          x = 1
+          foo { |x| }
+          CRYSTAL
+      end
+
+      it "does not reach a block before the assignment" do
+        reaches?(<<-CRYSTAL, "x").should be_false
+          foo { |x| }
+          x = 1
+          CRYSTAL
+      end
+
+      it "does not reach a block nested in its own initializing assignment" do
+        reaches?(<<-CRYSTAL, "x").should be_false
+          x = foo { |x| }
+          CRYSTAL
+      end
+
+      it "reaches through a splat multi-assignment target" do
+        reaches?(<<-CRYSTAL, "x").should be_true
+          a, *x = [1, 2, 3]
+          foo { |x| }
+          CRYSTAL
+      end
+
+      it "reaches an assignment inside a short-circuit operator" do
+        reaches?(<<-CRYSTAL, "x").should be_true
+          done || (x = compute)
+          foo { |x| }
+          CRYSTAL
+      end
+    end
+
+    context "conditional flow" do
+      it "does not reach across mutually exclusive if/else branches" do
+        reaches?(<<-CRYSTAL, "x").should be_false
+          if cond
+            x = 1
+          else
+            foo { |x| }
+          end
+          CRYSTAL
+      end
+
+      it "reaches when assigned before the conditional" do
+        reaches?(<<-CRYSTAL, "x").should be_true
+          x = 1
+          if cond
+            foo { |x| }
+          end
+          CRYSTAL
+      end
+
+      it "reaches when assigned in the condition" do
+        reaches?(<<-CRYSTAL, "x").should be_true
+          if x = compute
+            foo { |x| }
+          end
+          CRYSTAL
+      end
+
+      it "may reach when assigned in only one prior branch" do
+        reaches?(<<-CRYSTAL, "x").should be_true
+          if cond
+            x = 1
+          end
+          foo { |x| }
+          CRYSTAL
+      end
+    end
+
+    context "terminating branches" do
+      it "does not reach past a branch that returns" do
+        reaches?(<<-CRYSTAL, "x").should be_false
+          def foo
+            if cond
+              x = 1
+              return
+            end
+            bar { |x| }
+          end
+          CRYSTAL
+      end
+
+      it "does not reach past a branch that raises" do
+        reaches?(<<-CRYSTAL, "x").should be_false
+          def foo
+            if cond
+              x = 1
+              raise "boom"
+            end
+            bar { |x| }
+          end
+          CRYSTAL
+      end
+
+      it "reaches when the terminating branch does not assign the variable" do
+        reaches?(<<-CRYSTAL, "x").should be_true
+          def foo
+            x = 1
+            return if cond
+            bar { |x| }
+          end
+          CRYSTAL
+      end
+
+      it "reaches a block nested inside a terminating branch" do
+        reaches?(<<-CRYSTAL, "x").should be_true
+          def foo
+            if cond
+              x = 1
+              bar { |x| }
+              return
+            end
+          end
+          CRYSTAL
+      end
+    end
+
+    context "captured definitions" do
+      it "reaches through a nested block from an outer assignment" do
+        reaches?(<<-CRYSTAL, "x").should be_true
+          x = 1
+          foo { bar { |x| } }
+          CRYSTAL
+      end
+
+      it "treats arguments as reaching definitions" do
+        reaches?(<<-CRYSTAL, "x").should be_true
+          def foo(x)
+            bar { |x| }
+          end
+          CRYSTAL
+      end
+
+      it "does not capture across a def boundary" do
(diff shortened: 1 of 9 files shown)


Sija:
@veelenga Is this PR intentionally left in the *draft* state?

veelenga:
@Sija Sorry, don't have time right now. Will return to it later this week.

If you would like to hop-in or help, it would be helpful for sure.

src/ameba/ast/reaching_definition_analyzer.cr
```diff
@@ -0,0 +1,202 @@
...
+    alias DefinedSet = Set(String)
+
+    @definitions = {} of UInt64 => DefinedSet
+    @inner_scope_nodes : Set(UInt64)
+
+    # Creates a new analyzer for *scope*. *entry* is the set of variable names
+    # already defined when the scope is entered (its arguments plus any
+    # captured outer definitions).
+    def initialize(@scope : Scope, @entry : DefinedSet)
+      @inner_scope_nodes = @scope.inner_scopes.map(&.node.object_id).to_set
```
Sija:
```suggestion
      @inner_scope_nodes = @scope.inner_scopes.to_set(&.node.object_id)
```

src/ameba/ast/reaching_definition_analyzer.cr
```diff
@@ -0,0 +1,202 @@
...
+      end
+      defined
+    end
+
+    private def transfer(node : Crystal::UninitializedVar, defined : DefinedSet) : DefinedSet
+      define(node.var, defined)
+    end
+
+    private def transfer(node : Crystal::TypeDeclaration, defined : DefinedSet) : DefinedSet
+      node.value.try { |value| defined = propagate(value, defined) }
```
Sija:
```suggestion
      node.value.try do |value|
        defined = propagate(value, defined)
      end
```

src/ameba/ast/reaching_definition_analyzer.cr
```diff
@@ -0,0 +1,202 @@
...
+      define(node.var, defined)
+    end
+
+    private def transfer(node : Crystal::Call, defined : DefinedSet) : DefinedSet
+      node.obj.try { |obj| defined = propagate(obj, defined) }
```
Sija:
```suggestion
      node.obj.try do |obj|
        defined = propagate(obj, defined)
      end
```

src/ameba/ast/reaching_definition_analyzer.cr
```diff
@@ -0,0 +1,202 @@
...
+      node.args.each do |arg|
+        defined = propagate(arg, defined)
+      end
+      node.named_args.try &.each do |named_arg|
+        defined = propagate(named_arg.value, defined)
+      end
+      node.block_arg.try { |arg| defined = propagate(arg, defined) }
+      node.block.try { |block| defined = propagate(block, defined) }
```
Sija:
```suggestion
      node.block_arg.try do |arg|
        defined = propagate(arg, defined)
      end
      node.block.try do |block|
        defined = propagate(block, defined)
      end
```

src/ameba/ast/scope.cr
```diff
@@ -110,6 +112,37 @@ module Ameba::AST
...
 
+    # Returns `true` if a local variable named *name* has a definition that
+    # reaches the program point where *node* (an inner scope's node) is
+    # introduced.
+    #
+    # ```
+    # scope.declared_at?("foo", block_node)
+    # ```
+    def declared_at?(name : String, node) : Bool
+      reaching_definitions[node.object_id]?.try(&.includes?(name)) || false
```
Sija:
```suggestion
      !!reaching_definitions[node.object_id]?.try(&.includes?(name))
```

src/ameba/ast/scope.cr
```diff
@@ -110,6 +112,37 @@ module Ameba::AST
...
+    protected def reaching_definitions : Hash(UInt64, Set(String))
+      @reaching_definitions ||=
+        ReachingDefinitionAnalyzer.new(self, entry_definitions).inner_scope_definitions
+    end
+
+    # The set of variable names already defined when this scope is entered:
+    # its own arguments plus any captured definitions from the outer scope.
+    protected def entry_definitions : Set(String)
+      defined = Set(String).new
+      arguments.each { |argument| defined << argument.name }
```
Sija:
```suggestion
      defined = arguments.to_set(&.name)
```

src/ameba/ast/scope.cr
```diff
@@ -110,6 +112,37 @@ module Ameba::AST
...
+
+      if inherited? && (outer = outer_scope)
+        outer.reaching_definitions[node.object_id]?.try { |defs| defined.concat(defs) }
```
Sija:
```suggestion
        outer.reaching_definitions[node.object_id]?.try do |defs|
          defined.concat(defs)
        end
```

src/ameba/ast/reaching_definition_analyzer.cr
```diff
@@ -0,0 +1,202 @@
...
+
+    {% for type in INNER_SCOPE_NODES %}
+      private def transfer(node : Crystal::{{ type.id }}, defined : DefinedSet) : DefinedSet
+        defined
+      end
+    {% end %}
+
+    private def transfer(node : Crystal::ASTNode, defined : DefinedSet) : DefinedSet
+      children = [] of Crystal::ASTNode
+      node.accept_children(ChildCollector.new(children))
```
Sija, line 153:
```suggestion
      node.accept_children(ChildCollector.new(children))

```

src/ameba/ast/scope.cr
```diff
@@ -110,6 +112,37 @@ module Ameba::AST
...
+    def declared_at?(name : String, node) : Bool
+      reaching_definitions[node.object_id]?.try(&.includes?(name)) || false
+    end
+
+    # Reaching definitions snapshotted at each inner scope's node. Computed
+    # lazily and memoized, since several inner scopes share the same outer one.
+    protected def reaching_definitions : Hash(UInt64, Set(String))
+      @reaching_definitions ||=
+        ReachingDefinitionAnalyzer.new(self, entry_definitions).inner_scope_definitions
+    end
```
Sija:
```suggestion
    protected getter reaching_definitions : Hash(UInt64, Set(String)) do
      ReachingDefinitionAnalyzer.new(self, entry_definitions).inner_scope_definitions
    end
```

Sija:
Sure! What's left to be done?

veelenga:
### Tested on the Crystal compiler repo

To make sure this change does not add false positives, I ran the rule on [`crystal-lang/crystal`](https://github.com/crystal-lang/crystal) repo (`src` + `spec`). It reports **115** shadowed block/proc parameters.

I checked every one of the 115 against how Crystal really scopes local variables:

- A variable is in scope only from its **first textual assignment** and onward (scoping is lexical, top to bottom).
- A loop does not change this: a variable assigned **below** a block is not in scope **at** that block, even inside the same loop (the back edge does not count).
- Block parameters do not leak into the outer scope.

**Result: all 115 are real shadows — 0 false positives.** Most are block params that reuse a method parameter name; the rest reuse a local assigned earlier on a path that reaches the block (including assignments inside `if`/`while` bodies and conditions, which Crystal keeps in scope afterwards).

I also added specs for the loop case, so a variable assigned **below** a block in a loop is **not** reported.

straight-shoota (approved):

spec/ameba/ast/reaching_definition_analyzer_spec.cr
```diff
@@ -0,0 +1,218 @@
+require "../../spec_helper"
+
+private def scopes_for(code)
+  rule = Ameba::ScopeRule.new
+  source = Ameba::Source.new(code)
+  Ameba::AST::ScopeVisitor.new(rule, source)
```
Sija:
```suggestion
  Ameba::AST::ScopeVisitor.new(rule, Ameba::Source.new(code))
```

src/ameba/ast/dataflow.cr
```diff
@@ -0,0 +1,47 @@
+module Ameba::AST
+  # Building blocks for dataflow analysis: the AST node groupings used to
+  # generate type-specific handlers, body extraction, and immediate-child
+  # collection.
+  module Dataflow
+    BRANCH_NODES      = %w[If Unless]
+    LOOP_NODES        = %w[While Until]
+    CASE_NODES        = %w[Case Select]
+    INNER_SCOPE_NODES = %w[
```
Sija:
```suggestion
    private BRANCH_NODES      = %w[If Unless]
    private LOOP_NODES        = %w[While Until]
    private CASE_NODES        = %w[Case Select]
    private INNER_SCOPE_NODES = %w[
```

src/ameba/ast/dataflow.cr
```diff
@@ -0,0 +1,47 @@
...
+      when Crystal::LibDef            then node.body
+      when Crystal::CStructOrUnionDef then node.body
+      when Crystal::Assign            then node.value
+      when Crystal::OpAssign          then node.value
+      when Crystal::ProcLiteral       then node.def.body
+      when Crystal::EnumDef           then Crystal::Expressions.from(node.members)
+      when Crystal::TypeOf            then Crystal::Expressions.from(node.expressions)
+      when Crystal::Expressions       then node
+      else                                 node
+      end
```
Sija, line 33:
```suggestion
      case node
      when Crystal::Def,
           Crystal::FunDef,
           Crystal::Block,
           Crystal::ClassDef,
           Crystal::ModuleDef,
           Crystal::LibDef,
           Crystal::CStructOrUnionDef
        node.body
      when Crystal::Assign, Crystal::OpAssign
        node.value
      when Crystal::ProcLiteral then node.def.body
      when Crystal::EnumDef     then Crystal::Expressions.from(node.members)
      when Crystal::TypeOf      then Crystal::Expressions.from(node.expressions)
      else
        node
      end
```

src/ameba/ast/reaching_definition_analyzer.cr
```diff
@@ -0,0 +1,203 @@
...
+      defined
+    end
+
+    private def define(target, defined : DefinedSet) : DefinedSet
+      target = target.exp if target.is_a?(Crystal::Splat)
+      return defined unless target.is_a?(Crystal::Var)
+      return defined if defined.includes?(target.name)
+
+      defined.dup << target.name
+    end
```
Sija:
```suggestion
    private def define(target : Crystal::Var, defined : DefinedSet) : DefinedSet
      return defined if target.name.in?(defined)
      defined.dup << target.name
    end

    private def define(target : Crystal::Splat, defined : DefinedSet) : DefinedSet
      define(target.exp, defined)
    end

    private def define(target, defined : DefinedSet) : DefinedSet
      defined
    end
```

src/ameba/ast/reaching_definition_analyzer.cr
veelenga: Can you elaborate on the benefit?

Sija: Sorry, I should have written that: it's readability.

veelenga: Well, "readability" argument is always subjective to me. The original version looks more readable for me, so should I keep it or follow you readability preferences?

Sija: I'd prefer the version I've suggested, although I'm not gonna die on that hill.

src/ameba/ast/reaching_definition_analyzer.cr
veelenga: Fixed in https://github.com/crystal-ameba/ameba/pull/838/changes/1ff698907a3fe822e804543b7721f6ae5f4b72f0

src/ameba/ast/scope.cr
veelenga: Fixed in https://github.com/crystal-ameba/ameba/pull/838/changes/1ff698907a3fe822e804543b7721f6ae5f4b72f0

src/ameba/ast/scope.cr
veelenga: Fixed in https://github.com/crystal-ameba/ameba/pull/838/changes/1ff698907a3fe822e804543b7721f6ae5f4b72f0

spec/ameba/ast/reaching_definition_analyzer_spec.cr
veelenga: Fixed in https://github.com/crystal-ameba/ameba/pull/838/changes/1ff698907a3fe822e804543b7721f6ae5f4b72f0

src/ameba/ast/dataflow.cr
veelenga: Fixed in https://github.com/crystal-ameba/ameba/pull/838/changes/1ff698907a3fe822e804543b7721f6ae5f4b72f0

src/ameba/ast/dataflow.cr
veelenga: Fixed in https://github.com/crystal-ameba/ameba/pull/838/changes/1ff698907a3fe822e804543b7721f6ae5f4b72f0

src/ameba/ast/reaching_definition_analyzer.cr
veelenga: Fixed in https://github.com/crystal-ameba/ameba/pull/838/changes/1ff698907a3fe822e804543b7721f6ae5f4b72f0

src/ameba/ast/scope.cr
veelenga: There is a performance benefit in keeping the memoization. According to my tests it is 3x difference when there are 50+ nested blocks

src/ameba/ast/reaching_definition_analyzer.cr
veelenga: Fixed

src/ameba/ast/scope.cr
Sija: The refactored version keeps the memoization; it just uses the block version of the `getter` to achieve it.

veelenga: Sorry, missed that. Fixed

Sija (approved):

PR #840, opened 2026-06-03
Silently ignore unknown config attributes, instead of failing

### LINKED ISSUE #836: Ameba master breaks on `ExcludeTypeDeclarations` in `Lint/UslessAssign` config
2d4bcb94160223ac5aef053efad7297b37c6e1ba removed the `ExcludeTypeDeclarations` config for `Lint/UslessAssign` because it's no longer necessary.
But as a result, Ameba master (with that change) fails to parse existing configurations using this config option.
That's a configuration compatibility issue because we cannot have a `.ameba.yml` configuration that works across versions:
* Ameba 1.6 requires this option to avoid false positives
* Ameba 1.7 rejects the option

I'd suggest allowing the option and ignoring it. Or maybe report the config option as a violation in Ameba 1.7.

Sija:
Resolves #836

### DIFF
--- a/spec/ameba/rule/base_spec.cr
+++ b/spec/ameba/rule/base_spec.cr
@@ -2,6 +2,12 @@ require "../../spec_helper"
 
 module Ameba
   describe Rule::Base do
+    describe ".from_yaml" do
+      it "ignores unknown attributes" do
+        DummyRule.from_yaml("{ not: there }").should be_a(DummyRule)
+      end
+    end
+
     describe "#catch" do
       it "accepts and returns source" do
         source = Source.new
--- a/src/ameba/config/rule_config.cr
+++ b/src/ameba/config/rule_config.cr
@@ -209,7 +209,6 @@ class Ameba::Config
 
       macro inherited
         include YAML::Serializable
-        include YAML::Serializable::Strict
 
         def self.new(config = nil)
           if (raw = config.try &.raw).is_a?(Hash)


nobodywasishere (approved):

straight-shoota:
This feels like the other extreme: From disallowing anything unknown, we're going to allow anything. I suppose that's an improvement for practicality, but neither is really great.
This way you can end up with invalid configuration that doesn't do anything and is potentially confusing.

A bit of a middle ground would be to ignore _known_ attributes that were allowed in earlier versions, maybe warn about them at some point. And always error or warn on entirely unknown attributes.
Or something along that way. Might be worth looking into how other configuration schemas evolve with backward-compatibility (and potentially forward-compatibility).

Sija:
> This feels like the other extreme: From disallowing anything unknown, we're going to allow anything.

Correct. It's also more aligned with the current behaviour of ignoring unknown rule entries in the config file.

> I suppose that's an improvement for practicality, but neither is really great.

The practical difference you've mentioned is the key.

> This way you can end up with invalid configuration that doesn't do anything and is potentially confusing.

IMO it's better than failing (e.g., running an older Ameba version with the newer config). Unknown/obsolete options could be marked as issues by the rule (TBD) that would lint the `.ameba.yml` configuration file.

> A bit of a middle ground would be to ignore _known_ attributes that were allowed in earlier versions, maybe warn about them at some point. 

In theory, but we don't have those capabilities now, and we likely won't have them in the future, since this level of granularity is not sth I see as practical, at least now.

> And always error or warn on entirely unknown attributes. Or something along that way.

We could, yet we don't have any established notion of _warnings_.

PR #841, opened 2026-06-04
Always preserve excluded files in config + add `--ignore-config-file` CLI flag

### LINKED ISSUE #396: `--only` CLI flag appears to ignore excluded files within config
In my `.ameba.yml` I have this file excluded for the CyclomaticComplexity rule.

```yml
Metrics/CyclomaticComplexity:
  Excluded:
  - src/invidious/videos/parser.cr
```
But when running ameba with the `--only` option
`./lib/ameba/bin/ameba --only Metrics/CyclomaticComplexity`

It seems like Ameba still complains about the issue
```
src/invidious/videos/parser.cr:157:5
[W] Metrics/CyclomaticComplexity: Cyclomatic complexity too high [58/10]
> def parse_video_info(video_id : String, player_response : Hash(String, JSON::Any)) : Hash(String, JSON::Any)
```

Ameba is able to ignore the file fine when running without `--only` 

It makes sense for options like `--all` to ignore rules but I don't think the same applies for `--only`. Is this behavior intended?

Sija:
Resolves #396

### DIFF
--- a/src/ameba/config.cr
+++ b/src/ameba/config.cr
@@ -172,13 +172,13 @@ class Ameba::Config
   # config = Ameba::Config.load
   # config.update_rule "MyRuleName", enabled: false
   # ```
-  def update_rule(name, enabled = true, excluded = nil)
+  def update_rule(name, *, enabled, excluded = nil)
     rule = @rules.find(&.name.==(name))
     raise ArgumentError.new("Rule `#{name}` does not exist") unless rule
 
+    rule.enabled = enabled
+    rule.excluded = excluded.to_set if excluded
     rule
-      .tap(&.enabled = enabled)
-      .tap(&.excluded = excluded.try &.to_set)
   end
 
   # Updates rules properties.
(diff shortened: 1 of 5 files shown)


src/ameba/config.cr
```diff
@@ -172,13 +172,13 @@ class Ameba::Config
...
-  def update_rule(name, enabled = true, excluded = nil)
+  def update_rule(name, *, enabled, excluded = nil)
     rule = @rules.find(&.name.==(name))
     raise ArgumentError.new("Rule `#{name}` does not exist") unless rule
 
+    rule.enabled = enabled
+    rule.excluded = excluded.to_set if excluded
     rule
-      .tap(&.enabled = enabled)
-      .tap(&.excluded = excluded.try &.to_set)
```
veelenga, line 181: can we cover by test the actual fix of https://github.com/crystal-ameba/ameba/issues/396?

straight-shoota:
I'm wondering if we really need this additional flag? It doesn't seem like a very common use case. And I presume `--config /dev/null`/`--config NUL` should work just as well.

veelenga (approved):
lgtm

veelenga:
> And I presume `--config /dev/null`/`--config NUL` should work just as well.

It works, but very unintuitive. Having an explicit flag feels like a proper/officially supported way to skip the config.

straight-shoota:
I've never seen a similar option in any similar tool using a config file. So it's a bit surprising.

If we really need it, I'd suggest calling the flag `--no-config`: that's shorter and matches with well `--config`.

PR #846, opened 2026-06-08
Add `Performance/AnyInsteadOfEmpty` rule back (as deprecated)

### LINKED ISSUE #765: Renamed rules report `Lint/BadDirective` and `Lint/UnneededDisableDirective`

Sija:
Resolves #765


veelenga (approved):
lgtm

PR #857, opened 2026-06-17
Sort doc version index upon addition

Sija:
ATM, the doc [index](https://github.com/crystal-ameba/ameba/blob/gh-pages/versions.json) is not ordered at all, and added versions are pushed to the end of the array.
This PR changes the behaviour, so:

1. Unreleased versions will be displayed on top
2. Released versions will follow in descending order


veelenga (approved):
lgtm

PR #864, opened 2026-06-19
Raise a compile-time error on invalid build version

Sija:
Otherwise, passing an invalid version would result in a silently broken binary.

### DIFF
--- a/src/ameba/version.cr
+++ b/src/ameba/version.cr
@@ -12,8 +12,13 @@ module Ameba
         env("AMEBA_BUILD_VERSION") ||
           read_file?("#{__DIR__}/../../VERSION") ||
           `shards version "#{__DIR__}"`.stringify
+      version = version.chomp
+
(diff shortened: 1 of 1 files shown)


src/ameba/version.cr
```diff
@@ -12,8 +12,13 @@ module Ameba
         env("AMEBA_BUILD_VERSION") ||
           read_file?("#{__DIR__}/../../VERSION") ||
           `shards version "#{__DIR__}"`.stringify
+      version = version.chomp
+
+      unless version =~ /\A\d+\.\d+\.\d+(?:-[0-9a-z.-]+)?(?:\+[0-9a-z.-]+)?\z/i
+        raise "Not a semantic version: #{version}"
+      end
```
straight-shoota, line 19: **suggestion:** We might consider using `compare_versions` which should also ensure the version is valid.
```suggestion
      # validate `version`
      compare_versions(version, version)
```

Semantics might slightly differ from the regex.

Sija: I was thinking of that, yet it then errors out within the `compare_versions` macro making it impossible to use a custom exception message.

nobodywasishere (approved):

PR #866, opened 2026-06-23
Crystal 1.21 support

### LINKED ISSUE #887: Issue with `preview_mt` and latest version of crystal
Latest Crystal version removed the `-Dpreview_mt` as execution context feature got green-lighted.

So, when adding ameba in the project now, the  shard install will fails:

```
Failed postinstall of ameba on shards build -Dpreview_mt:
I: Dependencies are satisfied
I: Building: ameba
E: Error target ameba failed to compile:
The 'preview_mt' flag is deprecated. Resize the default execution context, or start additional contexts instead.
```

Is there any way to remove `-Dpreview_mt` without me having to fork the project ?

Sija:
- Drops usage of deprecated `-Dpreview_mt` compiler option
- Introduces default parallelism by resizing the default execution context to the logical number of CPU cores—while still respecting the `CRYSTAL_WORKERS` env variable value

> [!NOTE]
> While discouraged, it still supports building with the `-Dpreview_mt` flag _(for the time being)_

### DIFF
--- a/.github/workflows/docker-image.yml
+++ b/.github/workflows/docker-image.yml
@@ -79,7 +79,7 @@ jobs:
           cache-from: type=gha
           cache-to: type=gha,mode=max
           build-args: |
-            CRFLAGS=-Dpreview_mt --release
+            CRFLAGS=--release
           platforms: |
             linux/amd64
             linux/arm64
--- a/README.md
+++ b/README.md
@@ -20,7 +20,6 @@
   - [Autocorrection](#autocorrection)
   - [Explain issues](#explain-issues)
   - [Describe rules](#describe-rules)
-  - [Run in parallel](#run-in-parallel)
 - [Installation](#installation)
   - [As a project dependency](#as-a-project-dependency)
   - [OS X](#os-x)
@@ -105,17 +104,6 @@ You can use `--describe` flag to get a detailed description of a rule:
 $ ameba --describe Lint/UselessAssign
 ```
 
-### Run in parallel
-
-Some quick benchmark results measured while running Ameba on Crystal repo:
-
-```sh
-$ CRYSTAL_WORKERS=1 ameba #=> 29.11 seconds
-$ CRYSTAL_WORKERS=2 ameba #=> 19.49 seconds
-$ CRYSTAL_WORKERS=4 ameba #=> 13.48 seconds
-$ CRYSTAL_WORKERS=8 ameba #=> 10.14 seconds
-```
-
 ## Installation
 
 ### As a project dependency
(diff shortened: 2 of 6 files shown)


.github/workflows/docker-image.yml
```diff
@@ -79,7 +79,7 @@ jobs:
           cache-from: type=gha
           cache-to: type=gha,mode=max
           build-args: |
-            CRFLAGS=-Dpreview_mt --release
+            CRFLAGS=--release
```
veelenga, line 82: could you please elaborate or describe in PR description on "why" ? Is the flag removed from Crystal? Should we bump the version in shard.yml? 

https://github.com/crystal-ameba/ameba/blob/fcbd5703c50b47a5f280a11ab1c3416f5f3a9ea9/shard.yml#L14

Sija: > The 'preview_mt' flag is deprecated. Resize the default execution context, or start additional contexts instead.

This flag has been deprecated since Crystal 1.21.0. For the time being, though, we still support building with it, so there's no need to bump the minimum Crystal version.

README.md
```diff
@@ -105,17 +104,6 @@ You can use `--describe` flag to get a detailed description of a rule:
 $ ameba --describe Lint/UselessAssign
 ```
 
-### Run in parallel
-
-Some quick benchmark results measured while running Ameba on Crystal repo:
-
-```sh
-$ CRYSTAL_WORKERS=1 ameba #=> 29.11 seconds
```
veelenga, line 113: CRYSTAL_WORKERS is still a [valid env var](https://github.com/crystal-lang/crystal/blob/cf9298818b8531b52c7f70e6a74f9ab23645d534/src/fiber/execution_context.cr#L135) in Crystal 1.21 and that was the only place where the ameba parallelism was documented.  

Should we keep it documented (but maybe change the wording) ?

Sija: I intentionally removed this section, since now we do it automatically (Crystal 1.21.0+). Also, from what I've seen in the wild, no one is using it, so there's no downside to getting rid of this section altogether.

veelenga: Still see no reason to drop it completely instead of updating it. The feature didn't go away, `CRYSTAL_WORKERS` still works, since `ExecutionContext.default_workers_count` reads it first and only falls back to the CPU count. Only the benchmark numbers got stale.

And this was the only place the parallelism was mentioned at all. The feature itself is not simple, taking into account fibers per source, channels, mutex locks spread across the project. So a little hint in the README is valuable for newcomers and contributors IMO

straight-shoota: `CRYSTAL_WORKERS` is not used automatically. You have to explicitly call `ExecutionContext.default_workers_count`. I don't think ameba does that.

Sija: @straight-shoota This PR introduces such a call via https://github.com/crystal-ameba/ameba/commit/e0a0d4849896b2c3db504c411a4a83f0643da4a3

Sija: @veelenga I agree it would be nice to have it mentioned somewhere, though the README isn't the right place for it, as IMO it should be geared mainly towards general use cases rather than fine-tuning. Especially in the case of a setting that's ~never~ rarely used (fwiw, I've never used it tbh, maybe outside of testing)

veelenga: Since there is no other place to mention it right now, what is your recommendation? Just remove and forget?

Sija: Move it into the wiki, perhaps?

veelenga (approved):
lgtm

README.md
Sija: https://github.com/crystal-ameba/ameba/wiki#parallelism

PR #871, opened 2026-07-13
Turn `Rule::Base#documentation_url` into an `URI`

Sija:


straight-shoota:
What's the benefit of this? For the use cases in ameba, it doesn't seem to make a huge difference whether the url is typed as `URI` or `String`, and the latter is easier to handle.

Sija:
I agree, was waiting for someone to point that out xD

PR #873, opened 2026-07-13
Rename rule `Lint/DuplicatedRequire` -> `Lint/DuplicateRequire`

Sija:


nobodywasishere (approved):
Wish we had a way to still alias the old name

Issue #592, opened 2025-03-09
Lint SQL heredocs


nobodywasishere:
This may be outside the scope of this repo and more suited towards an ameba plugin. It's very common to write SQL queries in Crystal using heredocs, with one issue being that the query needs to be tested at runtime to verify its functionality. It would be useful to integrate with a tool like SQLFluff to provide lints for these heredocs. 

```cr
<<-SQL
  SELECT * FROM users;
# ^^^^^^^^^^^^^^^^^^^ error: [AM04] ambiguous.column_count: Query produces an unknown number of result columns.
  SQL
```

I have a WIP implementation on [this branch](https://github.com/nobodywasishere/ameba/tree/nobody/sql-lint), but running SQLFluff via subprocess seems to be very slow (~0.25-0.4 seconds per heredoc), probably due to Python startup times. Open to using suitable alternatives.

nobodywasishere:
sqruff (https://github.com/quarylabs/sqruff) may also work / be faster

Sija:
Sounds like a great idea, yet the execution times are troubling. If they could be reduced I'd be willing to include such rule in the codebase.

nobodywasishere:
I have a few ideas for how to get these down. One of which is using `sqruff` instead of `sqlfluff` (rust vs python, part of the time is needed for starting the python interpreter). 

Another was to pull all the SQL heredocs first, write them to temp files, then run sqlfluff once over all the files at once. Don't know if this would end up being faster though.

And yet another was to create a custom CLI for sqlfluff that could be communicated with back and forth, instead of doing one-offs. This would be accomplished using a custom python script that would be a part of this codebase, which isn't ideal, running in a single subprocess for the extent of the run. Should be the fastest but will require python / the sqlfluff library to be accessible (which _should_ be the case is sqlfluff is installed but we would have to see).

Thoughts?

straight-shoota:
It might be fine to focus on SQL as a common guest language.
However, I think it might be worth to approach this more generically and consider support for different languages as well. There are many use cases for embedded source code other than SQL, such as YAML, XML, Markdown, GraphQL etc.

With that in mind, it should be good to have some flexibility for the linter tools to use for each language. Ameba should perhaps just extract the sources and pass them on to the target tool for each language, extracted from a configurable lookup table.

[...]

PR #875, opened 2026-07-18
feat: security rule group

veelenga:
This PR adds a new `Security` rule group — a first step towards making Ameba useful as a security scanner for Crystal.

## Added rules

| Rule | Detects | CWE |
|---|---|---|
| `Security/CommandInjection` | Shell commands (`system`, backticks, `%x`) built from interpolated strings | [CWE-78](https://cwe.mitre.org/data/definitions/78.html) |
| `Security/SqlInjection` | SQL statements built from interpolated strings (includes heredocs, see #592) | [CWE-89](https://cwe.mitre.org/data/definitions/89.html) |
| `Security/WeakCrypto` | `Digest::MD5`, `Digest::SHA1`, `OpenSSL::Digest.new("MD5"/"SHA1")` | [CWE-327](https://cwe.mitre.org/data/definitions/327.html) |
| `Security/InsecureRandom` | `Random` used to generate tokens/secrets instead of `Random::Secure` | [CWE-338](https://cwe.mitre.org/data/definitions/338.html) |
| `Security/NoTlsVerify` | `OpenSSL::SSL::VerifyMode::NONE` | [CWE-295](https://cwe.mitre.org/data/definitions/295.html) |
| `Security/TimingAttack` | `==` / `!=` comparison of digests instead of `Crypto::Subtle.constant_time_compare` | [CWE-208](https://cwe.mitre.org/data/definitions/208.html) |
| `Security/HardcodedSecret` | Secrets in string literals: secret-named targets and well-known token formats (AWS, GitHub, Stripe, ...) | [CWE-798](https://cwe.mitre.org/data/definitions/798.html) |

## Decisions

These follow what mature tools do (Brakeman, bandit, gosec, Semgrep, CodeQL):

- **In core, not in [ameba-security](https://github.com/crystal-ameba/ameba-security).** I considered a separate shard, but all these rules only match stdlib APIs, so they belong in Ameba itself — every install gets them. Having a `Security/` group here and an `ameba-security` shard at the same time would be confusing, so that repo stays empty.
- **Extensions are organized by target library, not by category.** Framework- and shard-specific rules (Kemal, crystal-db, ...) should live in extensions like `ameba-kemal` later — the `rubocop-rails` model. Extension rules can still define themselves inside the `Security` group, so `--only Security` and group config keep working across core and extensions.
- **Precision over recall.** A noisy security rule gets disabled as a whole group. Rules skip static interpolations, recognized sanitizers (`Process.quote`, numeric casts) and placeholder values. Secret names match whole words only (`api_key` yes, `compass` no).
- **Confidence on injection rules.** Each finding gets `High` / `Medium` / `Low` confidence based on the interpolated expression (direct input read / variable / other). `MinConfidence` in the config filters findings. This lives inside the `Security` group only — core `Issue` is untouched.
- **`Error` severity by default**, like other groups have their own default.
- **Spec files are skipped.** Test fixtures are full of fake secrets and shell snippets.
- **No autocorrection.** Security fixes change behavior, so rules explain instead of rewriting.
- `Crystal::Path` was added to the visitable nodes so rules can inspect constant references (`VerifyMode::NONE`, `Digest::MD5`). Other rules are not affected.

## Notes

Happy to adjust: split this into smaller PRs, rename rules, change defaults, or ship the group as opt-in for the first release cycle.

### DIFF
--- a/spec/ameba/base_spec.cr
+++ b/spec/ameba/base_spec.cr
@@ -20,6 +20,7 @@ module Ameba::Rule
           Metrics
           Naming
           Performance
+          Security
           Style
           Typing
         ]
--- a/spec/ameba/rule/security/command_injection_spec.cr
+++ b/spec/ameba/rule/security/command_injection_spec.cr
@@ -18,6 +18,27 @@ module Ameba::Rule::Security
         CRYSTAL
     end
 
+    it "passes when dynamic parts are shell-quoted or safely cast" do
+      expect_no_issues subject, <<-'CRYSTAL'
+        system("ls -l #{Process.quote(path)}")
+        system("kill #{pid.to_i}")
+        CRYSTAL
+    end
+
+    it "respects MinConfidence" do
+      rule = CommandInjection.new
+      rule.min_confidence = "High"
+
+      expect_no_issues rule, <<-'CRYSTAL'
+        system("ls -l #{path}")
+        CRYSTAL
+
+      expect_issue rule, <<-'CRYSTAL'
+        system("cat #{env.params.query["file"]}")
+        # ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ error: Shell command built from interpolated string can lead to command injection
+        CRYSTAL
+    end
+
     it "reports system calls with interpolated strings" do
       expect_issue subject, <<-'CRYSTAL'
         system("ls -l #{path}")
--- a/spec/ameba/rule/security/command_injection_spec.cr
+++ b/spec/ameba/rule/security/command_injection_spec.cr
@@ -1,0 +1,41 @@
+require "../../../spec_helper"
+
+module Ameba::Rule::Security
+  describe CommandInjection do
+    subject = CommandInjection.new
+
+    it "passes for commands without interpolation" do
+      expect_no_issues subject, <<-CRYSTAL
+        system("ls -la")
+        `date`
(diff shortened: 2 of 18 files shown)


straight-shoota:
The general approach looks good to me.

But the individual rules need some discussion, so it might be best to separate them into their own PRs.

Some thoughts off the top of my head:
* Direct shell commands (i.e. the `system` and `` ` `` methods) are likely to be deprecated and removed from the language eventually (following [RFC 25]). So the `CommandInjection` rule might be of limited use.
* Using `Digest::MD5` or `Random.random_bytes` should not be considered a security issue. If it was, we would drop these features from stdlib. Typical use cases are not security relevant. A threat only exists when a weak hash or PRNG is used in a place where a stronger one would be necessary. But that's impossible to detect with only static analysis. So IMO these rules cannot produce a reliable signal, and tacking a disable directive every time you want to use MD5 is not helpful.
* Hard coded secrets are not specific to Crystal source files. These things tend to end up in configuration files as well. Thus `HardcodedSecret` is incomplete if you want to protect an entire repo from accidentally spilling secrets into VCS. Tools like gitleak or talisman handle this for any file in a repository, including source files. It's not bad to defend the same issue with multiple tools, but that makes it more difficult to add ignores.

[RFC 25]: https://github.com/crystal-lang/rfcs/pull/25

veelenga:
Will split it into smaller PRs

PR #891, opened 2026-08-19
Add `OnlyTwoElementUnions` setting to the `Style/VerboseNilType` rule

### LINKED ISSUE #890: `Style/VerboseNilType` with more than 1 type
This rule currently marks:
```cr
foo : String | Nil = nil
```
as invalid in favor of:
```cr
foo : String? = nil
```
which I generally agree with.

However it would be nice if there was an option to only enforce this when there are only 2 types in the union. It's currently getting flagged on some athena code that's like:
```cr
src/components/validator/src/spec/constraint_validator_test_case.cr:166:16
[C] Style/VerboseNilType: Prefer `?` instead of `| Nil` in unions
> groups : Array(String) | String | AVD::Constraints::GroupSequence | Nil,
```
which I do not like using `?` as it feels more confusing to read.

Sija:
Resolves #890

### DIFF
--- a/spec/ameba/rule/style/verbose_nil_type_spec.cr
+++ b/spec/ameba/rule/style/verbose_nil_type_spec.cr
@@ -125,6 +125,32 @@ module Ameba::Rule::Style
           bar : String | Nil
           CRYSTAL
       end
+
+      it "#only_two_element_unions" do
+        rule = VerboseNilType.new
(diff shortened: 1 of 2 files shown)


veelenga (approved):
lgtm

nobodywasishere (approved):

veelenga (approved):

